PR #3, opened 2016-02-10
Initial production Puma config

mmhmm:
- Rails env is always production
- workers set to 2 for now, unknown how many cores production machines
  will have for now. (See https://github.com/puma/puma#clustered-mode )


bradwright:
Good choice on Puma, we're not that happy with Unicorn over on GOV.UK.

whpearson:
Starts up happily now with  socket :+1:

PR #4, opened 2016-02-10
Create Rails tmp directories post package install

mmhmm:
We need these directories to exist prior to the app first running, as we're going to write the Rack server's
PID files etc there.


whpearson:
:+1:

PR #8, opened 2016-02-12
3321: allow for localised routes

robinwhittleton:
The route_translator gem allows route names to be moved into the locale dictionaries. Redirects will need to have `I18n.locale` passed through manually.


willp-bl:
Looks good :+1:

PR #10, opened 2016-02-16
bau: remove action mailer

chrisholmes:
We won't be sending any e-mails so let's remove actionmailer from the app.


robinwhittleton:
LGTM

PR #11, opened 2016-02-17
3321: show error if no answer to start page question

chrisholmes:
If there is no answer then we should prompt for one.

@robinwhittleton: please, could you let me know if I should correct the styling for this error


robinwhittleton:
More work needs to be done to update the error styles but this is enough for the time being.

robinwhittleton:
Will do that in a seperate commit.

PR #12, opened 2016-02-18
bau: set a custom user agent for client

willp-bl:
It would be nice to do this globally without having a private method
to configure the client, but this works.

Pair: WillPa


robinwhittleton:
LGTM

PR #17, opened 2016-02-26
3321: Add gov uk selection button styles

richardTowers:
These set the whole label to be highlighted when an input is selected.
This needs reviewing by @robinwhittleton and will probably want to
change.

In particular: I had to modify one of the scss files inside gov uk elements. I'm not sure this was the right thing to do? Will it make upgrading difficult?

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,10 +1,15 @@
 //= require jquery
 //= require jquery_ujs
 //= require jquery.validate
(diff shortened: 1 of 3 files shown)


robinwhittleton:
Yeah, I don’t like editing elements directly. What was the reasoning for that change?

joelanman:
You could use `extend` in your own scss to start with Elements style and then override the stuff you need

PR #18, opened 2016-03-01
3321: Add gov uk selection button styles

richardTowers:
These set the whole label to be highlighted when an input is selected.
This needs reviewing by @robinwhittleton and will probably want to
change.

In particular: I had to modify one of the scss files inside gov uk elements. I'm not sure this was the right thing to do? Will it make upgrading difficult?

(Recreated from #17 which was targetting the wrong branch)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,10 +1,15 @@
 //= require jquery
 //= require jquery_ujs
 //= require jquery.validate
+//= require govuk/selection-buttons
 //= require_tree .
 
 window.GOVUK.validation.init();
 
 $(function () {
+  // Use GOV.UK selection-buttons.js to set selected and focused states for block labels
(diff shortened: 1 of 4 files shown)


richardTowers:
@robinwhittleton @joelanman - I couldn't think of a way of `extend`ing anything without changing the markup, so I just added a `govuk-elements-overrides` file and blotted the background back to `initial`.

Good enough for now?

gemmaleigh:
Hi @richardTowers, @robinwhittleton, [it looks like this should be changed in GOV.UK elements](https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/forms/_form-block-labels.scss#L59).

Rather than override, would you mind submitting a PR to remove the background-colour here?

richardTowers:
Thanks, I'll try and remember to upgrade govuk elements and remove the override if alphagov/govuk_elements#188 gets merged.

PR #19, opened 2016-03-04
3347: add support for zdd endpoints

chrisholmes:
The commit provides support for the ZDD orchestration we use in our microservices. It is hoped that with this functionality in place we can drop the new frontend into our architecture without changing how we run our deployments. Our main goal for ZDD compliance is the ability to drain new requests from an instance of our microservices.

Our services are ZDD compliant, or able to drain, when they provide the following functionality:
* An endpoint `/service-status` that reports a 200 when the service is available for new requests or a 503 when it is unavailable and should no longer issue new requests
* A filter/middleware that will add a 'Connection: close' response header to all responses when the service is unavailable and should no longer handle new requests.
* The ability to set a property which sets the service unavailable to no longer handle new requests.

In order to provide this functionality to our app I have done the following:

1. Introduced an ENVVAR `ZDD_LATCH` that points to a file
2. Created an object `ServiceStatus` that will be `unavailable?` when the file defined by `ZDD_LATCH` exists
3. Add an endpoint `/service-status` that reports a 200 when `ServiceStatus.unavailable?` is false and a 503 when `ServiceStatus.unavailable?` is true
4. Add a middleware `ServiceStatusFilter` that will add `Connection: close` when `ServiceStatus.unavailable?` is true
5. Add a rake task `zdd:set_unavailable` that will create the file defined by `ZDD_LATCH` then sleep for a few seconds
6. Add a rake task `zdd:reset` that will remove the file defined by `ZDD_LATCH`
7. Introduce a custom upstart script that will run `zdd:reset` on startup and `zdd:set_unavailable` before stopping

### DIFF
--- a/lib/configuration.rb
+++ b/lib/configuration.rb
@@ -1,0 +1,17 @@
+class Configuration
+  MissingEnvVarError = Class.new(StandardError)
+  def self.load!(&blk)
+    config = Configuration.new
+    config.instance_eval(&blk)
+    config
+  end
+
+  def option(name, envvar)
+    value = ENV.fetch(envvar) do
+      raise MissingEnvVarError, "An Environment Variable named '#{envvar}' could not be found"
+    end
+    instance_variable_set("@#{name}", value)
+    eigenclass = class << self; self; end
+    eigenclass.class_eval { attr_reader name }
+  end
+end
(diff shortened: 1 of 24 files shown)


lib/configuration.rb
```diff
@@ -0,0 +1,17 @@
+class Configuration
+  MissingEnvVarError = Class.new(StandardError)
+  def self.load!(&blk)
```
davbo: what's the `&blk` thing here all about, is this a block?

chrisholmes: yeah, it's treating the blk as a first class function that we can call or pass to other functions.

It's one of the two main ways to work with blocks.

PR #20, opened 2016-03-10
3416 add logstash to front

oswaldquek:


config/environments/production.rb
```diff
@@ -46,7 +46,7 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
```
richardTowers: Do you need to change the comment here to reflect that the value is no longer the lowest log level?

oswaldquek: i think config.log_level can be deleted altogether actually. i think it logs at INFO by default in production.

richardTowers: I'd be inclined to keep it anyway; [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/). Maybe just remove the comment?

davbo: Did someone just quote the Zen of Python on our Rails project? :+1: :gem: :snake: :smile_cat:

oswaldquek: Amen to that :+1:

PR #21, opened 2016-03-11
bau: support custom ssl context

chrisholmes:
Major aim of this PR is to open up the ability to configure the TLS behaviour of our HTTP client. This is achieved by the supplying a SSLContext object which affects the TLS behaviour (verification mode, certs for verification, protocols etc).

In order to support this, I have adapted the configuration class to support default or missing values.


config/initializers/configuration.rb
```diff
@@ -7,4 +7,6 @@
   option_string 'logo_directory', 'LOGO_DIRECTORY'
   option_string 'zdd_file', 'ZDD_LATCH'
   option_string 'polling_wait_time', 'POLLING_WAIT_TIME'
+  option_string 'polling_wait_time', 'POLLING_WAIT_TIME'
```
willp-bl: Double option

willp-bl:
Looks good to me :+1:

PR #22, opened 2016-03-14
3461: chunky radios

robinwhittleton:
We want to bring back the existing faked radio buttons to maintain our success rate. When Chrome/OSX and Firefox/Win support resizing native buttons we can evaluate removing this code.
There’s an outstanding issue with Firefox not having the yellow outlines on focus, but I believe this is a problem in govuk_elements rather than this code and am investigating.


app/assets/javascripts/application.js
```diff
@@ -1,3 +1,4 @@
+//= require vendor/polyfills/bind
```
richardTowers: Was this needed before? You haven't added any other JS in this PR.

robinwhittleton:
The selection buttons JS relies on `.bind()` to pass through context references. This isn’t supported natively in IE8, so govuk_frontend_toolkit provides a polyfill. We weren’t including it, which is fine for modern browsers but breaks the implementation in IE8.

app/assets/javascripts/application.js
robinwhittleton: The selection buttons JS relies on .bind() to pass through context references. This isn’t supported natively in IE8, so govuk_frontend_toolkit provides a polyfill. We weren’t including it, which is fine for modern browsers but breaks the implementation in IE8.

richardTowers: :+1:

richardTowers:
image

PR #41, opened 2016-04-01
bau: provide a 404 page matching live

chrisholmes:
This change provides a 404 page for the user that matches the behaviour of live.

In order to catch the "404" Exception I have had to configure the app to refer
to our own routes for errors where it will route to the new ErrorsController. We
may want to route the 500 error this way in the future.


config/locales/en.yml
```diff
@@ -111,3 +111,10 @@ en:
       heading: Sorry, something went wrong
       error_message: This may be because your session timed out or there was a system error.
       start_again: Because this is a secure service, you'll need to start again.
+    page_not_found:
+      title: This page can't be found
+      heading: This page can't be found
+      reason_pre: "This may be because:"
+      reason_one: the link you clicked is broken
+      reason_one: you entered an incorrect web address
```
lloydnye: reason_two?

PR #42, opened 2016-04-04
3453: Add banner explaining incomplete translation

davbo:
As we add new pages to the frontend we will not always have a full
translation ready. This banner which only appears for Welsh language
users explains this

  GOV.UK Verify is in the process of being translated into Welsh. We're
  translating pages every week between now and the end of April. Your
  feedback helps us improve the translation.

The banner includes a link for users to provide feedback.


willp-bl:
Looks good

PR #48, opened 2016-04-06
bau: Small fixes recommended by rubocop

davbo:
Running rubocop with a stricter config than the GDS defaults raises some
interesting things such as freezing strings which are intended to be
constant (strings are mutable by default in Ruby). I think we should
consider making some changes to our rubocop config to catch these
things.

Includes a separate change which will only include the JasmineRails
routes if we are in test or development env, they are currently excluded
by the dependency gem not being available in production.

### DIFF
--- a/app/models/session_proxy.rb
+++ b/app/models/session_proxy.rb
@@ -1,13 +1,13 @@
 
 class SessionProxy
-  PATH = "/session"
-  FEDERATION_INFO_PATH = "#{PATH}/federation"
-  SELECT_IDP_PATH = "#{PATH}/select-idp"
-  IDP_AUTHN_REQUEST_PATH = "#{PATH}/idp-authn-request"
-  PARAM_SAML_REQUEST = "samlRequest"
-  PARAM_RELAY_STATE = "relayState"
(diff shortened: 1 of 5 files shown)


davbo:
oops I broke all the stuff.

app/models/session_proxy.rb
```diff
@@ -1,13 +1,13 @@
...
-  PATH = "/session"
-  FEDERATION_INFO_PATH = "#{PATH}/federation"
-  SELECT_IDP_PATH = "#{PATH}/select-idp"
-  IDP_AUTHN_REQUEST_PATH = "#{PATH}/idp-authn-request"
-  PARAM_SAML_REQUEST = "samlRequest"
-  PARAM_RELAY_STATE = "relayState"
-  PARAM_ORIGINATING_IP = "originatingIp"
-  PARAM_ENTITY_ID = 'entityId'
+  PATH = '/session'.freeze
+  FEDERATION_INFO_PATH = '#{PATH}/federation'.freeze
```
richardTowers: Should still be double quotes for the interpolation :)

PR #55, opened 2016-04-08
3517: integrate raven into the rails logger

chrisholmes:
Now when we send error messages to the Rails logger they will be sent to Sentry
using Raven as well as our other loggers. This should allow us to not have to explicitly capture exceptions
where we are currently logging errors.

In order to integrate with the Rails logger we provider a logger for raven that
implements the ::Logger interface. It is then added to the Rails logger as an
additional output using ActiveSupport::Logger::broadcast.

Also, provided an integration test.

### DIFF
--- a/spec/features/user_encounters_error_page_spec.rb
+++ b/spec/features/user_encounters_error_page_spec.rb
@@ -42,12 +42,24 @@ RSpec.describe 'user encounters error page' do
   end
 
   it 'will present error page when standard error occurs in upstream systems' do
-    stub_request(:post, api_saml_endpoint).to_raise(StandardError)
+    e = StandardError.new("my message")
+    stub_request(:post, api_saml_endpoint).to_raise(e)
     stub_transactions_list
     visit '/test-saml'
     click_button "saml-post"
     expect(page).to have_content "Sorry, something went wrong"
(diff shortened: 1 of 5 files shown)


spec/features/user_encounters_error_page_spec.rb
```diff
@@ -42,6 +42,28 @@
...
   it 'will present error page when standard error occurs in upstream systems' do
+    e = StandardError.new("my message")
+    stub_request(:post, api_saml_endpoint).to_raise(e)
+    stub_transactions_list
+    visit '/test-saml'
+    click_button "saml-post"
+    expect(page).to have_content "Sorry, something went wrong"
+    expect(page).to have_css "#piwik-custom-url", text: "errors/generic-error"
+    expect(page.status_code).to eq(500)
+  end
```
willp-bl: This looks like a copy/paste duplicate that can be removed

willp-bl:
Looks good 👍

spec/features/user_encounters_error_page_spec.rb
chrisholmes: No, but these tests could do with a bit of a clean up.

chrisholmes: oh wait...

willp-bl:
👍

PR #56, opened 2016-04-11
bau : only officially translated Welsh in the cy.yml file

yolinas:
So far, we have been using google translate for welsh text. We are now moving
into a model where we only have welsh in the translations file which is
officially translated. Everything else will be in English. This way it
will be easier to spot what needs translations.

@yolinas


willp-bl:
👍

PR #57, opened 2016-04-12
3517: Bar access when session is in completed state

chrisholmes:
When a session is completed the sessionId is set to 'no-current-session' and
this should prevent a user from being view pages on the Hub.

Currently, this check is happening within our backend API and the FE is
receiving a 403 response, that's sent to Sentry, and which isn't very clear or
convenient.

By performing the check within the FE we remove the call to the API and the
Sentry output.


richardTowers:
Looks good to me.

PR #58, opened 2016-04-13
BAU: Remove duplicated code from IDP eligibility checker

vixus0:
The IdpEligibility::Checker class defined a mask consisting of all the
document evidence. This was duplicated in Evidence::DOCUMENT_ATTRIBUTES
which would mean two places to maintain a list of document-related
evidence items. Evidence has now been made a submodule of
IdpEligibility which had the added bonus of removing some requires from
specs.

Author: @vixus0

### DIFF
--- a/app/models/evidence.rb
+++ b/app/models/evidence.rb
@@ -1,6 +1,0 @@
-module Evidence
-  PHONE_ATTRIBUTES = [:mobile_phone, :smart_phone, :landline].freeze
-  DOCUMENT_ATTRIBUTES = [:passport, :driving_licence, :non_uk_id_document].freeze
-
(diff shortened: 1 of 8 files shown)


richardTowers:
Seems sensible. Will merge once Travis says it's ok.

chrisholmes:
Including Evidence isn't really a useful pattern here as we're not trying to mixin behaviour here

vixus0:
Ah, I was just going with what was there already. I was a bit surprised about all those includes.

chrisholmes:
TBH, I think we should merge this in as moving `Evidence` is an improvement and the includes were already there.

PR #62, opened 2016-04-15
3531: Include selected evidence in query params

leoshaw:
When redirecting to the "no mobile phone" dead-end page, the "select
phone" page needs to include the user's selected evidence in the query
params, to preserve the behaviour of the old front-end for analytics


leoshaw:
Merged by Tom on the command line

PR #67, opened 2016-04-18
Revert "Revert "3531: Enable localised route for no mobile phone""

leoshaw:
This reverts commit 847ed42f105050fbd70b7c88402eb8ed9c1cc8ab.


whpearson:
This should be good to merge as we've defaulted to the new frontend now.

PR #69, opened 2016-04-19
3532 may not work if you live overseas

oswaldquek:
PR for the may-not-work-if-you-live overseas page

### DIFF
--- a/app/views/may_not_work_if_you_live_overseas/index.html.erb
+++ b/app/views/may_not_work_if_you_live_overseas/index.html.erb
@@ -1,0 +1,9 @@
+<% content_for :page_title, t('hub.may_not_work_if_you_live_overseas.title') %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <h1 class="heading-xlarge"><%= t 'hub.may_not_work_if_you_live_overseas.heading' %></h1>
+
+    <p> <a href="<%= @try_to_verify %>"><%= t 'hub.may_not_work_if_you_live_overseas.try_to_verify' %></a> </p>
+  </div>
+</div>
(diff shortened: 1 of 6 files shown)


app/views/may_not_work_if_you_live_overseas/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      <li><%= t 'hub.may_not_work_if_you_live_overseas.gas_electricity_bills' %></li>
+      <li><%= t 'hub.may_not_work_if_you_live_overseas.mortgage' %></li>
+      <li><%= t 'hub.may_not_work_if_you_live_overseas.student_loan' %></li>
+      <li><%= t 'hub.may_not_work_if_you_live_overseas.payday_loan' %></li>
+      <li><%= t 'hub.may_not_work_if_you_live_overseas.mail_order_account' %></li>
+    </ul>
+
+    <p><%= t 'hub.may_not_work_if_you_live_overseas.youll_also_need' %></p>
+
+    <p> <a href="<%= @try_to_verify %>"><%= t 'hub.may_not_work_if_you_live_overseas.try_to_verify' %></a> </p>
```
richardTowers: Could use [`link_to`](http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to) here.

app/views/may_not_work_if_you_live_overseas/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+    <h2 class="heading-medium"><%= t 'hub.may_not_work_if_you_live_overseas.other_ways_heading', other_ways_description: @other_ways_description %></h2>
+    <div class="other-ways">
+      <p><%= @other_ways_text.html_safe %></p>
```
richardTowers: Maybe remove this `<p>` or replace it with a div. `<p>` tags in HTML can only contain "[Phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content)", which could cause (probably minor) styling bugs if there are block level elements (`div`s, `p`s, `ul`s etc.) in the `other_ways_text`.

PR #72, opened 2016-04-20
3532 will not work without uk address

oswaldquek:


richardTowers:
Looks good to me. Will wait for Travis and then merge.

PR #75, opened 2016-04-21
3568: start storing evidence answers in session cookie

chrisholmes:
Stores the selected evidence answers in the session, but for now they will not
be read anywhere in the application.

The intention is that we will use the session store to remove the use of the
propagated query params.

Answers are currently stored and associated with a key for the type of
evidence/page they are related to. This way we should be given a guarantee that
only the most recent answers to questions are stored in the cache.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,6 +40,18 @@ class ApplicationController < ActionController::Base
     }
   end
 
+  def store_selected_evidence(hash)
+    stored_selected_evidence.merge!(hash)
+  end
+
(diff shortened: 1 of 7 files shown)


joelanman:
will this affect the reporting of selected evidence in piwik?

Veda:
@joelanman It shouldn't as we would still report to piwik urls behind the scenes. We have evidence in the parameters and session at the moment, our plan is to get rid of the urls and retain in session store. Piwik analytics should not be affected :)

joelanman:
nice, cheers - just checking!

PR #76, opened 2016-04-22
3564: Adding officially translated welsh text.

Veda:
- cy.yml is updated as part of this.
- config locale specific text is yet to be updated.

Authors: @veda


SrinivasMurty:
The Yes and No option translations are unique to each question. We should take this into consideration and accommodate the changes.

Veda:
Working on it @SrinivasMurty

PR #77, opened 2016-04-22
3342: Move HTML formatting into localisation file

vixus0:
The template shouldn't constrain how the content for the why-companies
page can be formatted (within reason).

Author: @vixus0

### DIFF
--- a/app/views/why_companies/index.html.erb
+++ b/app/views/why_companies/index.html.erb
@@ -4,14 +4,8 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <%= link_to t('navigation.back'), @choose_a_certified_company_link, class: 'link-back' %>
-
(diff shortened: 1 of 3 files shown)


Veda:
I would like the formatting to be in the view rather than being localised. I don't see a reason why you should be localising the html formatting ? Am i missing some context here @vixus0 ?

vixus0:
There are a couple of reasons, that are similar to why we put the formatting for IDP descriptions in the localisation config files:

* We're confusing layout formatting (the view) with content formatting (the localisation file).
* Modifying content should not require changing the view.
* There's no reason we should be forcing translators to use exactly five paragraphs.
* There may be additional HTML elements required (`<em>` etc.) that are unavailable using the current structure.

willp-bl:
👍 good to have flexibility around the content

Veda:
Ah, that makes sense with paragraph as examples. 👍

PR #78, opened 2016-04-25
3530: Add dead-end page for no documents

leoshaw:
This adds the dead-end page that the user will see if no IdPs can verify
at the 'select documents' stage.  It's not currently possible to hit
this page because at least one IdP can verify with no documents, but
it's good to have this page as a fall-back in case that changes.


willp-bl:
Looks good, but has merge conflicts at the moment

PR #79, opened 2016-04-26
bau: Update welsh translations

richardTowers:
Copied the .yml file from an email @srinivasmurty forwarded me.
Replaced non-ASCII characters in routes with their closest ASCII
equivalents.

Solo: Rich


richardTowers:
@chrisholmes mentions that there's a risk of some service interruption during deployment for some Welsh users if we switch to the new routes without continuing to support the old `...-cy` routes.

I had a look in analytics and it looks like these routes are viewed pretty rarely though. I think it's probably fine to chance it.

richardTowers:
I added support for the old routes back in to avoid any issues during the deployment. It probably wouldn't have been a problem but it's not too much work to temporarily maintain support for the old routes.

PR #80, opened 2016-04-26
3559 add fingerprint to new frontend

leoshaw:


app/assets/javascripts/verify_fingerprint.js
```diff
@@ -0,0 +1,27 @@
...
+(function (exports) {
+   'use strict';
+
+    // based on jQuery's param implementation https://github.com/jquery/jquery/blob/master/src/serialize.js
+    function serialiseComponents(components) {
+        var componentsToExclude = ['webgl'];
+        var r = [];
+        jQuery.each(components, function() {
+            if(componentsToExclude.indexOf(this.key) == -1) {
+                r[r.length] = encodeURIComponent(this.key) + "=" + encodeURIComponent(this.value);
```
richardTowers: I appreciate that this has been lifted from jQuery, but I'd be inclined to use `r.push(...)` instead of `r[r.length] = ...`.

There's some [discussion around performance](http://stackoverflow.com/a/15650115/1344760) with unclear results, but push feels clearer to me.

app/views/shared/_fingerprint.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= image_tag('', style: 'border: 0', height: '1', width: '1', id: 'fp_img') %>
+<%= javascript_tag do %>
```
richardTowers: It would be good if this inlined script could be moved into the `verify_fingerprint` JS file. Could pull the `fingerprint_path` from a `data-` attribute on the image?

PR #81, opened 2016-04-27
3461: styling fixes

robinwhittleton:
This PR updates govuk_frontend_toolkit / elements / templates to the latest versions. As part of this it untangles the work that had previously been committed into app/assets/stylesheets/elements as this would have been wiped out during the upgrade.

Finally, it upgrades govuk-lint to 1.0.0 and makes the subsequent amends.


willp-bl:
LGTM

PR #87, opened 2016-05-03
3528: Port cookie page

oswaldquek:
This ports the cookie page. English and Welsh have their own static pages.

solo: @oswaldquek


config/routes.rb
```diff
@@ -63,18 +63,18 @@
   get 'service-status', to: 'service_status#index', as: :service_status
   get '/assets2/fp.gif', to: proc { |_| [200, {}, ['OK']] }
 
+  get 'cookies', to: 'cookies#index', as: :cookies
```
robinwhittleton: Does this not want to be inside the localised block? Or are we waiting to get the localisation first?

spec/features/user_visits_cookies_page_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'feature_helper'
+
+RSpec.describe 'When the user visits the cookies page' do
+  it 'displays the page in English' do
+    visit '/cookies?locale=cy'
```
oswaldquek: oops this needs to be be just /cookies

config/routes.rb
oswaldquek: @robinwhittleton fixed

PR #94, opened 2016-05-12
3636: new headings

robinwhittleton:
Make sure the question pages all have appropriate headings. Text and design courtesy of @alextea.

The only translation needed is the new heading on /will-it-work-for-me. The rest was just shuffling around.


willp-bl:
LGTM 👍

PR #95, opened 2016-05-12
3632: accessibility amends

robinwhittleton:
Grab bag of mini accessibility improvements, but as a quick list:

- fix the specified language of the remaining footer links
- update the cookies page to make it more legible
- fix broken `<p>` tags on the about page
- `<form>` tags need a non-empty action according to the specs
- make sure that all radio buttons are programmatically associated with their labels
- fix the red-on-yellow error text contrast

### DIFF
--- a/app/views/choose_a_certified_company/_idp_option.html.erb
+++ b/app/views/choose_a_certified_company/_idp_option.html.erb
@@ -6,16 +6,17 @@
     </p>
   </div>
   <% button_class = recommended ? 'button' : 'button-link' %>
-  <%= form_for(identity_provider, url: '', html: {class: 'idp-option'}) do |f| %>
+  <%= form_for(identity_provider, url: choose_a_certified_company_submit_path, html: {class: 'idp-option', id: nil}) do |f| %>
       <%= f.button t('hub.choose_a_certified_company.choose_idp', name: identity_provider.display_name),
                      class: button_class,
                      name: identity_provider.simple_id,
+                     id: nil,
                      type: 'submit',
                      value: identity_provider.display_name
       %>
-      <%= hidden_field_tag 'selected-idp', identity_provider.entity_id %>
-      <%= hidden_field_tag 'recommended-idp', recommended %>
-      <%= f.hidden_field :entity_id %>
-      <%= f.hidden_field :simple_id %>
+      <%= hidden_field_tag 'selected-idp', identity_provider.entity_id, id: nil %>
+      <%= hidden_field_tag 'recommended-idp', recommended, id: nil %>
+      <%= f.hidden_field :entity_id, id: nil %>
+      <%= f.hidden_field :simple_id, id: nil %>
   <% end %>
 </div>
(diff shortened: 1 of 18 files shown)


richardTowers:
> `<form>` tags need a non-empty action according to the specs

I did not know that! https://w3c.github.io/html/sec-forms.html#element-attrdef-form-action

We were using them originally because we wanted the browser to include the query params in its post (which seems to be consistent behaviour for empty action). No need for that now we're using the session cookie to store the selections though.

app/views/choose_a_certified_company/_idp_option.html.erb
```diff
@@ -6,16 +6,17 @@
     </p>
   </div>
   <% button_class = recommended ? 'button' : 'button-link' %>
-  <%= form_for(identity_provider, url: '', html: {class: 'idp-option'}) do |f| %>
+  <%= form_for(identity_provider, url: choose_a_certified_company_submit_path, html: {class: 'idp-option', id: nil}) do |f| %>
```
richardTowers: Why `id: nil`? Isn't this going to confuse Rails' `form_for` thing?

app/views/select_documents/index.html.erb
```diff
@@ -17,58 +17,58 @@
             <div class="form-group">
               <fieldset class="inline">
                 <span>1. <%= t 'hub.select_documents.question.driving_licence'%></span>
-                <label class="block-label">
+                <%= f.label :driving_licence_true, {class: 'block-label', onclick: ''} do %>
```
richardTowers: Why `onclick: ''` ? Is this an accessibility thing? Is there a wai- attribute that would achieve the same thing in a more self-documenting way?

joelanman: I _think_ this is to fix an old iOS bug where clicks on labels would not activate the checkbox unless theres an onclick

robinwhittleton: Yep, iOS 5’s Safari has a bug where clicking on the label won’t activate the input without this one weird hack. Not a problem usually, you’d just click on the radio, but we’re hiding the native implementation so we need this.

app/views/choose_a_certified_company/_idp_option.html.erb
robinwhittleton: `id` attributes need to be unique on a page. As this is a fragment that’s iterated we need to make sure that either they have unique `id`s or none. The form submission happens against the `name` so this doesn’t seem to break anything.

richardTowers: 👍 thanks for explaining.

app/views/select_documents/index.html.erb
richardTowers: Maybe we should think about pulling the markup for a radio button out into a partial / helper so we don't have to duplicate the `onclick: ''`. That way we could also put a comment explaining why it's there.

Discussed this and it'll probably be quite tricky, so I'll go ahead and merge this and look at that separately.

PR #96, opened 2016-05-12
transaction display info refactor

chrisholmes:
Changes include:
*   **use a helper method for current transaction info**:
Rather than duplicating the logic for retrieving the current transaction we can
instead use this helper.
*  **remove transaction info getter object**:
Transaction info getter was pretty much just behaving as a nil check so I've moved
this into the repository


app/models/display/rp/repository.rb
```diff
@@ -8,6 +8,9 @@ def initialize(translator)
       end
 
       def fetch(simple_id)
+        if simple_id.nil?
+          raise StandardError, 'No transaction simple id in session'
```
lloydnye: Would this be better in current_transaction_simple_id? as that's where we try to access the session

chrisholmes: Maybe, though it would be a bit of a pain to test it.

PR #97, opened 2016-05-13
bau: Add "funky" pass and fail messages to pre-commit

richardTowers:
This doesn't otherwise change the behaviour of pre-commit - it still
runs all the tests even if some fail.

This doesn't add checking of Jenkins or anything clever, it's literally
just the big text.

Solo: @richardtowers


willp-bl:
It would be nice if it skipped subsequent steps when the first failure occurs

willp-bl:
But 👍

richardTowers:
Yeah, @lloydnye said that too. I wasn't sure whether that was the desired behaviour. It's sometimes nice to see all the errors at once so you can go through and fix them all.

willp-bl:
Ok, will merge and we can change it later if we want to.  FWIW pre-commit takes 2m30s on my machine.

richardTowers:
Ouch. Took 44 seconds on mine.

PR #99, opened 2016-05-16
BAU: consolidate controllers

robinwhittleton:
There are a few pages that have their own controllers without needing to. This PR creates a controller for static pages (e.g. cookies, privacy policy) and migrates the fail state pages for the evidence questions into their parent question’s controller.


richardTowers:
Nice, this is much neater.

PR #100, opened 2016-05-16
BAU: Remove custom panels CSS

vixus0:
We had carried over some CSS from the old frontend for panels (elements
that were distinguished from surrounding text with a left border). The
panels CSS already exists in GOV.UK elements and our templates should
be using those classes as they are visually nearly identical.

Author: @vixus0


robinwhittleton:
Hah, great, I hadn’t realised that existed.

PR #102, opened 2016-05-17
BAU: Style "button-link"s to have pointer cursors

richardTowers:
These links are only used in the non-recommended IdPs section at the
moment, so they're not that prominent.

Solo: @richardtowers


robinwhittleton:
OK. I’m wondering if we should add `role="link"` to them too though.

richardTowers:
@robinwhittleton - yes, that sounds like a good idea. Added that to the PR.

PR #103, opened 2016-05-18
3626: announce form changes when selecting no documents checkbox

robinwhittleton:
Sighted users can see the checkboxes change to ‘no’ when the checkbox on the select documents form is checked. To replicate this behaviour for screenreader users we need to announce this change dynamically.

A visually hidden message is read out at the point that the checkbox is activated. When it is deactivated the message is blanked, but that doesn’t get read out because it’s empty.

### DIFF
--- a/app/views/select_documents/index.html.erb
+++ b/app/views/select_documents/index.html.erb
@@ -1,6 +1,8 @@
 <% content_for :page_title, t('hub.select_documents.title') %>
 <% content_for :feedback_source, 'SELECT_DOCUMENTS_PAGE' %>
 
+<div id="no-documents-message" class="visually-hidden" aria-live="assertive" data-no-documents-message="<%= h t('hub.select_documents.no_documents_message') %>"></div>
+
 <div class="grid-row">
   <div class="column-two-thirds">
     <h1 class="heading-large"><%= t 'hub.select_documents.title' %></h1>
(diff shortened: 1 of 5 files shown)


willp-bl:
👍 tested it in OSX.  Is there any way to make it read what the option is for the checkboxes when they have focus? It just says "No, selected, radio button" when I have tabbed into one of them.

robinwhittleton:
Do you mean for the radios? Ideally the text to the left would be a legend which would be read out alongside the radios, but it’s very hard to style a legend to look like the current layout. Lorna would like to revisit the designer of this page anyway, so let’s leave it until then.

app/views/select_documents/index.html.erb
```diff
@@ -1,6 +1,8 @@
 <% content_for :page_title, t('hub.select_documents.title') %>
 <% content_for :feedback_source, 'SELECT_DOCUMENTS_PAGE' %>
 
+<p id="no-documents-message" class="visually-hidden" aria-live="assertive" data-no-documents-message="<%= t 'hub.select_documents.no_documents_message' %>"></p>
```
richardTowers: Any reason for using a `p` over a more generic element (like a `span` or `div`?).

Do we need to attribute escape the no documents message in case someone sticks a quote in there in the future?

robinwhittleton: No particular reason, I’m happy to change it if you think it makes sense. Probably do need escaping yes.

robinwhittleton:
Now with escaping and a `div`.

PR #104, opened 2016-05-18
3642: Other ways to access service page

oswaldquek:
routes.rb includes other-ways-to-access-service-cy to ensure ZDD compliance
when we get to releasing the localized version in a later release.

@oswaldquek

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -246,6 +246,8 @@ en:
         <p>You may have selected the wrong company. Check your emails and text messages for confirmation of who verified you.</p>
         <p>You can be verified by another company at any time. You won’t lose any data on GOV.UK services you’ve used.</p>
       start_again: "Start again"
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -246,6 +246,8 @@ en:
         <p>You may have selected the wrong company. Check your emails and text messages for confirmation of who verified you.</p>
         <p>You can be verified by another company at any time. You won’t lose any data on GOV.UK services you’ve used.</p>
       start_again: "Start again"
+    other_ways_to_access_service:
+      title: Other ways to access the service
```
willp-bl: This is `Other ways to %{other_ways_description}` elsewhere

willp-bl: ...but old frontend has this for the title

willp-bl:
👍

PR #105, opened 2016-05-18
3648: set the locale for rescued errors

chrisholmes:
When an exception is thrown from within a controller a class method `rescue_from` is used to rescue and render an error page. As this is a occuring within the context of the class method, the locale set within an instance of a controller is lost and we need to call set_locale witin the error page rendering methods.


davbo:
Aha, that was a bit of a puzzle with the class method / instance context changing. Nice work.

PR #107, opened 2016-05-19
3626: fix iOS Voiceover skip link bug

robinwhittleton:
Fixed upstream in [govuk_template](https://github.com/alphagov/govuk_template) / [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit).


Gemfile.lock
```diff
@@ -128,7 +128,7 @@ GEM
       mime-types-data (~> 3.2015)
     mime-types-data (3.2016.0221)
     mini_portile2 (2.0.0)
-    minitest (5.8.4)
+    minitest (5.9.0)
```
lloydnye: Should this be changed in this commit?

richardTowers:
Is 3626 the right story number? The internal issue tracker has it as updating an interstitial page.

PR #109, opened 2016-05-19
3564: Added welsh translations

Veda:
Authors: @veda

### DIFF
--- a/app/views/static/privacy_notice.cy.html.erb
+++ b/app/views/static/privacy_notice.cy.html.erb
@@ -3,170 +3,177 @@
 
 <div class="grid-row">
   <div class="column-two-thirds">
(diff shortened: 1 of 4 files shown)


robinwhittleton:
Do we want to change the updated date? We’re translating but not altering the meaning of the content, and it might feel a bit odd to have a different date to the versions if the user switches between them.

Veda:
@robinwhittleton you are right.. I shall change it to 27th April perhaps so it is same as english version.

oswaldquek:
@robinwhittleton @Veda I'm confused - where is this updated date?

Veda:
It was all the way at the bottom of the privacy-notice file, since i put it back to what i was it isn't showing up in the changes :)

oswaldquek:
Ok. Best to comment on the line itself in the code in the future :)

PR #110, opened 2016-05-19
3642: localize other ways page

oswaldquek:
@oswaldquek

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -52,6 +52,7 @@ Rails.application.routes.draw do
     get 'confirmation', to: 'confirmation#index', as: :confirmation
     get 'failed_registration', to: 'failed_registration#index', as: :failed_registration
     get 'failed_sign_in', to: 'failed_sign_in#index', as: :failed_sign_in
+    get 'other_ways_to_access_service', to: 'other_ways_to_access_service#index', as: :other_ways_to_access_service
   end
 
   put 'redirect-to-idp-warning', to: 'redirect_to_idp_warning#continue_ajax', as: :redirect_to_idp_warning_submit_ajax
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -52,16 +52,14 @@
...
   end
 
   put 'redirect-to-idp-warning', to: 'redirect_to_idp_warning#continue_ajax', as: :redirect_to_idp_warning_submit_ajax
   put 'select-idp', to: 'sign_in#select_idp_ajax', as: :select_idp_submit_ajax
   get '/redirect-to-service/error', to: redirect("#{API_HOST}/redirect-to-service/error")
   get 'service-status', to: 'service_status#index', as: :service_status
   get '/assets2/fp.gif', to: proc { |_| [200, {}, ['OK']] }
-  get 'other-ways-to-access-service', to: 'other_ways_to_access_service#index', as: :other_ways_to_access_service
-  get 'other-ways-to-access-service-cy', to: 'other_ways_to_access_service#index' #TODO: delete after localization
-  get 'other-ways-to-access', to: 'other_ways_to_access_service#index' #TODO: delete after localization
```
willp-bl: Just checking that this line is safe to remove?

oswaldquek: Yes. /other-ways-to-access is only used in old frontend (see https://github.gds/gds/ida-frontend/blob/master/frontend/src/main/java/uk/gov/ida/hub/frontend/resources/OtherWaysResource.java#L41) and not new frontend. But now that /other-ways-to-access-service is live and is being routed to the new frontend (https://github.gds/gds/ida-webops/wiki/Verify-frontend-release-notes#ida-webops-3) we can remove this line.

PR #113, opened 2016-05-20
3657: stop sending 404s to sentry

chrisholmes:
We currently hook up sentry to the raven logger to capture exceptions and error
messages. 404s are logged as errors using the `ActionController::RoutingError`.
As 404s are not something we can action usually this change will remove them
from Sentry.


willp-bl:
LGTM

PR #114, opened 2016-05-23
bau: add https://www.signin.service.gov.uk/humans.txt

willp-bl:
Text based on this: https://www.gov.uk/humans.txt

Pair: WillPa


app/controllers/static_controller.rb
```diff
@@ -6,4 +6,8 @@ def cookies
 
   def privacy_notice
   end
+
+  def humanstxt
+    render 'static/humans.txt.erb'
```
davbo: For things like this I believe it would be possible to just 

`render plain: 'GOV.UK Verify is built by a team at the Government Digital Service in London. If you'd like to join us, see https://identityassurance.blog.gov.uk/work-with-us/'`

Rather than rendering an erb template.

robinwhittleton: Or just chuck it in `/public`?

davbo: or that! :)

willp-bl: done

PR #116, opened 2016-05-23
BAU: Use the I18n module's html convention

vixus0:
Instead of manually calling .html_safe on our translation string in the
view, use the I18n module's convention for automatically allowing HTML
in translation keys that end in _html. This has the added benefit of
making clear in the translation file that the translation string will
NOT be sanitised for HTML.

Author: @vixus0

### DIFF
--- a/app/views/errors/no_cookies.html.erb
+++ b/app/views/errors/no_cookies.html.erb
@@ -7,7 +7,7 @@
     <h1 class="heading-xlarge"><%= t 'errors.no_cookies.heading' %></h1>
     <p><%= t 'errors.no_cookies.access_restriction' %></p>
     <%= render partial: 'errors/transaction_list' %>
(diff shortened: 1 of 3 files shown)


app/views/errors/no_cookies.html.erb
```diff
@@ -7,7 +7,7 @@
     <h1 class="heading-xlarge"><%= t 'errors.no_cookies.heading' %></h1>
     <p><%= t 'errors.no_cookies.access_restriction' %></p>
     <%= render partial: 'errors/transaction_list' %>
-    <p><%= t('errors.no_cookies.read_more').html_safe %></p>
+    <p><%= t('errors.no_cookies.read_more_html') %></p>
```
robinwhittleton: If we’re editing this line can we remove the parens to match the other translation calls?

vixus0: Yeah, makes sense

PR #117, opened 2016-05-23
3628: store current identity providers in session

chrisholmes:
Identity Providers will be lazily loaded into the session object as they're
required. This should reduce the number of internal API calls needed when we
need to work with the list of available IDPs.

This has been taken as a step towards returning the list of IDPs at the start of
a session alongside the session's transaction.

I'm of two minds regarding whether this code should be placed inside a mixin or
interstitial controller base class (e.g. ChooseACertifiedCompanyController
< FederationController < ApplicationController) so I'm open to opinions.

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -9,15 +9,9 @@ class AboutController < ApplicationController
   end
 
   def certified_companies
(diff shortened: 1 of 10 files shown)


richardTowers:
:+1:, this is a nice approach. My view is that we should leave it in ApplicationController for now, and consider the options for a refactoring later.

I'd probably lean towards a mixin if that's achievable. Any controller hierarchy is going to feel quite forced I think...

PR #118, opened 2016-05-23
3592: remove phase banner and add feedback link

robinwhittleton:
The phase banner needs to be removed for go-live, but we want to keep the feedback link. This PR removes the banner and adds in a GOV.UK-style link at the bottom of the content. Rather than pop open a short form in the GOV.UK style this just links through to the existing feedback form on old frontend.


willp-bl:
Looks good to me 👍

PR #120, opened 2016-05-24
3475 welsh titles in english

Veda:
This PR will render page titles in english in a meta tag so that we could use english titles irrespective of the language chose to report to analytics.


yolinas:
:+1:

PR #122, opened 2016-05-26
3461: fix details / summary arrow in Firefox nightly / developer edition

robinwhittleton:
Mozilla has implemented native `<details>` / `<summary>` elements in Firefox in such a way that it broke their arrow with our styling. The latest version of [govuk_elements](https://github.com/alphagov/govuk_elements) has a fix for this.


vixus0:
Is there a reason for commenting out instead of just removing?

vixus0:
Ah right, that makes sense.

PR #123, opened 2016-05-27
3322: Localise response processing

lloydnye:
Solo: @lloydnye


config/routes.rb
```diff
@@ -56,6 +54,7 @@
     get 'failed_sign_in', to: 'failed_sign_in#index', as: :failed_sign_in
     get 'other_ways_to_access_service', to: 'other_ways_to_access_service#index', as: :other_ways_to_access_service
     get 'forgot_company', to: 'static#forgot_company', as: :forgot_company
+    get 'response_processing' => 'response_processing#index', as: :response_processing
```
robinwhittleton: We presumably want to use `to:` instead of `=>`?

PR #128, opened 2016-06-03
3673 Show no docs interstitial page content when only foreign id is ch…

Veda:
…osen

This is a need only for experian at the moment.
However, we don't drive the display logic by IDPs
instead we manage it through the content configuration
for IDPs. We have no_docs content only for experian at the moment.

Authors: @veda

### DIFF
--- a/spec/features/user_visits_redirect_to_idp_warning_page_spec.rb
+++ b/spec/features/user_visits_redirect_to_idp_warning_page_spec.rb
@@ -37,6 +37,13 @@ RSpec.describe 'When the user visits the redirect to IDP warning page' do
       selected_evidence: { phone: %w(mobile_phone smart_phone), documents: [] },
     )
   }
+  let(:given_a_session_with_non_uk_id_document_evidence) {
+    page.set_rack_session(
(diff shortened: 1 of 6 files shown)


spec/features/user_visits_redirect_to_idp_warning_page_spec.rb
```diff
@@ -147,6 +154,16 @@
...
   end
 
+  it 'includes specific IDP text and link to the other ways when user has only foreign id document' do
+    page.set_rack_session(transaction_simple_id: 'test-rp')
+    given_a_session_with_non_uk_id_document_evidence
+    visit '/redirect-to-idp-warning'
+
+    expect(page).to have_content 'You’ll now verify your identity on Best ID’s website.'
+    expect(page).to have_content 'Additional IDP Instructions'
+    expect(page).to have_link 'other ways to register for an identity profile', href: other_ways_to_access_service_path
```
yolinas: Can we use the translations here instead of plain text?

Veda: Ofcourse.... Sorry @yolinas i keep forgetting to do so

PR #131, opened 2016-06-09
BAU: fix confirmation button alignment

robinwhittleton:
The confirmation page is using the `.hub-start` class to get a bigger heading than usual, but that has the side effect of pushing the button left by 15px (we do this on the start page to align with the radio buttons).

If we use [elements’ `.heading-x` classes](http://govuk-elements.herokuapp.com/typography/#typography-headings) to control heading styles we can remove our custom weights and the class from the confirmation page that was affecting the button margin. The only change we need to make to our CSS is to remove the top margin on headings.

### DIFF
--- a/app/assets/stylesheets/pages/_slides.scss
+++ b/app/assets/stylesheets/pages/_slides.scss
@@ -28,14 +28,12 @@
   }
 
   h1 {
-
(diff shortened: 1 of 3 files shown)


willp-bl:
👍

PR #134, opened 2016-06-13
3638: fix error page feedback links

robinwhittleton:
`session_timeout` and `cookie_expired` pages have broken feedback links. This updates their link paths, and adds a check to the session_timeout spec to make sure it’s there.


willp-bl:
👍

PR #135, opened 2016-06-13
bau: log at :warn and above in prod-like environments

willp-bl:
Pair: WillPa


oswaldquek:
what's the rationale for changing the level? is logging at INFO giving us too much noise?

PR #136, opened 2016-06-13
BAU: refactor GOV.UK elements overrides SASS

robinwhittleton:
`_govuk-elements-overrides.scss` has become a dumping ground for homeless styles. This migrates the styles out into view-specific or helper components, and normalises the `.scss` naming convention to standardise on hyphens.


willp-bl:
LGTM 👍

PR #140, opened 2016-06-15
3685 dont trust user input

leoshaw:
There are a few places where we rely on form params to provide
information about the IdP selected by the user, for instance the entity
id and display name.

There's no need to do this any more, because this information can be
retrieved from the session given the simple id, and doing it this way
reduces the opportunity for a malicious user to introduce bad data.

This change removes the entity id and display name from IdP selection
forms, and just uses the simple id to look up the IdP and get the rest
of the information.


chrisholmes:
LGTM

PR #142, opened 2016-06-20
3668: Remove selected evidence from session

richardTowers:
**Please do not merge** - 1a0ab09 needs to have been live for at least one session length (2 hours) to ensure all the users have the `selected_answers` hash in their sessions.

We are now using selected answers exclusively for determining which IDPs
are displayed to the user. SelectedAnswerStore is responsible for
mapping the user's answers to the evidence array that is used for IDP
filtering.

Authors: @vixus0 @andrien.ricketts @richardTowers

### DIFF
--- a/spec/models/select_documents_form_spec.rb
+++ b/spec/models/select_documents_form_spec.rb
@@ -59,30 +59,6 @@ describe SelectDocumentsForm do
     expect(form.errors.full_messages).to eql ['Please check your selection']
   end
 
-  describe '#selected_evidence' do
-    it 'should return a list of the selected evidence' do
-      form = SelectDocumentsForm.new(
-        driving_licence: 'true',
-        passport: 'true',
-        non_uk_id_document: 'false',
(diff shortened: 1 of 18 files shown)


spec/models/select_documents_form_spec.rb
```diff
@@ -102,15 +78,15 @@
...
+      expect(evidence).to eql(driving_licence: true)
     end
 
     it 'should return all documents answers as false if no documents is checked' do
       form = SelectDocumentsForm.new(
         no_documents: 'true'
       )
       evidence = form.selected_answers
-      expect(evidence).to eql(driving_licence: 'false', passport: 'false', non_uk_id_document: 'false')
+      expect(evidence).to eql(driving_licence: false, passport: false, non_uk_id_document: false)
```
chrisholmes: s/evidence/answers

richardTowers: Fixed!

PR #144, opened 2016-06-21
3729: proxy ip addr as header when request RP response

chrisholmes:
The IP address for the user is sent as the X-Forwarded-For header


willp-bl:
👍

PR #145, opened 2016-06-22
3609: omit idps with non supporting profiles on picker page

chrisholmes:
If an IDP is not able to support any of the evidence profiles that a user has declared to us then we will no longer list them on the /choose-a-certified-company page. If we feel that an IDP supports some of the user's evidence profile and is likely to verify the user then we mark the IDP as recommended, but if we don't feel they are  likely verify while supporting some of the user's evidence profiles then the IDP is display within the unlikely to verify group.

To do this we have created two tiers of profiles (recommended and non-recommended) which allow for the grouping stated above. If an IDP is not found to be in either of these groups when the user's evidence is supplied then it is omitted from the page.

In addition to these changes we have performed a small refactor of the code primarily aimed at renaming rules to profiles.


stub/idp-rules/stub-idp-two.yml
```diff
@@ -1,3 +1,4 @@
 simpleIds: [ "stub-idp-two" ]
-rules:
```
lloydnye: Should the filenames change too?

lloydnye:
There are a few instances of rules left actually, might be worth a quick grep to clean them up.

willp-bl:
LGTM when it's rebased and any rules -> profiles clean up that can be done is

stub/idp-rules/stub-idp-two.yml
chrisholmes: because of the hints works I'm not entirely sure whether a rename is appropriate at this stage as these files are starting to gain extra responsibilities.

davbo:
Great, thanks for doing all that rebasing!

PR #146, opened 2016-06-22
refactor: check federation locale data at startup

chrisholmes:
On starting up the application the locales for federation members is checked for
completeness and the app will crash if data is missing.

A new DisplayData object has been defined that is used to provide the
information we want displayed.

Translations are still generated as they're needed.


spec/spec_helper.rb
```diff
@@ -90,5 +91,7 @@
   Kernel.srand config.seed
 =end
 end
+
 $:.unshift File.expand_path('../../app/', __FILE__)
```
lloydnye: Can we switch to $LOAD_PATH, it's more obvious to non rubyists

PR #147, opened 2016-06-23
BAU: move page title to Rails 3+ syntax

robinwhittleton:
Rails 2 used to require specifying the partial and locals keys, but in Rails 3+ you can omit them. Might as well do this now rather than propagating the older and more verbose syntax.


richardTowers:
✨

PR #148, opened 2016-06-24
BAU: bump govuk_template to 0.18.0

robinwhittleton:
From our point of view, this removes a big chunk of JS for IE8 users, and [gets rid of a block of inline JS from the header](https://github.com/alphagov/govuk_template/pull/219/files#diff-3) that would otherwise have to be taken account of when the CSP work kicks off.


richardTowers:
Smashing 👍

PR #149, opened 2016-06-24
zd21812: Fix undefined page titles

oswaldquek:
Piwik was reporting the "about idp" page titles as "undefined". This is because
    about.html.erb wasn't using _page_title.html.erb which sets the
    meta[name="verify|title"] which is required by analytics.js to send the title.

    Changed the design to call ApplicationHelper.page_title to render titles, and
    getting rid of app/views/shared/_page_title.html.erb. This
    will allow "about idp" page to pass the display name and still have a common
    solution for displaying page titles on all pages.

solo: @oswaldquek


oswaldquek:
DO NOT MERGE, @chrisholmes has a better solution

app/views/choose_a_certified_company/about.html.erb
```diff
@@ -1,6 +1,4 @@
-<% content_for :page_title, t('hub.choose_a_certified_company.about.title', display_name: @idp.display_name) %>
-<% content_for :feedback_source, "CHOOSE_A_CERTIFIED_COMPANY_ABOUT_#{@idp.simple_id.upcase}_PAGE" %>
```
richardTowers: Should we be removing feedback source here?

oswaldquek:
@richardTowers good spot, thanks. I've fixed it.

chrisholmes:
LGTM

PR #154, opened 2016-06-30
3712: Allow flag to not be in config

phss:
Build is failing because 'allow_continue_on_failed_registration' is not
set in the verify-frontend-federation-config configuration. This change
allows for the flag not being set, returning empty array instead of nil.


davbo:
Why are we doing this rather than setting it to a value in verify-frontend-federation-config ?

config/initializers/rp_config.rb
```diff
@@ -1 +1 @@
-CONTINUE_ON_FAILED_REGISTRATION_RPS = RP_CONFIG.fetch('allow_continue_on_failed_registration')
+CONTINUE_ON_FAILED_REGISTRATION_RPS = RP_CONFIG.fetch('allow_continue_on_failed_registration', ifnone: [])
```
lloydnye: This will return a hash of `{ifnone => []}`
Try `RP_CONFIG.fetch('allow_continue_on_failed_registration', [])`

PR #155, opened 2016-07-01
3733 move age questions to start

leoshaw:
This adds support for a flag in the session which if set will show the
'Age' questions first.  By default it isn't set - the next release will
make it the default, but it's necessary to do it in two steps to avoid
breaking journeys that are in flight at the time of the roll-out.

### DIFF
--- a/spec/features/user_visits_about_choosing_a_company_page_spec.rb
+++ b/spec/features/user_visits_about_choosing_a_company_page_spec.rb
@@ -18,10 +18,10 @@ RSpec.describe 'When the user visits the about choosing a company page' do
     expect(page).to have_content 'Dod o hyd i’r cwmni iawn i’ch dilysu chi'
   end
 
-  it 'will take user to select documents page when user clicks "Continue"' do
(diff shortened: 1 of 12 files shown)


spec/features/user_visits_about_choosing_a_company_page_spec.rb
```diff
@@ -18,10 +18,25 @@
     expect(page).to have_content 'Dod o hyd i’r cwmni iawn i’ch dilysu chi'
   end
 
-  it 'will take user to select documents page when user clicks "Continue"' do
+  it 'will take user to will-it-work-for-me page when user clicks "Continue"' do
     visit '/about-choosing-a-company'
 
     click_link 'Continue'
-    expect(page).to have_current_path('/select-documents')
+    expect(page).to have_current_path(select_documents_path)
```
robinwhittleton: This path doesn’t match the test description.

robinwhittleton:
Looks good. Will merge when tests pass.

PR #158, opened 2016-07-05
3717: return IDPs from profile fitlers in sets

chrisholmes:
IDPs returned from profile filters are not being returned as a Set. This has caused
a problem for sessions from a RP where we do not want demos to be enabled as
the IDP was appearing twice as it had both demo and non-recommended profiles
defined so was introduced in both lists that were joined.


willp-bl:
LGTM 👍

PR #160, opened 2016-07-06
bau: use httprb 2.0.2

chrisholmes:
No major or breaking changes here, but should bring in a few performance
improvements


lloydnye:
Looks good, shame they aren't using semantic versioning or release notes.

PR #161, opened 2016-07-06
Special-case route for cycle3 further-information page

benhyland:
Since we are still in the middle of migrating cycle3 to
the new frontend, some tests break when run without
nginx to do redirection to the old frontend to get at
the unmigrated page.

Introduce an environment variable we can set in dev-land
to do the redirection for us.

We should introduce this redirect even though we are about to migrate the last bit of cycle3 since
we shouldn't leave tests broken for any significant length of time.

@benhyland and @christopherholmes

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -67,7 +67,13 @@ Rails.application.routes.draw do
   put 'select-idp', to: 'sign_in#select_idp_ajax', as: :select_idp_submit_ajax
   get 'service-status', to: 'service_status#index', as: :service_status
   get '/assets2/fp.gif', to: proc { |_| [200, {}, ['OK']] }
-  get 'further-information', to: 'further_information#index', as: :further_information
+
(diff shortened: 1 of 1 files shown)


lloydnye:
This makes the code behave differently in different environments, so a local test pass won’t give us confidence things are working in any other environment.

benhyland:
@lloydnye It's true a local pass won't give you confidence in the routing, but it will still give you confidence in the business logic from the controller on down.

Acceptance tests can't give you confidence in routing until they are run in a network with production-similar topology. Staging gives us this, I'm not sure to what extent joint does, and local definitely doesn't, even if the routes.rb had no conditional logic.

chrisholmes:
LGTM

I believe the benefits outweigh the temporary technical debt as the lack of local acceptance tests have caused a few avoidable problems.

PR #163, opened 2016-07-08
BAU: Automatically install dep changes in pre-commit

lloydnye:
This automates the manual step of installing changed dependencies,
saving a few back and forth commands. With no dependencies changed, it
doesn't add noticable time to the build.

Solo: @lloydnye


robinwhittleton:
👍🏻

PR #164, opened 2016-07-08
bau: Check for transaction simple id presence

chrisholmes:
We've had a few journeys  occur where after a user submits a duplicate SAML
message (perhaps after hitting the back and forwards button) they are treated
with an error message, but after a refresh are able to continue on for a few
pages despite their session missing a transaction simple id.

We want users to be see errors page as soon as they see an unhealthy state
so this PR introduces a check for the user's transaction simple as part of our
exisitng cookie validation. If the simple id is missing then the user will see
the standard something went wrong page.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -38,7 +38,7 @@ class ApplicationController < ActionController::Base
   end
 
   def validate_cookies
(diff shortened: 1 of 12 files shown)


willp-bl:
👍

PR #165, opened 2016-07-11
bau: don't send 404 messages to sentry

chrisholmes:
We currently hook up sentry to the raven logger to capture exceptions and error
messages.

Rails is currently logging 404s as errrors using the string representation of
the error (not just the exception).

As 404s are not something we can usually action this change will stop them from
being sent to Sentry.


willp-bl:
HTTP 202

PR #168, opened 2016-07-12
bau: only use minimal config for ZDD

chrisholmes:
We occasionally run into a problem in some of our environments where we're missing
an envvar required for the service thats causes our zdd rake tasks to fail (as
they were dependent on the full Rails environment).

This currently requires some manual intervention before we can fix the
environment. By only using config that is required for ZDD we can mitigate the
impact of missing config on our build pipeline and hopefully remove these manual
steps form our process.


willp-bl:
👍

PR #172, opened 2016-07-15
3795: refactor custom radios into a form builder helper

robinwhittleton:
Our custom radios are spread out through the codebase and require a few fixes and workarounds for less-common browsers that aren’t obvious. This PR refactors them out into a form builder helper.

Implementing these requires a form object, so the start form has been changed to a proper model as part of this work. This means that the parameter names have changed, which could be a ZDD issue. In preparation for this I changed the old start form such that it can [consume the new parameters](https://github.com/alphagov/verify-frontend/commit/2236e97a632218a8a1d3d6ca363e3faf964ca4db). This has already been deployed. After this PR is merged and deployed a clean-up commit should be made to [remove the code that handles the old form’s parameters](https://github.com/alphagov/verify-frontend/blob/359268a8430341d19de1a0390647afce61047d21/app/controllers/start_controller.rb#L9).

There are similar PRs for the perf and acceptance tests that will need to be merged to cope with the change to the start form parameters.


richardTowers:
💯  Great job!

PR #173, opened 2016-07-15
Remove Rails authenticity_token from SAML forms

davbo:
SAML forms generally only include SAMLRequest and RelayState, with the
exception of our extensions (registration, hints) we shouldn't go beyond
these standard form parameters.

Rails includes an authenticity_token to prevent cross-site request
forgery however in this case the form 'action' is a separate domain
beyond our control so csrf tokens like this are not useful.

Rails docs say the utf8 ✓

 > enforces browsers to properly respect your form's character encoding

Given the content of these forms is generated and not user submitted we
can be confident that we don't place any unicode characters in there.
Generally the bodies of these forms are base64 encoded XML so there is
very little room for unicode characters in any case.

### DIFF
--- a/app/views/redirect_to_idp/index.html.erb
+++ b/app/views/redirect_to_idp/index.html.erb
@@ -6,7 +6,7 @@
       <div class="application-notice info-notice">
         <p><%= t('hub.redirect_to_idp.description')%></p>
       </div>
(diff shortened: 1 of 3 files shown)


davbo:
Had some trouble with selenium locally so please wait for the travis build before merging..

PR #178, opened 2016-07-21
Stub api calls

richardTowers:
**Don't merge yet - wants discussion**

This moves the experimental `richardTowers/mock-verify-hub` code into verify-frontend proper. There are some basic unit tests which hopefully should be enough (just about) to prevent the stub-api from diverging from the real API. See the changes to the README to get an idea of how it would work.

The main reason we'd want to do this is to give non-verify developers an easy way to run the frontend on their machines. At the moment this isn't possible because (for now) there's still closed-source code running behind the API.

Feedback on whether this is the best way to do this / the right place for the code / a good idea in the first place welcome.

### DIFF
--- a/kill-service.sh
+++ b/kill-service.sh
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+if [ -a 'tmp/stub-api.pid' ]
+then
+  kill "$(cat tmp/stub-api.pid)"
+fi
+
 if [ -a 'tmp/puma.pid' ]
 then
   kill "$(cat tmp/puma.pid)"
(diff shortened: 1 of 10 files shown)


kill-service.sh
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+if [ -a 'tmp/stub-api.pid' ]
+then
+  kill "$(cat tmp/stub-api.pid)"
```
richardTowers: > Now if you shoot my dog, I'mma kill yo' cat
> Just the unwritten laws in rap, know dat

- Jay Z

lloydnye:
👍 for convenience

oswaldquek:
nice one

robinwhittleton:
I’m happy with this, and it’s absolutely a user need. Will merge.

PR #180, opened 2016-07-22
BAU: update JS to follow new GOV.UK module pattern

robinwhittleton:
[govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit) has a [new module pattern](https://github.com/alphagov/govuk_frontend_toolkit/pull/290). Our JS should match the standard where possible for simplicity. This commit updates govuk_frontend_toolkit to the latest version and brings our JS in line.

It also updates jasmine_rails to the latest point version, which fixes a potential memory leak issue.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,7 +83,7 @@ GEM
       govuk_frontend_toolkit (>= 4.12.0)
       rails (>= 4.1.0)
       sass (>= 3.2.0)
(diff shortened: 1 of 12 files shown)


richardTowers:
Yeah, big 👍 . I really didn't like the `.call(this)` style.

richardTowers:
Will merge once Travis finishes the PR.

PR #181, opened 2016-07-27
3843: Move service name out of the button

phss:
Feedback from services is that the button can be too big, depending on
the service name. 

Suggestion from UX folks is to replace the service name by 'Continue' on the button and introduce a new paragraph that informs the user can proceed to the service.

Authors: @adityapahuja @phss

### DIFF
--- a/spec/features/user_visits_confirmation_page_spec.rb
+++ b/spec/features/user_visits_confirmation_page_spec.rb
@@ -39,11 +39,6 @@ RSpec.describe 'When user visits the confirmation page' do
     expect(page).to have_css('html[lang=en]')
   end
 
-  it 'displays link button to Continue' do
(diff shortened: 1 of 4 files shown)


spec/features/user_visits_confirmation_page_spec.rb
```diff
@@ -38,15 +39,15 @@
     expect(page).to have_css('html[lang=en]')
   end
 
-  it 'displays link button with Transaction name' do
+  it 'displays link button to Continue' do
```
robinwhittleton: We can probably just dump this test as it’s implicitly tested in `sends user to response-processing page when they click the link`.

robinwhittleton:
LGTM, will merge when tests pass.

PR #183, opened 2016-07-28
bau: fix firefox instances starting up for all tests

richardTowers:
The check to force firefox in the feature tests had caused FF to start
on every test. This is super annoying, so we think we should just remove
the check for now.

Authors: @richardtowers @datamineruk

@chrisholmes @tunylund - discussion / premature merges welcome :trollface:


chrisholmes:
I don't think this error will occur often enough that it requires its own check that will impact the speed of tests therefore I am happy to remove it.

PR #184, opened 2016-07-28
BAU: update Ruby

robinwhittleton:
2.3.1 release notes: http://svn.ruby-lang.org/repos/ruby/tags/v2_3_1/ChangeLog

(also update libv8 to fix compilation on Clang 3.7.0 / OS X 10.11.4+)


robinwhittleton:
One thing that needs to be discussed: `Gemfile.lock` [still specifies a ruby version of 2.3.0](https://github.com/alphagov/verify-frontend/blob/bau-bump-ruby/Gemfile.lock#L331). I dug into why this wasn’t being updated and found [this issue](https://github.com/bundler/bundler/issues/4627) filed against bundler, which references [this PR](https://github.com/bundler/bundler/pull/4603) that unfortunately isn’t in a release version at the moment. Locally, rbenv uses the version defined in `.ruby-version` but I’m not sure what will happen on staging/production.

chrisholmes:
I've encountered a few problems when trying to build this using our packager job. I'll try and have a closer look later, but we'll need to hold on merging.

lloydnye:
@chrisholmes does your latest commit (use heroku buildpack directly) solve the packager issue you mentioned?

chrisholmes:
No, unfortunately. I'm going to be raising a story to address this problem formally.

I'm going to be closing this now if that's OK.

PR #189, opened 2016-08-02
bau: new frontend is not new anymore

willp-bl:
All user facing pages are now in this frontend, update the docs to
reflect that.

Pair: WillPa


richardTowers:
✨

PR #191, opened 2016-08-08
bau: Fix a test that seems to depend on resolution

richardTowers:
This tests was failing consistently for me when firefox was popping up
on my laptop screen. I had a play around with screen resolutions and it
looks like the test fails on low resolutions and passes on high
resolutions. I think this is due to a bug (?) in Capybara / Selenium
where it can't click on an element that is "below the fold".

This fix is definitely a hack. Better suggestions welcome.

Solo: @richardtowers

### DIFF
--- a/spec/feature_helper.rb
+++ b/spec/feature_helper.rb
@@ -16,6 +16,12 @@ if ENV['HEADLESS'] == 'true'
   end
 end
 
+RSpec.configure do |config|
+  config.before(:each, js: true) do
+    page.driver.browser.manage.window.resize_to(1280, 1024)
+  end
+end
+
 def current_time_in_millis
   DateTime.now.to_i * 1000
 end


vixus0:
Would it be better to set this globally for all browser-based tests? Bit crappy since it doesn't really involve responsiveness but I guess we're not testing that as such.

richardTowers:
Perhaps. I did try that first, but couldn't work out how to make it happen (I didn't try very hard).

Robin suggested to try upgrading Capybara, in case there's an upstream fix. I'll try that, and then think about moving this into a global location.

vixus0:
I believe you can put the window size stuff in a block like

    Before '@javascript' do
        page.driver.browser.manage.window.resize_to(1920, 1080)
    end

wherever we configure Capybara.

richardTowers:
Ok, I force pushed an update which will maximise firefox for all the JS tests. This also fixes the issue, but maybe it will be annoying having such big FF windows?

richardTowers:
Ok, now with 1280 x 1024. Should be long enough to avoid most fold issues without taking up massive expanses of the screen. It's probably a good idea to have a consistent size across machines anyway.

PR #193, opened 2016-08-09
BAU: Run stub-api on a different port to the real API

richardTowers:
This avoids problems where the real API tries to start on a port that's
already been grabbed by the stub API.

Solo: @richardtowers

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ You can start the application without having any of the closed source components
 
 `./startup.sh --stub-api`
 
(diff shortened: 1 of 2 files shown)


willp-bl:
50191 is the old fe/api admin port so could still have port issues... 50199?

richardTowers:
@willp-bl - yeah, @chrisholmes also spotted this. Me and Lloyd switched to 50199 in 150b093.

PR #195, opened 2016-08-11
bau: use middle to store session id for logging

chrisholmes:
The session is missing from some of our logs because it is only captured for
logstash when we reach the controller and after the CSRF check is performed.
This isn't helpful in some cases when trying to understand a user's journey and
session flows.

By switching to a middleware approach we can capture the session id before request
based logs are made.


lloydnye:
LGTM, been hoping we could improve this for a while

PR #197, opened 2016-08-11
bau: validate session before CSRF

chrisholmes:
We're encountering a few InvalidAuthenticityToken errors in production which are
raising errors in our logs. It appears that these errors are occuring because
somebody is submitted something when they don't have a valid session.

If we validate the session early then we should be able to remove these errors
and display something more appropriate for the user.


robinwhittleton:
👍

PR #199, opened 2016-08-15
bau: add session id to raven user context

chrisholmes:
When we send an error to sentry it would be helpful if we include the session id
as a parameter so that we can correlate it easily against other logs/errors.

Usually, we wouldn't have to do this ourselves with Raven as it usually includes
cookies with the error, but we have disabled this so that we don't leak the
user's secure cookie. To work around this we can set a user context containing
the session id.


willp-bl:
LGTM 👍

PR #200, opened 2016-08-15
bau: log session cookie size if it's large

willp-bl:
We store a lot of stuff in the session cookie. This commits add code
to log a warning if it's getting too large.  Without going too deep
into Rails we can log an approximation of the size, at a slightly
lower threshold.

Perhaps we can move this into the ApplicationController; I don't mind
too much as long as we can have some sort of visibility of this
becoming an issue.

Pair: WillPa

### DIFF
--- a/app/models/session_validator.rb
+++ b/app/models/session_validator.rb
@@ -5,7 +5,8 @@ class SessionValidator
       MissingCookiesValidator.new,
       SessionIdCookieValidator.new,
       TransactionSimpleIdPresence.new,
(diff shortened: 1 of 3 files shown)


richardTowers:
```
# --- Caused by: ---
     # NoMethodError:
     #   undefined method `flatten' for #<ActionDispatch::Request::Session:0x00000006c932d0>
     #   ./app/models/session_validator/cookie_size_validator.rb:5:in `validate'
```

PR #201, opened 2016-08-16
Upgrade to Rails 5

chrisholmes:
This should provide a working version of the FE under Rails 5 with minimal deprecation warnings

Most of the changes are contained within a single purpose commit message where the majority of changes were made to remove deprecation warnings.

One of the significant changes is that the stub api app has been isolated into its own directly and no longer shares a Gemfile with the FE. This was required as the Sinatra dependency of the stub has its own dependencies that conflict with the FE and Rails. It also provides us with a stricter separation of concerns which is lovely.

One change aside from basic renaming is that we have to use #add (which is additive) in place of #set (which replaces) when there are validation errors. I'm confident that our unit tests for ours validations are strong enough that this won't produce any funky behaviour.

We've switched our jasmine library because jasmine-rails was producing a lot of
deprecation warnings on rails 5. The new jasmine gem is maintained by the jasmine team
so this is probably the better long term option for updates/upgrades.


robinwhittleton:
Not sure if we want to specifically support that route, but removing Sinatra breaks the `--stub-api` startup flag.

chrisholmes:
@robinwhittleton: there's a commit that explicitly addresses [that](https://github.com/alphagov/verify-frontend/pull/201/commits/a2116895f0b1f0d1b5150a6d175b2e092a52b2d2)

As I stated above, I still need to write out why certain changes were made.

willp-bl:
This needs to be rebased, it would be good to get this merged soon

chrisholmes:
I've finally filled in my missing comments. I'm going to wait until other PRs are merged before rebasing hopefully for the last time.

chrisholmes:
OK, I think this is ready for a final look

richardTowers:
Looks pretty good to me. Had some deprecation warnings when running pre-commit:

```
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from load at /Users/richardtowers/.rbenv/versions/2.3.0/bin/rake:23)
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from load at /Users/richardtowers/.rbenv/versions/2.3.0/bin/rake:23)
DEPRECATION WARNING: ActionDispatch::IntegrationTest HTTP request methods will accept only
the following keyword arguments in future Rails versions:
params, headers, env, xhr, as

Examples:

get '/profile',
  params: { id: 1 },
  headers: { 'X-Extra-Header' => '123' },
  env: { 'action_dispatch.custom' => 'custom' },
  xhr: true,
  as: :json
 (called from load at /Users/richardtowers/.rbenv/versions/2.3.0/bin/rake:23)
```

Is that expected?

I also tested the stub-api thing, and that looks to be working correctly :)

chrisholmes:
Those should have been removed with the jasmine change. 

Can you check that you're running the latest version of the branch and report back?

richardTowers:
Yeah, sorry, I hadn't fetched 😧.

On the latest version pre-commit was failing with `Don't know how to build task 'spec:javascripts'`. I pushed a commit to fix that (should be `jasmine:ci`).

robinwhittleton:
Checked by @chrisholmes / @willp-bl / @robinwhittleton / @richardTowers.

willp-bl:
🚈 🚈 :shipit:

PR #203, opened 2016-08-19
5078: make link buttons behave like buttons to assistive tech

robinwhittleton:
All links styled to look like buttons should have a `role="button"` attribute to correctly define to assistive tech what the expectation of their use is. To make sure that they behave as buttons to the user, we also add the spacebar-to-activate behaviour with Javascript.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,7 +83,7 @@ GEM
       govuk_frontend_toolkit (>= 4.14.1)
       rails (>= 4.1.0)
       sass (>= 3.2.0)
-    govuk_frontend_toolkit (4.14.1)
+    govuk_frontend_toolkit (4.16.1)
       rails (>= 3.1.0)
       sass (>= 3.2.0)
     govuk_template (0.18.0)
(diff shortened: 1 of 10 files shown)


lloydnye:
Looks good to me

lloydnye:
As spacebar to activate wouldn't be added in a non js environment, would it be a good idea to use buttons for non js to get the behaviour and replace them with links via js?

robinwhittleton:
We’d have to replace every link with a `<form action="…" method="get"><button>Continue</button></form>` and that breaks behaviour like showing the link target in an overlay (there would also be no point in replacing them with links with JS in that situation). This enhances the current system for most people and doesn’t at least leave it any worse for the rest.

richardTowers:
@lloydnye, or just use buttons everywhere, but that would mean:

```
<form method=GET action=/blah>
<button>link text</button>
</form>
```

Which feels horribly verbose compared to `<a href=/blah>...`

I think this is a step in the right direction, so I'm going to go ahead and merge.

PR #204, opened 2016-08-19
5078: make sure that link buttons have roles

robinwhittleton:
Creating a `button_link_to` helper reduces the chance that link buttons won’t have the correct role applied. If a `class` attribute is supplied it’ll append `button` to the existing one, but we’d only ever want a `role="button"` so that is overridden. Other parameters are passed straight through.


willp-bl:
LGTM 👍

PR #205, opened 2016-08-22
BAU: Cookie mismatch is a bad request

lloydnye:
Discussion in code quality meeting came to consensus that this is a client error so should have a 400 series http response.

Solo: @lloydnye


lloydnye:
@SrinivasMurty Stopping the cookie mismatch from showing up as a 500.

willp-bl:
:ok

PR #209, opened 2016-08-26
Set a basic, lenient CSP header

davbo:
Content-Security-Policy (CSP) is applied to restrict browser behavior to
the features used within the verify-frontend.

This policy is purposefully quite lenient, allowing any inline-scripts
to execute. We intend to add a reporting-uri in the future and iterate
our policy to further restrict browser behavior.

Specifically this policy only allows resources to be fetched from
'self' (origin of the document being served) it further restricts the
fonts to load using the 'data' scheme as they are provided from the
frontend_toolkit. We disallow embedding any object/applets. Scripts can
be loaded from self or as inline html, the same policy is applied to
stylesheets.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -33,8 +33,10 @@ module VerifyFrontend
 
     config.exceptions_app = self.routes
 
-    #remove default rails headers as they are added by reverse proxy
-    config.action_dispatch.default_headers.clear
+    # Apply a basic lenient Content Security Policy
+    config.action_dispatch.default_headers = {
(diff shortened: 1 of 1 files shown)


davbo:
Immediately re-reading this and wondering why I chose to characterize this as simple...

robinwhittleton:
Interesting that the spec suggests that `data:` [is a security risk](https://www.w3.org/TR/CSP/#directives). Although not having a CSP at all is worse.

davbo:
That makes sense in the context of script-src where allowing data: is the
same as an inline script but I don't see a risk allowing it just for fonts.

Also keep in mind this is just a start on setting a CSP. With reporting
we'll be and to make it much more strict on script inclusion.

willp-bl:
Looks good to me, need to give it some thorough testing in a non-prod environment

richardTowers:
:shipit:

PR #210, opened 2016-08-30
5026: update cookie notice

chrisholmes:
the cookie notice must be changed to reflect the removal of the secure cookie.

@gdsamit has reviewed the changes and has given it an OK.

### DIFF
--- a/app/views/static/cookies.cy.html.erb
+++ b/app/views/static/cookies.cy.html.erb
@@ -34,18 +34,13 @@ end %>
   </thead>
   <tbody>
     <tr>
-      <td><var>x-govuk-secure-cookie</var></td>
-      <td>When you’re trying to use a service on GOV.UK, this cookie makes sure that no other user can login with your information</td>
-      <td>When you close your browser window</td>
-    </tr>
-    <tr>
       <td><var>x_govuk_session_cookie</var></td>
       <td>This cookie contains randomly generated characters so we can identify you during your visit</td>
       <td>When you close your browser window</td>
     </tr>
     <tr>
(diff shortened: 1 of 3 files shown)


robinwhittleton:
Looks fine, will merge when tests finish.

app/views/static/cookies.cy.html.erb
```diff
@@ -34,18 +34,13 @@ end %>
...
-    </tr>
-    <tr>
       <td><var>x_govuk_session_cookie</var></td>
       <td>This cookie contains randomly generated characters so we can identify you during your visit</td>
       <td>When you close your browser window</td>
     </tr>
     <tr>
       <td><var>_verify-frontend_session</var></td>
-      <td>This cookie is used to hold information regarding your session, including when you started your session, which government service you’re using, your answers to filter questions, and which certified company you select so that we can present information relevant to your choices. This information is secured so that only GOV.UK Verify can read it</td>
+      <td>When you’re trying to use a service on GOV.UK, this cookie makes sure that no other user can login with your information. It is also used to hold information regarding your session, including when you started your session, which government service you’re using, your answers to filter questions, and which certified company you select so that we can present information relevant to your choices. This information is secured so that only GOV.UK Verify can read it</td>
```
willp-bl: GOV.UK -> GOV.UK Verify

chrisholmes: It's GOV.UK because the service in this case is the Relying Party

gdsamit:
I think we need to say "log in" not "login" as we're using it as a verb not a noun.

chrisholmes:
amended!

PR #211, opened 2016-08-30
5124: update the fingerprint library

willp-bl:


robinwhittleton:
So either PR #212 will need to be rebased once this has merged or vice versa.

PR #212, opened 2016-08-30
BAU: move inline Fingerprint code to external asset

robinwhittleton:
For our CSP work we want to minimise the use of inline scripts so that we can have a small whitelist. The fingerprinting code can happily live externally so this PR moves it and rewrites the tests for the new scenario.


chrisholmes:
LGTM, but on a partially unrelated note is FINGERPRINT_CONFIG really config if it's a constant value?

willp-bl:
I'll change it to FINGERPRINT_PATH in the other open fingerprint PR when I rebase

PR #216, opened 2016-08-31
bau: tune CSP directive

willp-bl:
We were seeing an issue with the existing CSP directive in staging,
this was due to our piwik.js loading an image from a data: URI which
was disallowed.

This fix has been tested locally in both Chrome and Firefox with
tweaked PIWIK_HOST to point to the same host/port as the frontend (as
happens in prod).  When running locally with piwik on a different port
etc the CSP difrective will still be violated => we should consider
what to do with CSP when running locally.

Pair: WillPa

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -35,7 +35,7 @@ module VerifyFrontend
 
     # Apply a basic lenient Content Security Policy
     config.action_dispatch.default_headers = {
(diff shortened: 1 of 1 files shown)


willp-bl:
Whilst this worked locally, it did not in a real environment. Closing.

PR #217, opened 2016-09-01
5057: Revert all commits adding "(position x of n)"

richardTowers:
This was needed to gather data on whether IdPs at the top of the list
are selected more often than those further down. We have enough data to
be confident that it does now, so this extra logging is no longer
required.

We'd leave it in place, but having such a large number of permutations
for this virtual page title confuses the piwik UI, and that will be
annoying for the analytics team. More seriously, the page title we
modified is used for calculating the success rate in demo periods. We're
not sure if the change might have skewed the results a bit, so we'd like
to go back to the old format for the next demo period.

If we need similar behaviour in the future (which we probably will) we
should have a bit more of a think about where to put it and what the
effect on the piwik UI will be. Considerations:

* It would be nice not to have to store anything in flash in the code.
  We could just as easily log this page title on the picker page, not
  the redirect-to-idp-warning page.
* Page titles with a high number of permutations will mess up the piwik UI

Also note: once this commit merges to master we will need to make the
corresponding change to ida-hub-acceptance-tests to stop asserting that
the position is logged.

This commit reverts the following:

* "5057: Rename index to position" (112e42ae17636307bc848612b2c3b69ae05f5619)
* "5057: cache the result of grouped_identity_providers" (9b90f0bb97b3dbbdc14af65b01bf2e472e55fb6a)
* "5057: Use `Integer(thing)` instead of `thing.to_i if thing =~ /\d+/" (9f9293a859c982e2a996ec855c533dae2c79530a)
* "5057: Add flash.keep to RedirectToIdpWarningController#index" (fe2292439095254123c14a38e6dbe673253952da)
* "5057: Refactor feature tests check for registration piwik report" (6df98338cb28c23bb0716119c7778a15952b47f0)
* "5057: Add number of idps to idp selected piwik report" (eacc5879546803af904713bebc6e7edc4328430e)
* "5057: Send position of chosen IdP to piwik" (c7611af1abf2d044452cd39fd3e7ae0c84904fef)

Solo: @richardtowers

### DIFF
--- a/app/controllers/choose_a_certified_company_controller.rb
+++ b/app/controllers/choose_a_certified_company_controller.rb
@@ -1,17 +1,14 @@
 class ChooseACertifiedCompanyController < ApplicationController
   def index
+    grouped_identity_providers = IDP_RECOMMENDATION_GROUPER.group_by_recommendation(selected_evidence, current_identity_providers, current_transaction_simple_id)
     @recommended_idps = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(grouped_identity_providers.recommended)
     @non_recommended_idps = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(grouped_identity_providers.non_recommended)
   end
 
   def select_idp
     select_viewable_idp(params.fetch('entity_id')) do |decorated_idp|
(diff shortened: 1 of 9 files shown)


richardTowers:
When this merges we'll also need to merge ida-hub-acceptance-tests/pull/155 on the internal github.

PR #219, opened 2016-09-02
BAU: stop Rails 5 automatically disabling submit buttons

robinwhittleton:
Rails 5 automatically disables submit buttons after they’ve been clicked on once. If you go back on our pages it remembers the disabled state, thus breaking the system. We can turn this functionality off globally.


robinwhittleton:
This was added to Rails with this PR: https://github.com/rails/rails/pull/21135

willp-bl:
Looks good 👍

PR #220, opened 2016-09-02
BAU: allow for changes in content without introducing new keys

robinwhittleton:
In several content sections individual paragraphs are being internationalised. If in the future we change the number of paragraphs we’d have to rework the keys in the localisation. The current method also fixes both translations to the same number of paragraphs, which might not be ideal.

This commit migrates these content sections into extended _html sections in the YAML. It also changes a spec that was testing for text node equality: as Capybara’s output for the page lacks HTML tags the content of the YAML file wasn’t matching.


willp-bl:
Looks good

PR #223, opened 2016-09-09
BAU: point cookies message at Verify

robinwhittleton:
The current cookies banner points at GOV.UK’s cookies page, but should point at Verify’s. This also enables Welsh translation.


willp-bl:
👍

PR #224, opened 2016-09-13
bau: return head 404 instead of redirection

oswaldquek:
It is possible that some assets on the page might request invalid resources as
well. In this case it's better to return a head 404 rather than redirect.

@oswaldquek


richardTowers:
👍 this feels like a better solution

PR #229, opened 2016-09-21
BAU: Initialisers refactor (2nd attempt)

vixus0:
- Consolidate small but related initialisers.
- Keep the default Rails skeleton initialisers for easier tracking
  through Rails versions.
- Some typo and code-style changes.

Author: @vixus0


config/initializers/constants.rb
```diff
@@ -0,0 +1,4 @@
+FINGERPRINT_CONFIG = '/assets2/fp.gif'.freeze
+SESSION_VALIDATOR = SessionValidator.new(Integer(CONFIG.session_cookie_duration))
+SAML_PROXY_HOST = CONFIG.saml_proxy_host
+METADATA_CLIENT = MetadataClient.new(SAML_PROXY_HOST)
```
chrisholmes: I think the old arrangement for these constants is more appropriate as these things don't belong together.

vixus0:
This gets rid of some of the constants but creating things in ApplicationController might not be what we want to do.

app/controllers/application_controller.rb
```diff
@@ -98,6 +98,10 @@ def session_validator
     SESSION_VALIDATOR
   end
 
+  def metadata_client
+    METADATA_CLIENT
+  end
+
```
chrisholmes: I think this can be moved to the MetadataController

chrisholmes:
Happy to merge after my last comment is addressed!

vixus0:
Done!

PR #230, opened 2016-09-28
BAU: cope with broken JS in some situations

robinwhittleton:
[govuk_template 0.18.2](https://github.com/alphagov/govuk_template/releases/tag/v0.18.2) adds the ability for the styling to recover in the event that external JS has failed to load. These scenarios are rare but potentially problematic.


willp-bl:
👍

PR #233, opened 2016-10-04
Bau: Add windows support

niyioyelade:
Puma doesn't support multiple workes on windows
Gem config file needed 3 extra dependencies for windows

Authors: @niyi @richardtowers


richardTowers:
Closing this, as it looks like travis needs an updated Gemfile.lock to work properly. I'll recreate this branch locally and run a bundle install, then recreate the PR.

PR #234, opened 2016-10-04
Bau: Improve windows support

richardTowers:
Puma doesn't support multiple workers on windows. Gem config file needed
3 extra dependencies for windows.

Puma also doesn't support being run in daemon mode, which means that
./startup.sh doesn't work. It's possible to work around this by manually
running puma without the daemonize flag.

Authors: @niyi @richardtowers

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,7 +84,7 @@ gem 'request_store', '~> 1.3.1'
 gem 'zendesk_api'
 gem 'email_validator'
 
-platforms :x64_mingw_23 do
+platforms :mswin, :mingw, :x64_mingw do
   gem 'windows-pr'
   gem 'win32-process'
   gem 'tzinfo-data'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -83,3 +83,9 @@ gem 'logstash-logger'
 gem 'request_store', '~> 1.3.1'
 gem 'zendesk_api'
 gem 'email_validator'
+
+platforms :x64_mingw_23 do
```
willp-bl: :( that's not confusing at all

Reading that doc, seems we could just use :x64_mingw ?

richardTowers: We tried using `mswin` first, but that didn't seem to match (perhaps because we were using bash on windows). Perhaps `mingw` or `x64_mingw` would work though, and it would be more future proof I suppose.

willp-bl:
Looks good

PR #235, opened 2016-10-05
bau: Add bundle install to startup.sh

richardTowers:
This adds a few 100ms to the time taken to run the script, but it
ensures that you don't get the horrible "you need to run bundle install"
error.

Authors: @richardtowers


willp-bl:
bundle exec merge

richardTowers:
`bundle exec thanks`

PR #237, opened 2016-10-14
5065: Use Heroku Ruby buildpack

maurorappa:
Make pkgr use the current Heroku Ruby buildpack
Update postinst.sh to allow deployer to access files under
/opt/front/vendor/bundle as these files are owned by root

Authors: @adityapahuja @maurorappa @mmhmm

### DIFF
--- a/packaging/postinst.sh
+++ b/packaging/postinst.sh
@@ -13,3 +13,6 @@ chown -R deployer:deployer /opt/front/log
 chown -R deployer:deployer /var/log/front
 chown -R deployer:deployer /opt/front/tmp
 chgrp deployer /etc/front
+
+# deployer needs to access to all those files under bundle which are owned by root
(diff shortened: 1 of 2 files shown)


packaging/postinst.sh
```diff
@@ -13,3 +13,6 @@ chown -R deployer:deployer /opt/front/log
 chown -R deployer:deployer /var/log/front
 chown -R deployer:deployer /opt/front/tmp
 chgrp deployer /etc/front
+
+# deployer needs to access to all those files under bundle which are owned by root
+find /opt/front/vendor/bundle -exec chmod 755 {} \;
```
jinnko: This is a very broad setting of permissions.  Do you genuinely want to make everything executable?  Perhaps you'd like to set the folders to be readable and the files to be readable?  Perhaps something like this would be more suitable?

```
find /opt/front/vendor/bundle \( -type d -exec chmod go+rx {} \; \) , \( -type f -exec chmod go+r {} \;  \)
```

maurorappa: I agree, I will try your find magic

PR #238, opened 2016-10-14
5175: Send not_ukphotolicense as well...

oswaldquek:
...when user has selected NO to GB licence. This is to ensure backwards
compatability as not_ukphotolicense is currently being sent in this case.

solo: @oswaldquek


willp-bl:
Looks good, will merge when Travis is finished

PR #240, opened 2016-10-25
Update shared GOV.UK code

robinwhittleton:
We’re a few versions behind on the shared frontend code and the linter. This bumps to the latest versions, and fixes a bug with Safari and the skip links. Pre-commit passes and I’ve run a quick visual sweep through the system, but a second pair of eyes never hurts.

Specific changes in the new code that affect Verify:
- external link icons are gone
- extraneous copy of HTML5shiv that wasn’t being used is removed
- fix a potential bug with `<details>` elements


richardTowers:
`./pre-commit.sh` passes on my machine also, and a quick manual run through didn't throw anything up. I'd be happy for this to merge :)

richardTowers:
In fact, I'm going for it.

PR #243, opened 2016-10-31
bau: add csp checksums

willp-bl:
Also add script to generate checksums of inline JS, and add them to the
policy.

Tested with reporting in latest Firefox, Chrome & Safari

Pair: WillPa / DavidK

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -39,7 +39,16 @@ module VerifyFrontend
       'X-Frame-Options' => 'DENY',
       'X-XSS-Protection' => '1; mode=block',
       'X-Content-Type-Options' => 'nosniff',
-      'Content-Security-Policy' => "default-src 'self'; font-src data:; img-src 'self'; object-src 'none'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; style-src 'self' 'unsafe-inline'"
+      'Content-Security-Policy' => "default-src 'self'; " +
+                                    "font-src data:; " +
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -39,7 +39,12 @@ class Application < Rails::Application
       'X-Frame-Options' => 'DENY',
       'X-XSS-Protection' => '1; mode=block',
       'X-Content-Type-Options' => 'nosniff',
-      'Content-Security-Policy' => "default-src 'self'; font-src data:; img-src 'self'; object-src 'none'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; style-src 'self' 'unsafe-inline'"
+      'Content-Security-Policy' => "default-src 'self'; "+
+                                    "font-src data:; " +
+                                    "img-src 'self'; " +
+                                    "object-src 'none'; " +
+                                    "script-src 'self' 'unsafe-eval' 'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=' 'sha256-G29/qSW/JHHANtFhlrZVDZW1HOkCDRc78ggbqwwIJ2g=' 'unsafe-inline'; " +
```
robinwhittleton: Is it worth adding a comment about where these come from and how to generate new versions?

willp-bl: good point, added a comment that has details, and I've removed the script that was added

PR #244, opened 2016-11-01
bau: route test analytics and test csp reporter via frontend

willp-bl:
CSP errors were happening locally because Piwik requests were not
being routed via the frontend. Therefore, rewrite requests through to
local Piwik when running locally.

This also adds a test CSP reporter that can be turned on to view what
browsers are sending.  It's not enabled by default; see how to enable
it in routes.rb.  Interesting to note that we have to do this because
Firefox requires the CSP reporter to be on the same host/port as the
site.  Safari and Chrome are ok with it being a different host/port.

Also fixes the linting issues that I should have fixed before the last
PR :(

Pair: WillPa


lloydnye:
LGTM

Issue #213, opened 2016-08-30
Verify radio and checkboxes don't visually change if user has custom colours


edwardhorsford:
If a user has chosen custom colours (for accessibility / visibility reasons), the Verify checkboxes and radios do not visually change when selected.

Video of the issue:
![verify-custom-colours-custom-radios](https://cloud.githubusercontent.com/assets/2204224/18093449/c537d168-6ec7-11e6-8064-dc5c68c981a0.gif)

What's wrong / expected:
The radios and checkboxes should visually change when selected.

Tested on:
Firefox with custom colours picked (in settings)
Verify production environment.

Tried after reading this blog post:
https://accessibility.blog.gov.uk/2016/05/26/accessibility-and-me-marian-foley/

robinwhittleton:
Hopefully this will be resolved by fixing https://github.com/alphagov/govuk_elements/pull/296, cutting a new release of elements and integrating that into Verify. I’ve been using Verify as a test-bed, so that shouldn’t be too difficult.

edwardhorsford:
I imagine that elements PR will want to be heavily tested, which may take some time...

PR #249, opened 2016-11-18
Use new radio buttons and checkboxes from GOV.UK elements

robinwhittleton:
This removes the existing Verify radio buttons and checkboxes implementation and replaces it with the [now standard GOV.UK version](http://govuk-elements.herokuapp.com/form-elements/). We’d previously not done this as we found our version tested better than the old implementation, but the new one is based on the same original designs as Verify and has had considerably more compatibility work put in.

Overall the buttons match the current visual style, albeit slightly larger. In the original elements design there was a grey background and padding away from the edge. This padding remained in the original Verify implementation, but now isn’t needed so some of the spacing has been decreased.

docs

The most major change is on the start page. We’ve built the new elements inputs to work regardless of user choices of background and colour (see [Marian Foley’s blog post](https://accessibility.blog.gov.uk/2016/05/26/accessibility-and-me-marian-foley/)). We decided that the best way to do this was to purely use the text colour to define the colours of the input and to leave the background transparent. This is a reasonably big change for Verify and it might need some user testing before we’re satisfied that it’s the right thing to do.

start

Fixes https://github.com/alphagov/verify-frontend/issues/213


robinwhittleton:
Forgot to mention, this also potentially removes some CSP over-reporting by removing the `onclick=""` hack we had in there for old iOS. That’s no longer necessary with the new code.

robinwhittleton:
These now match the old visual design on the start page:

screen shot 2016-12-01 at 09 38 33

mikieet:
This all looks good from a design point of view

richardTowers:
:shipit:

willp-bl:
Minor nit, seems slightly oval shaped?
screen shot 2016-12-01 at 09 59 22

Plus occasionally seems like there's a bit of lag, but was that the thing discussed here: https://github.com/alphagov/govuk_elements/pull/363 ?

But, looks good to me :shipit:

robinwhittleton:
Lag should have been fixed by https://github.com/alphagov/verify-frontend/pull/249/commits/825e2b8beb0ba76c83a0e082b15525f980bb78f1, so if you’re seeing it still I’d be interested in finding out where.

If that turns out to be nothing I’m happy for this to ship now.

robinwhittleton:
Re oval, I think that’s your eyes. There’s an open [googly eyes bug](https://github.com/alphagov/govuk_elements/issues/354) but that doesn’t seem to be the case here 👀

richardTowers:
@willp-bl 's image looks squashed to me too, but I can't reproduce the issue in any of the browsers on my machine at any zoom level. Maybe it's an optical illusion?

willp-bl:
Chrome says it's 38x38, but it still seems elongated/oval shaped.  Might help to see it if you turn the screen 90 degrees?  Perhaps it is just an artefact of how it is being drawn?

vixus0:
Just to be a bit more scientific, I've modified the screenshot from @willp-bl:

Radio button pedantry

There does appear to be a pixel missing from either side of the unselected radio button. Doesn't seem to happen in my browser though.

robinwhittleton:
Huh. I'm off today but maybe you could file a bug against govuk_elements? Will look at it next week.

robinwhittleton:
Seeing as we’re having difficult replicating this problem and it’s a minor visual bug can I suggest we merge this PR anyway? I’ll leave the [elements issue](https://github.com/alphagov/govuk_elements/issues/369) open and keep on investigating.

PR #250, opened 2016-12-02
5226: Raise an error on startup if missing a secret

tunylund:
There was a short downtime on verify frontend due to a missing
configuration that needs to be put in place by hand. The error could
have been noticed sooner if the application would have raised an error
on the missing configuration.

Solo: @tunylund


richardTowers:
LGTM. Merged.

PR #251, opened 2016-12-09
5482: AB test for IDP rankings

phss:
Adds a ranking step to the Choose a Certified Company page. The ranking
is defined in verify-frontend-federation-config.

NOTE: we are still waiting for Olly's ranking data. Therefore we are
not sending requests to piwik!

We have test data in place to test it only in the Joint
environment.

Authors: @srinivasmurty @phss

### DIFF
--- a/spec/models/display/identity_provider_display_decorator_spec.rb
+++ b/spec/models/display/identity_provider_display_decorator_spec.rb
@@ -79,7 +79,9 @@ module Display
 
         result = decorator.decorate_collection_with_ranking(idp_list, ranking)
 
-        expect(result.map(&:simple_id).first).to eql 'idp_three'
+        idps = result.map(&:simple_id)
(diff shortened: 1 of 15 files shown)


spec/models/display/identity_provider_display_decorator_spec.rb
```diff
@@ -44,9 +45,54 @@ module Display
...
+
+        expect(result.map(&:simple_id)).to eql %w(idp_two idp_one idp_three)
+      end
+
+      it 'returns a decorated IDP collection based on ranking only one rank' do
+        ranking = Display::IdpRanking.new(['idp_three'])
+
+        result = decorator.decorate_collection_with_ranking(idp_list, ranking)
+
+        expect(result.map(&:simple_id).first).to eql 'idp_three'
```
oswaldquek: should this test test that the other idps are returned too?

PR #252, opened 2016-12-15
5090: Add proxy config option for zendesk

lloydnye:
We want to be able to make external calls via a proxy.
This change allows us to set a proxy for zendesk connections.
The value is optional so no existing configuration will continue to work.

Solo: lloydnye


willp-bl:
👍

PR #261, opened 2017-03-02
bau: Fix kill-service.sh to kill the stub api properly

richardTowers:
pid file is stub_api.pid, not stub-api.pid.

Also removes a useless use of cat (replacing it with a probably useless
use of xargs, but hey-ho).

Authors: @richardtowers


adityapahuja:
Looks good to me.

PR #266, opened 2017-03-21
Upgrade to elements 3

robinwhittleton:
(first off, sorry for the chunky PR)

This updates to GOV.UK Elements 3.0.1 ([changelog](https://github.com/alphagov/govuk_elements/blob/0195c6a76287584dc84415b9c840ac14690cfa1c/CHANGELOG.md)) and GOV.UK Frontend Toolkit 5.2.1 ([changelog](https://github.com/alphagov/govuk_frontend_toolkit/blob/f9fc75a1252805b2fa28bac149e49ad28ac9b545/CHANGELOG.md)). There are a couple of minor accessibility improvements but the big changes are the removal of the JavaScript requirement for radio buttons and checkboxes, and the change in the class applied to form groups and controls that are invalid.

This change has necessitated reworking the markup for radios and checkboxes (along with the error class changes). Along the way I noticed a few places where common markup could be refactored so there are a couple of commits to address that. Hopefully the commits should be fairly readable in order, but if there are any questions then I’m happy to answer.


app/views/feedback/index.html.erb
```diff
@@ -5,11 +5,7 @@
   <div class="column-two-thirds">
     <h1 class="heading-xlarge"><%= t 'hub.feedback.heading' %></h1>
 
-    <% if flash[:errors] %>
-        <div id="validation-error-message-no-js" class="validation-message form-group">
-          <p class="error"><%= flash[:errors] %></p>
-        </div>
-    <% end %>
+    <%= render 'shared/form-errors', errors: flash[:errors] %>
```
oswaldquek: @robinwhittleton just to make sure, it's `shared/form-errors` and not `shared/_form-errors`?

robinwhittleton: Yep. References to partials in Rails omit the underscore: http://guides.rubyonrails.org/layouts_and_rendering.html#naming-partials

>Note the leading underscore character: partials are named with a leading underscore to distinguish them from regular views, even though they are referred to without the underscore.

oswaldquek:
test this please

PR #267, opened 2017-03-22
Upgrade to Elements 3

robinwhittleton:
(first off, sorry for the second attempt and chunky PR)

This updates to GOV.UK Elements 3.0.1 (changelog) and GOV.UK Frontend Toolkit 5.2.1 (changelog). There are a couple of minor accessibility improvements but the big changes are the removal of the JavaScript requirement for radio buttons and checkboxes, and the change in the class applied to form groups and controls that are invalid.

This change has necessitated reworking the markup for radios and checkboxes (along with the error class changes). Along the way I noticed a few places where common markup could be refactored so there are a couple of commits to address that. Hopefully the commits should be fairly readable in order, but if there are any questions then I’m happy to answer.


oswaldquek:
looking

app/assets/stylesheets/_govuk-elements.scss
```diff
@@ -17,7 +17,7 @@
 // Mixins to generate components (chunks of UI)
 // @import "design-patterns/alpha-beta";
 @import "design-patterns/buttons";
-@import "design-patterns/breadcrumbs";
+// @import "design-patterns/breadcrumbs";
```
oswaldquek: is there a reason why this line is commented out and not deleted?

oswaldquek:
generally LGTM!

app/assets/stylesheets/_govuk-elements.scss
robinwhittleton: `_govuk-elements.scss` is provided by GOV.UK Elements (https://github.com/alphagov/govuk_elements/blob/8262e18a0cfa12128ed6a4399f914841ce3ce369/public/sass/_govuk-elements.scss). By simply commenting out the bits that we don’t want it’s more obvious when you do a comparison in a normal editor to see the differences when upgrades happen. Obviously if you’re using something like FileMerge that’s different, but for simplicity’s sake this works.

PR #268, opened 2017-03-24
TT-246: Fix feedback page routing when switching to welsh

oswaldquek:
This commit fixes the following issue:

1. Start at verify product page
2. Click link to send feedback
3. At feedback page, change page to be in Welsh
4. Enter data and send feedback
5. User sees "Session timed out" instead of "Return to the GOV.UK product page"

On investigation it appears this issue isn't just related  to TT-246 but for
all feedback pages in general. When the user goes to a feedback page there is a
feedback source query parameter in the URL but this feedback source disappears
when the language is changed.

DEPRECATED:
To fix this, this commit introduces a feedback_source_query_param method in
application_helper which is made use of in
app/views/shared/_available_languages.html.erb. This doesn't break the language
link on non-feedback pages as the url_for method doesn't place a
feedback-source query param if none exists.

CURRENT:
To fix this, this commit uses Rails' flash to preserve the feedback-source.

@oswaldquek

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -4,7 +4,8 @@ class FeedbackController < ApplicationController
   def index
     @form = FeedbackForm.new({})
     flash['feedback_referer'] = request.referer
-    feedback_source = params['feedback-source']
+    feedback_source = params['feedback-source'].nil? ? flash['feedback_source'] : params['feedback-source']
+    flash['feedback_source'] = feedback_source
     if feedback_source.nil?
       render
     elsif FEEDBACK_SOURCE_MAPPER.is_feedback_source_valid(feedback_source)
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -4,7 +4,14 @@ class FeedbackController < ApplicationController
   def index
     @form = FeedbackForm.new({})
     flash['feedback_referer'] = request.referer
-    flash['feedback_source'] = params['feedback-source']
+    feedback_source = params['feedback-source']
+    if feedback_source.nil?
+      render
+    elsif FEEDBACK_SOURCE_MAPPER.is_feedback_source_valid(feedback_source)
+      flash['feedback_source'] = feedback_source
+    else
+      render 'errors/404', status: 400
+    end
   end
 
   def submit
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -5,7 +5,6 @@ class FeedbackController < ApplicationController
     @form = FeedbackForm.new({})
     flash['feedback_referer'] = request.referer
     feedback_source = params['feedback-source'].nil? ? flash['feedback_source'] : params['feedback-source']
-    flash['feedback_source'] = feedback_source
     if feedback_source.nil?
       render
     elsif FEEDBACK_SOURCE_MAPPER.is_feedback_source_valid(feedback_source)
--- a/app/models/feedback_source_mapper.rb
+++ b/app/models/feedback_source_mapper.rb
@@ -9,6 +9,7 @@ class FeedbackSourceMapper
       'CERTIFIED_COMPANY_UNAVAILABLE_PAGE' => 'choose_a_certified_company',
       'CONFIRM_YOUR_IDENTITY' => 'confirm_your_identity',
       'CONFIRMATION_PAGE' => 'confirmation',
+      'ERROR_PAGE' => 'start',
       'FAILED_REGISTRATION_PAGE' => 'failed_registration',
       'FAILED_SIGN_IN_PAGE' => 'failed_sign_in',
       'CYCLE_3_PAGE' => 'further_information',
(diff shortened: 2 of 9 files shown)


oswaldquek:
DO NOT MERGE YET - waiting on Welsh translation from Srini, and also keen to get people's thoughts on implementation, since iamnotarailsexpert

app/views/shared/_available_languages.html.erb
```diff
@@ -5,14 +5,14 @@
       <% if params[:locale] == 'en' %>
         <span>English</span>
       <% else %>
-        <%= link_to 'English', url_for(locale: 'en'), lang: 'en', hreflang: 'en' %>
+        <%= link_to 'English', url_for(locale: 'en', :'feedback-source' => feedback_source_query_param), lang: 'en', hreflang: 'en' %>
```
willp-bl: we shouldn't be re-using unsanitised user input

app/views/feedback_sent/index.html.erb
```diff
@@ -10,7 +10,7 @@
 
   <p>
     <%if @from_product_page%>
-        <%= link_to 'Return to the GOV.UK Verify product page', @link_back %>
+        <%= link_to t('hub.feedback_sent.product_page'), @link_back %>
```
willp-bl: does this need to be localised? the product page is currently only in English

app/views/shared/_available_languages.html.erb
```diff
@@ -5,14 +5,14 @@
...
       <% end %>
     </li>
     <li class="translation last">
       <% if params[:locale] == 'cy' %>
         <span>Cymraeg</span>
       <% else %>
-        <%= link_to 'Cymraeg', url_for(locale: 'cy'), lang: 'cy', hreflang: 'cy' %>
+        <%= link_to 'Cymraeg', url_for(locale: 'cy', :'feedback-source' => feedback_source_query_param), lang: 'cy', hreflang: 'cy' %>
```
willp-bl: Same as above

app/views/feedback_sent/index.html.erb
oswaldquek: yes am still waiting on translation from Srini as per my comment on this PR

app/views/shared/_available_languages.html.erb
oswaldquek: @willp-bl good point. is using this sanitizer (https://github.com/alphagov/verify-frontend/blob/master/app/models/feedback_form.rb#L15) sufficient?

app/controllers/feedback_controller.rb
```diff
@@ -4,7 +4,12 @@ class FeedbackController < ApplicationController
   def index
     @form = FeedbackForm.new({})
     flash['feedback_referer'] = request.referer
-    flash['feedback_source'] = params['feedback-source']
+    feedback_source = params['feedback-source']
+    if FEEDBACK_SOURCE_MAPPER.is_feedback_source_valid(feedback_source)
+      flash['feedback_source'] = feedback_source
+    else
+      render 'errors/404', status: 400
```
oswaldquek: @sgreensmith and I think the right response code should be 400 Bad Request, but we also don't think creating another 400.html page is necessary.

app/views/shared/_available_languages.html.erb
oswaldquek: this is now sanitised in the feedback controller

app/views/feedback_sent/index.html.erb
oswaldquek: this is now localised (see cy.yml)

app/controllers/feedback_controller.rb
```diff
@@ -4,7 +4,12 @@ class FeedbackController < ApplicationController
   def index
     @form = FeedbackForm.new({})
     flash['feedback_referer'] = request.referer
-    flash['feedback_source'] = params['feedback-source']
+    feedback_source = params['feedback-source']
+    if FEEDBACK_SOURCE_MAPPER.is_feedback_source_valid(feedback_source)
```
oswaldquek: note this does mean calls to /feedback must have a feedback-source query param

app/controllers/feedback_controller.rb
```diff
@@ -4,7 +4,15 @@ class FeedbackController < ApplicationController
   def index
     @form = FeedbackForm.new({})
     flash['feedback_referer'] = request.referer
-    flash['feedback_source'] = params['feedback-source']
+    feedback_source = params['feedback-source'].nil? ? flash['feedback_source'] : params['feedback-source']
+    flash['feedback_source'] = feedback_source
+    if feedback_source.nil?
+      render
+    elsif FEEDBACK_SOURCE_MAPPER.is_feedback_source_valid(feedback_source)
+      flash['feedback_source'] = feedback_source
```
tunylund: Why do we assign `flash['feedback_source']` again here? This is already done on line 8.

app/controllers/feedback_controller.rb
```diff
@@ -4,7 +4,15 @@ class FeedbackController < ApplicationController
...
+    else
+      render 'errors/404', status: 400
```
tunylund: 404 page with 400 status code? Shouldn't it be 404?

app/models/feedback_source_mapper.rb
```diff
@@ -34,6 +35,10 @@ def initialize(product_page_url)
   }.freeze
   end
 
+  def is_feedback_source_valid(feedback_source)
+    @page_to_source_mappings[feedback_source].nil? ? false : true
```
tunylund: `@page_to_source_mappings.has_key?(feedback_source)` might be more readable

app/controllers/feedback_controller.rb
oswaldquek: It's a 400 because if the feedback source is not valid, it's a bad request, not a "not found". We (@joncrawf @sgreensmith) also thought it's not worth making another error page just for thsi scenario so we're rendering the errors/404 page.

PR #271, opened 2017-03-30
Bump the versions of the front-end assets

robinwhittleton:
- [better hit areas for radios / checkboxes](https://github.com/alphagov/govuk_elements/pull/435)
- [better colour contrast for focused links](https://github.com/alphagov/govuk_template/pull/272)
- [GDS branding on Chrome Android tab view](https://github.com/alphagov/govuk_template/pull/278) (thanks @willp-bl!)


tunylund:
Has this been tested with verify-frontend?

robinwhittleton:
I’ve tested it locally and the contrast and hit areas changes work. The black tab in Chrome/Android code is in there, but I don’t have an appropriate device to test that with. Having said that, the same code went through testing in the original template PR without any problems.

tunylund:
The black tab change is not something that would break things so I feel happy to merge this.

PR #272, opened 2017-03-31
bau: add COOKIE_NOT_FOUND_PAGE page source

oswaldquek:
https://github.com/alphagov/verify-frontend/blob/7379b463065c0d9f52413daa012c3fbc2e1ea2f8/spec/features/user_visits_start_page_spec.rb#L38
shows that there is a COOKIE_NOT_FOUND_PAGE page source. This commit adds it to
the feedback_source_mapper.rb. It doesn't map to any url as no cookies means no
session, so the 'Your session has timed out' text should show after the
submission of feedback. This is reflected in the new test added in
user_visits_feedback_page_spec.rb.

@oswaldquek

### DIFF
--- a/app/models/feedback_source_mapper.rb
+++ b/app/models/feedback_source_mapper.rb
@@ -9,6 +9,7 @@ class FeedbackSourceMapper
       'CERTIFIED_COMPANY_UNAVAILABLE_PAGE' => 'choose_a_certified_company',
       'CONFIRM_YOUR_IDENTITY' => 'confirm_your_identity',
       'CONFIRMATION_PAGE' => 'confirmation',
+      'COOKIE_NOT_FOUND_PAGE' => nil,
       'ERROR_PAGE' => 'start',
       'FAILED_REGISTRATION_PAGE' => 'failed_registration',
       'FAILED_SIGN_IN_PAGE' => 'failed_sign_in',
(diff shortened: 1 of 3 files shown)


app/models/feedback_source_mapper.rb
```diff
@@ -9,6 +9,7 @@ def initialize(product_page_url)
       'CERTIFIED_COMPANY_UNAVAILABLE_PAGE' => 'choose_a_certified_company',
       'CONFIRM_YOUR_IDENTITY' => 'confirm_your_identity',
       'CONFIRMATION_PAGE' => 'confirmation',
+      'COOKIE_NOT_FOUND_PAGE' => nil,
```
tunylund: Do we need a key => nil here?
Should the value at least be the page that the user should be redirected to?

oswaldquek: In this case where there are no cookies, there is no page the user should be redirected to, hence the nil value.
There will still be a list of Services on the "Feedback Sent" page for them to click on to.

app/models/feedback_source_mapper.rb
```diff
@@ -43,6 +44,8 @@ def page_from_source(feedback_source, locale)
     route_name = route_name_from(feedback_source)
     if route_name =~ /https?:.*/
       route_name
+    elsif route_name.nil?
+      return
```
tunylund: It is confusing that this function returns a nil. It would be better if it would always resolve to an existing route.

willp-bl:
What's the status of this PR?

oswaldquek:
Sorry this should have been closed. It's already been deployed.

PR #278, opened 2017-04-26
EID-65: Add select country user journey and test

adityapahuja:


app/controllers/choose_a_country_controller.rb
```diff
@@ -15,6 +15,12 @@ def choose_a_country_submit
       render 'choose_a_country'
       return
     end
+
+    # Call into Policy to change state
+    # POST /api/countries (NL)
+    API_CLIENT.post("/countries/#{session['verify_session_id']}/#{country}",
```
chrisholmes: This behaviour is usually implemented within the SessionProxy class. Please could you move it into there.

app/controllers/choose_a_country_controller.rb
```diff
@@ -15,6 +15,15 @@ def choose_a_country_submit
       render 'choose_a_country'
       return
     end
```
chrisholmes: current behaviour for IDP selection is to render the 'something went wrong' page. I think we should duplicate that.

adityapahuja: The devs think this is a 'mandatory field missing' error and ordinarily the user would be redirected back to the page where the mandatory field should have been entered - ideally with the error code (for internationalisation/translation) and field name (so that the field can be identified on the page/form).  Not going to change at this point.

app/controllers/choose_a_country_controller.rb
adityapahuja: Done.

PR #280, opened 2017-04-28
BAU: Tidy up countries method calls

adityapahuja:
- Updated setup_countries not to return session id to avoid causing confusion.
- Updated get_countries method to validate response and returns a list of country objects.
- Replaced hard coded URL with API endpoint in select_a_country method.
- Updated redirect to country page to stop non-eIDAS users from visiting this page.

Solo: @adityapahuja

### DIFF
--- a/app/controllers/choose_a_country_controller.rb
+++ b/app/controllers/choose_a_country_controller.rb
@@ -3,11 +3,12 @@ class ChooseACountryController < ApplicationController
   before_action :ensure_session_eidas_supported
 
   def choose_a_country
-    setup_countries
+    setup_countries(session['verify_session_id'])
   end
 
   def choose_a_country_submit
-    session_id = setup_countries
+    session_id = session['verify_session_id']
+    setup_countries(session_id)
 
     country = params[:country]
     if country.empty?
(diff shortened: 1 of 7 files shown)


app/controllers/choose_a_country_controller.rb
```diff
@@ -16,18 +17,16 @@ def choose_a_country_submit
...
 
-  def setup_countries
-    session_id = session['verify_session_id']
-    response = SESSION_PROXY.get_countries(session_id)
-    countries_map = response.map { |country| Country.from_api(country) }
-    @countries = COUNTRY_DISPLAY_DECORATOR.decorate_collection(countries_map)
+  def setup_countries(session_id)
+    countries_map = SESSION_PROXY.get_countries(session_id)
+    @countries = COUNTRY_DISPLAY_DECORATOR.decorate_collection(countries_map.countries)
     session_id
```
chrisholmes: does this need to return the session_id now?

adityapahuja: Thanks for spotting this. I forgot to remove this one last line ```session_id```. :)

adityapahuja: Done.

PR #287, opened 2017-05-23
Mob app tt 225

kgarwood:


kgarwood:
Adding the first changes to story TT-225. Setting up UI property to change the "I don't know" option on the select phone view to "Yes, but I'd prefer not to".  Adding if else to the view to show one or the other of these depending on whether visit is part of A or B group in AB Test

PR #290, opened 2017-06-09
TT225 - App Transparency AB Test

NasAshrafThoughtworks:
Story Specific Change
------------------------ 
  - On the select phone page the "I don't know" option has been replaced with "Yes, but I'd prefer not to
   - When this option is selected the IDP picker page will not filter the IDPs but  will display information on whether the IDP may/will/will not need an app installed.
   - The "yes" and "no" option behaviours remain the same.

General AB Testing 
--------------------
- New approach to make it easier to create and remove AB test code
- Works by copying model/view/controllers of the routes affected for the B journey
- Added ability to determine A or B routes in routes.rb based on session cookie

FYI, more info: https://govukverify.atlassian.net/secure/attachment/13019/ABTestingApproach.rtf


NasAshrafThoughtworks:
This pull request has been closed and another will be raised to replace it

PR #291, opened 2017-06-12
TT225 - App Transparency AB Test

NasAshrafThoughtworks:
Story Specific Change
-------------------------

- On the select phone page the "I don't know" option has been replaced with "Yes, but I'd prefer not to
- When this option is selected the IDP picker page will not filter the IDPs but will display information on whether the IDP may/will/will not need an app installed.
- The "yes" and "no" option behaviours remain the same.

General AB Testing
---------------------
- New approach to make it easier to create and remove AB test code
- Works by copying model/view/controllers of the routes affected for the B journey
- Added ability to determine A or B routes in routes.rb based on session cookie
- Any issues with the AB cookie will result in the A route being followed by default

FYI, more info: https://govukverify.atlassian.net/secure/attachment/13019/ABTestingApproach.rtf


app/views/choose_a_certified_company_variant/_idp_option.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      %>
+      <%= hidden_field_tag 'entity_id', identity_provider.entity_id, id: nil %>
+      <div class="company-app-install">
+        <% if @reluctant_mob_installation %>
+            <%= identity_provider.mobile_app_installation.html_safe %>
+        <% end %>
+      </div>
+  <% end %>
+  </div>
+  <!--<div class="company-app-install">-->
```
vixus0: We can get rid of this.

spec/javascripts/ab_test_selector_spec.js
```diff
@@ -0,0 +1,91 @@
+
+describe('AB Test Selector', function () {
+
+  var EXPERIMENT_NAME = 'app_transparency'
+  var VARIANT_EXPERIMENT = EXPERIMENT_NAME + '_variant'
+  var CONTROL_EXPERIMENT = EXPERIMENT_NAME + '_control'
+  var VARIANT_COOKIE = "ab_test=%7B%22" + EXPERIMENT_NAME + "%22%3A%22" + VARIANT_EXPERIMENT + "%22%7D";
```
vixus0: I think you can use `CGI.escape` to do the `%7B...` stuff.

NasAshrafThoughtworks:
Fixed issues raiser by reviewer above

PR #293, opened 2017-06-19
BAU: Make pre-commit run specs through Rake

vixus0:
Travis runs the spec tests through `rake` whereas the pre-commit script
(used for local dev and Jenkins) runs them directly via `rspec`. This
led to issues where the tests ran locally but failed on Travis. The
`RAILS_ENV` environment variable must be set to `test` for Rake to work.

Author: @vixus0

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -2,6 +2,8 @@
 
 . scripts/deploy.sh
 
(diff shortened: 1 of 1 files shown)


adityapahuja:
Looks good.

PR #294, opened 2017-06-20
TT225 - Added analytic to record when the "Yes, but I prefer not to" option is chosen on select phone page

NasAshrafThoughtworks:
- this will allow us to determine if a user actually chose an IDP that required an app to be installed despite originally preferring not to
- analytic is set in the evidence using the identifier smart_phone_prefer_not_to (adding this to the evidence doesn't affect IDP rules matching)

### DIFF
--- a/app/models/select_phone_variant_form.rb
+++ b/app/models/select_phone_variant_form.rb
@@ -20,6 +20,11 @@ class SelectPhoneVariantForm
         answers[attr] = result == 'true' || result == 'reluctant_yes'
       end
     end
+
+    if public_send(:smart_phone) == 'reluctant_yes'
+      answers[:smart_phone_prefer_not_to] = true
+    end
+
     answers
   end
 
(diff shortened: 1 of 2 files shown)


app/models/select_phone_variant_form.rb
```diff
@@ -20,6 +20,11 @@ def selected_answers
         answers[attr] = result == 'true' || result == 'reluctant_yes'
       end
     end
+
+    if public_send(:smart_phone) == 'reluctant_yes'
```
vixus0: Can you replace the `public_send(:smart_phone)` with `smart_phone` or does that not work?

vixus0: Can you replace the `public_send(:smart_phone)` with `smart_phone` or does that not work?

NasAshrafThoughtworks:
public_send removed as per review suggestion

PR #298, opened 2017-06-22
TT-664: Send requested LOA in page titles to Piwik

vixus0:
We currently report the requested and achieved LOAs to Piwik via custom variables. However, attempting to partition user journeys by custom variables through the Piwik UI leads to Piwik only using a subset of all the data available. Until a more permanent solution is found, we can report the requested LOA along with the page title (action name in Piwik terms).

Author: @vixus0


javindo:
Reviewed this, looks good, can't merge myself.

PR #301, opened 2017-06-27
BAU: Make session keys consistently symbols

vixus0:
We currently use a mix of strings and symbols to refer to the same
Rails session values which is confusing. Symbols feel somewhat more
appropriate for defining session keys as Ruby hashes typically use
symbols for keys.

Author: @vixus0


tunylund:
LGTM

PR #310, opened 2017-07-04
BAU: Fingerprint javascript fixes

vixus0:
The way we were creating the `fp_img` fingerprint image element with JS
enabled was causing warnings while running the Jasmine tests. The
fingerprint javascript also didn't follow our standard conventions for
modularising our JS and didn't have any Jasmine spec tests.

Author: @vixus0


richardTowers:
Looks good to me ✨

PR #312, opened 2017-07-11
Tt 721

kgarwood:
These changes ensure that if a user answers “Yes” to an IDP’s interstitial question, it goes to the IDP warning page and if they answer “No”, it goes to a new page that explains why the IDP won’t work for the user.  

**Changing the form and controller for showing interstitial questions.**
The InterstitialQuestionForm was enhanced with a method is_yes_selected that would make the controller logic in redirect_to_idp_question_controller easier to do. 

**Creating a new form to show “IDP won’t work for you page”.**
A new page view was created to describe why an IDP that was advertised to the user would not work for a user with few documents.  The view was put under the management of the redirect_to_idp_question_controller. New form messages had to be added to the en.yml and cy.yml pages to support the new page.

**Supporting a new IDP-specific message**.
The text on the idp_wont_work_for_you_one_doc view will depend on the IDP.  Therefore, a new YML property called “interstitial_explanation” was added.  idp_display_data had to be changed to ensure that the display fields for an IDP would include the new property and that a default would be set for IDPs that had no value for it.

**Changing tests**.
A controller test was developed for the redirect_to_idp_question page, and tests were added to exercise journey routing scenarios when users answered “Yes” or “No” to the IDP’s interstitial question.  Some feature tests were removed because of the coverage provided by the new controller tests.


app/models/interstitial_question_form.rb
```diff
@@ -16,6 +16,10 @@ def selected_answers
     end
   end
 
+  def is_yes_selected
```
vixus0: Don't think you need to wrap `has_extra_info?` in another method, just use it directly.

app/controllers/idp_wont_work_for_you_one_doc_controller.rb
```diff
@@ -0,0 +1,5 @@
+class IdpWontWorkForYouOneDocController < ApplicationController
```
vixus0: I'm inclined to follow the example of some of the other dead-end pages we have and make this page a method+view on the `RedirectToIdpQuestion` controller instead of being an entirely independent controller. What do you think?

PR #322, opened 2017-07-20
EID-153: Add a new endpoint country_response

adityapahuja:
Country response in authn response controller will be used to receive eIDAS Authn Response from a country proxy node and will be forwarded to a new endpoint in saml-proxy. This will enable saml-proxy to process eIDAS Authn Response in a different way compared to Verify Authn Response (e.g. verifying country's signature using metadata aggregator instead of federation metadata).

Authors: @adityapahuja @lamerexter @sfkamath

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ before_script:
 cache:
   directories:
     - $HOME/.phantomjs 
+addons:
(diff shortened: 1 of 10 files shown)


adityapahuja:
Merging as this was done in pair.

PR #324, opened 2017-07-24
TT-721:  Removed IDP-specific comments

kgarwood:
Authors: @kgarwood, @cobainc0


DataMinerUK:
Travis is failing but it appears to be an issue with Travis

PR #326, opened 2017-07-26
Tt 721

kgarwood:
Adding test coverage for scenarios that relate to a person having one document of evidence.  Also modified stub_api.rb so that it can show more than just the one stub-idp-one IDP.  While doing the story, we realised that the coverage for the select documents page needed to be improved to consider cases where only an NI driving licence was selected for document evidence.


kgarwood:
Removed one doc IDP from the idp list that is used in stub_api.rb

PR #327, opened 2017-07-26
TT-869: Fix inconsistencies in AB testing data

hugh-emerson:
Report RP to Piwik on the start page and when reporting AB test name

Authors @michaelwalker @hugh-emerson


hugh-emerson:
I've added a commit which removes the logo on the redirect to service page (the one with the loading spinner)

PR #330, opened 2017-08-01
TT-878: Implement event tracking on the Verify landing page

hugh-emerson:
Extract the Piwik event logging previously in select_proof_of_address into its own file and make the logic more general. Add corresponding spec file. Edit the landing page so we can track the user's selection of 'First Time' and 'Sign In' buttons.

Add feature tests for tracked events on the start page and on the proof of address page.

Authors: @hugh-emerson, @michaelwalker, @jakubmiarka


hugh-emerson:
After speaking to Anne, need to speak to Gabor and Anne together to find out what was really wanted.

PR #333, opened 2017-08-10
TT-936: Small text changes on IDP picker and successfully verified pages

hugh-emerson:
Change IDP picker page header to large h1 format for consistency.
Small text changes on IDP picker loa2 pages and loa1 successfully verified page.

Author: @hugh-emerson


hugh-emerson:
Needs welsh translation for a single word change. English is in the cy.yml for now.

PR #335, opened 2017-08-14
BAU: Rescue from KeyError when no session data retrieved

javindo:
We're seeing a few errors cropping up where no idp_was_recommended value
is available in the session. This can occur for a number of reasons, but
leads to an irrecoverable journey from the user perspective.

The point of failure is currently at the reporting stage, which should
not cause the entire request to fail, therefore the error is now being
rescued from and a message will be included in the reporting explaining
why and where it failed.


javindo:
Just making some changes to this, please don't merge yet.

PR #356, opened 2017-09-01
TT-918 and TT-997 AB tests teardown

hugh-emerson:
Teardown TT-997 loa1 logos AB test.
First stage out of two for the teardown of TT-918 proof of address AB test.

Authors: @hugh-emerson, @jakubmiarka


config/routes.rb
```diff
@@ -7,27 +7,13 @@
...
-  report_to_piwik = -> (experiment_name, reported_alternative, transaction_id, request) {
-    AbTest.report(experiment_name, reported_alternative, transaction_id, request)
-  }
-
-  proof_of_address_a_and_report_to_piwik = SelectRoute.new(PROOF_OF_ADDRESS_EXPERIMENT, 'control', report_to_piwik)
-  proof_of_address_b_and_report_to_piwik = SelectRoute.new(PROOF_OF_ADDRESS_EXPERIMENT, 'with_bank_account', report_to_piwik)
-  proof_of_address_c_and_report_to_piwik = SelectRoute.new(PROOF_OF_ADDRESS_EXPERIMENT, 'without_bank_account', report_to_piwik)
-
-  proof_of_address_a = SelectRoute.new(PROOF_OF_ADDRESS_EXPERIMENT, 'control')
   proof_of_address_b = SelectRoute.new(PROOF_OF_ADDRESS_EXPERIMENT, 'with_bank_account')
```
javindo: What's going on here? Are we keeping proof of address b and c or am I misreading this diff?

hugh-emerson: Yeah, because we're removing a page, there's a complication to handle users who are already on the page at the time of release.

We need to keep the controllers and views so that we can handle them refreshing or submitting the page.

This release tears down the variants for the other controllers, so we don't send people to the proof of address page anymore.
A second release will then remove the proof of address pages/controllers.

javindo: Thanks for clarifying

PR #366, opened 2017-09-20
TT-988: Implement grouping by taxon for timeout page.

MichaelWalker:
Previously the timeout page just showed a list of RPs. This list was
getting too long, so we have changed it so that it is grouped by taxon.
(ie group together all RPs related to Benefits etc)
To do this, a new correllator has been added which groups the transactions
by taxon, and a 'taxon_name' parameter has been added to the rp config.
The transaction list view has also been updated.

Authors: @michaelwalker @jakubmiarka


MichaelWalker:
Please don't merge until the related https://github.com/alphagov/verify-frontend-federation-config/pull/2 has been merged.

PR #377, opened 2017-10-05
TT-1309 Adding content to company picker for the

cobainc0:
current LOA1 experiment.

Solo: @cobainc0


hugh-emerson:
Shouldn't this be on both variants' views?

PR #381, opened 2017-10-11
TT-1327: Remove ReponseForRPApiResource from FrontendApi to Verify Frontend

MahmudH:
- Moved response for RP from SessionProxy to SamlProxyApi
- Added the endpoints of SamlProxy from Frontend Api to Frontend in SamlProxyApiEndpoints
- Created test for SamlProxyApi to test response for RP for both success and error

Authors: @mahmudh @callumknights @MichaelWalker


spec/models/saml_proxy_api_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  let(:session) {
+    {
+        start_time: 'my-session-start-time'
+    }
+  }
+  let(:session_proxy) { SamlProxyApi.new(api_client, originating_ip_store) }
+  let(:ip_address) { '127.0.0.1' }
+
+  include SamlProxyEndpoints
+  # def endpoint(suffix_path)
```
MichaelWalker: It looks like we don't need this commented out code.

spec/models/saml_proxy_api_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    {
+        CookieNames::SESSION_ID_COOKIE_NAME => session_id,
+    }
+  }
+  let(:session) {
+    {
+        start_time: 'my-session-start-time'
+    }
+  }
+  let(:session_proxy) { SamlProxyApi.new(api_client, originating_ip_store) }
```
MichaelWalker: Is this now a SAML Proxy rather than a session proxy?

spec/models/saml_proxy_api_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  # def endpoint(suffix_path)
+  #   session_endpoint(session_id, suffix_path)
+  # end
+
+  describe '#response_for_rp' do
+    it 'should return an rp response' do
+      expect(originating_ip_store).to receive(:get).and_return(ip_address)
+      expect(api_client).to receive(:get)
+        .with(response_for_rp_endpoint(session_id), headers: { "X-Forwarded-For" => ip_address })
+        .and_return('postEndpoint' => 'http://www.example.com',
```
MichaelWalker: Is this still the correct response? ie, have you double checked that the SAML Proxy returns the same response that the frontend_api used to?

PR #382, opened 2017-10-13
Update govuk dependencies

robinwhittleton:
This PR bumps govuk_elements, govuk_frontend_toolkit, govuk_template and govuk-lint to the latest versions, and fixes any associated problems (mostly due to the govuk-lint bump). Each commit should be fairly self-explanatory, would suggest reviewing them in order.


robinwhittleton:
Oh yeah, I tried to write a test for the new error state for the title but couldn’t get past problems accessing the flash object from the spec helper context. If anyone else is able to contribute one that’d be good. Should be fairly simple: basically just setting a flash error and seeing if the subsequent title is prefixed with `Error: `.

hugh-emerson:
Now on master, so closing this. (it was rebased so git will say this branch has different commits)

PR #383, opened 2017-10-19
TT-1410: update text on uplift and registration failure pages

hugh-emerson:
Tweak the wording in the list of reasons for failure.
Make those lists on the two pages match.
Delete 'you can try the following' on the registration non-continue failure page

Solo: @hugh-emerson


hugh-emerson:
Approved by @jakubmiarka

PR #392, opened 2017-10-27
BAU: Add autoprefixer

jakubmiarka:
Adding autoprefixer for CSS to automatically add vendor prefixes for CSS properties requiring prefix in browsers specified in browserlist.

Solo: @jakubmiarka


jakubmiarka:
To test, don't forget to clear rake cache:
`rake tmp:clear`

To see what config is going to be used, based on the browser selection (and what browsers):
`rake autoprefixer:info`

PR #393, opened 2017-10-30
Tt 1304 separate loa1 and loa2

NasAshrafThoughtworks:
Separate controllers so they no longer contain LOA1 and LOA2 branching logic. Affected controllers have been split into separate LOA1 and LOA2 controllers. 

This is not only for better separation of concerns but also us to run LOA1 and LOA2 AB tests concurrently


app/constraints/loa_match.rb
```diff
@@ -0,0 +1,8 @@
+module LoaMatch
+  is_loa = ->(level, _, request) {
+    request.session['requested_loa'] == level
```
MichaelWalker: Is it ever possible for this not to be set?
Do all RPs always give us this information?

Currently I think we implicitly default to loa2 if its not set (do to the `if...else` pattern), so we might not have had to consider this so carefully.

app/controllers/choose_a_certified_company_loa1_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    loa1_idps = current_identity_providers.select { |idp| idp.levels_of_assurance.min == 'LEVEL_1' }
+    @recommended_idps = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(loa1_idps)
+    FEDERATION_REPORTER.report_number_of_idps_recommended(current_transaction, request, @recommended_idps.length)
+    render 'choose_a_certified_company/choose_a_certified_company_LOA1'
+  end
+
+  def select_idp
+    selected_answer_store.store_selected_answers('interstitial', {})
+    select_viewable_idp(params.fetch('entity_id')) do |decorated_idp|
+      session[:selected_idp_was_recommended] = IDP_RECOMMENDATION_GROUPER.recommended?(decorated_idp.identity_provider, selected_evidence, current_identity_providers, current_transaction_simple_id)
```
MichaelWalker: Can this ever be false for LoA1?
It might be added in the future, but I think for now LoA1 only has 'recommended' IDPs,

app/controllers/choose_a_certified_company_loa1_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def warning_or_question_page(decorated_idp)
+    if interstitial_question_flag_enabled_for(decorated_idp)
+      redirect_to_idp_question_path
+    else
+      redirect_to_idp_warning_path
+    end
+  end
+
+  def only_one_uk_doc_selected
```
MichaelWalker: Is this method used?

app/controllers/why_companies_loa1_controller.rb
```diff
@@ -0,0 +1,5 @@
+class WhyCompaniesLoa1Controller < ApplicationController
+  def index
+    render :why_companies_LOA1
```
MichaelWalker: How come this controller is still able to use the symbols, while the others need to use the strings.
Is there a particular advantage to using either one?

spec/controllers/about_loa1_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+require 'controller_helper'
+require 'api_test_helper'
+
+describe AboutLoa1Controller do
+  let(:identity_provider_display_decorator) { double(:IdentityProviderDisplayDecorator) }
+
+  before(:each) do
+    stub_request(:get, CONFIG.config_api_host + '/config/transactions/enabled')
```
MichaelWalker: What makes a call to this - I wasn't expecting a call to enabled transactions in the about controller?

spec/controllers/choose_a_certified_company_loa2_controller_spec.rb
```diff
@@ -4,43 +4,51 @@
...
+      'simpleId' => 'stub-idp-loa1-onboarding',
+      'entityId' => 'http://idcorp-loa1-onboarding.com',
+      'levelsOfAssurance' => %w(LEVEL_1 LEVEL_2)
+  }.freeze
 
   context '#index' do
-    subject { get :index, params: { locale: 'en' } }
+    before :each do
+      stub_const('LOA1_ONBOARDING_IDPS', ['stub-idp-loa1-onboarding'])
+      stub_api_idp_list([STUB_IDP_LOA1, STUB_IDP_ONE_DOC, STUB_IDP_LOA1_ONBOARDING])
```
MichaelWalker: Tiny change - I think that this stub isn't needed any more.
Its immediately superseded by line 33. Once it's gone, I think we can also remove the `STUB_IDP_LOA1_ONBOARDING` object.

PR #395, opened 2017-11-01
TT-1462: metadata endpoints no longer return html on errors

hugh-emerson:
XML endpoints shouldn't start returning HTML when they throw an error

Solo: @hugh-emerson


javindo:
Let's test this

MichaelWalker:
looks good - but the tests are failing? I think there is a test which checks that this returns the silly response (and has the TODO next to it)

PR #401, opened 2017-11-07
TT-1345: AB test - LOA1 radio picker

jakubmiarka:
Pair: @jakubmiarka @MahmudH


app/assets/stylesheets/pages/_choose-a-certified-company.scss
```diff
@@ -21,3 +21,22 @@
...
 }
+
+.idp-radios {
+  .multiple-choice > img {
+    position: absolute;
+    left: 230px;
+    height: 40px;
+  }
+
+  @media (max-width: 640px) {
```
hugh-emerson: Is this a magic width at which the normal layout stops working? Or should this be something like `@include media (tablet) `?

hugh-emerson: position: absolute and left: 230px means that it can overlap with the IDP name even when there's loads of space on the page.

app/assets/javascripts/piwik_idp_picker_tracking_radio_variant.js
```diff
@@ -0,0 +1,50 @@
...
+        var eventAction = numberOfVisibleRadioButtons.toString();
+    }
+
+    _paq.push(['trackEvent', 'Engagement', eventAction, 'IDP visibility']);
+
+    $(window).one('scroll', function() {
+        _paq.push(['trackEvent', 'Engagement', 'Picker scroll', 'scrolled']);
+    })
+
+    var alreadySelected = false;
```
hugh-emerson: can you change this to something like `var alreadySelected = !!$('.multiple-choice input:checked').length`?

PR #402, opened 2017-11-16
TT-1623: Get list of countries from policy directly

MahmudH:
- Frontend calls policy endpoint directly and gets list of countries instead of session proxy
- Updated tests and stubs to call policy and not the session

Authors: @mahmudh @jakubmiarka

### DIFF
--- a/app/models/policy_proxy.rb
+++ b/app/models/policy_proxy.rb
@@ -41,4 +41,13 @@ class PolicyProxy
   def cycle_three_cancel(session_id)
     @api_client.post(cycle_three_cancel_endpoint(session_id), nil, {})
   end
+
+  def get_countries(session_id)
+    response = @api_client.get(countries_endpoint(session_id))
+    CountryResponse.validated_response(response)
+  end
+
+  def select_a_country(session_id, country)
+    @api_client.post(select_a_country_endpoint(session_id, country), '', {})
+  end
 end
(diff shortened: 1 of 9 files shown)


app/models/policy_proxy.rb
```diff
@@ -41,4 +41,15 @@ def submit_cycle_three_value(session_id, value)
...
   end
+
+  def get_countries(session_id)
+    response = @api_client.get(countries_endpoint(session_id))
+    CountryResponse.validated_response(response)
+  end
+
+  def select_a_country(session_id, country)
+    # Call into Policy to change state
+    # POST /api/countries (NL)
```
hugh-emerson: You kept these comments but I don't think they're helpful or accurate

spec/models/policy_proxy_spec.rb
```diff
@@ -92,4 +92,33 @@ def endpoint(suffix_path)
       policy_proxy.cycle_three_cancel(session_id)
     end
   end
+
+  describe('#get_countries') do
+    countries_json = [
+        { 'entityId' => 'http://netherlandsEnitity.nl', 'simpleId' => 'NL', 'enabled' => false },
+        { 'entityId' => 'http://spainEnitity.es', 'simpleId' => 'ES', 'enabled' => true }
```
hugh-emerson: May as well fix these "Enitity" typos while you're making changes here?

PR #405, opened 2017-11-20
TT-1615: LOA1 shortened journey AB test

jakubmiarka:
Re-run of the previous LOA1 shortened journey AB test, this time using _loa1_shortened_journey_v3_ as experiment name


config/routes.rb
```diff
@@ -73,9 +73,14 @@ def add_routes(routes_name)
...
       get 'about_identity_accounts', to: 'about_loa1#identity_accounts', as: :about_identity_accounts
       get 'about_choosing_a_company', to: 'about_loa1#choosing_a_company', as: :about_choosing_a_company
+
+      add_routes :loa1_shortened_journey_ab_test_routes
     end
 
     constraints IsLoa2 do
+      get 'start', to: 'start#index', as: :start
+      post 'start', to: 'start#request_post', as: :start
+      get 'begin_registration', to: 'start#register', as: :begin_registration
```
MichaelWalker: why do we need to change any LoA2 routes?

jakubmiarka: it's just moving them from the 'common' place to split them into LoA1/2 journeys

app/controllers/start_controller.rb
```diff
@@ -1,12 +1,13 @@
...
+include LoaMatch
 
 class StartController < ApplicationController
   layout 'slides'
 
   def index
     @form = StartForm.new({})
 
-    FEDERATION_REPORTER.report_start_page(current_transaction, request)
+    FEDERATION_REPORTER.report_start_page(current_transaction, request) unless session[:requested_loa] == 'LEVEL_1' # ab test variant hack, remove with teardown
```
MichaelWalker: This helps the race condition (from 3 racing things down to 2) but doesn't eliminate it.

In the past, we have added a hack to add AB Test reporting to the client side start page as well as the server side reporting. That way the race condition still exists (we've not got a good way of removing it yet) but it doesn't matter so much which one wins. 

We can have a chat about it in the morning!

PR #416, opened 2017-12-04
tt-1609 

CharlesIC:
Updated frontend to get idp list with loa from new endpoint in hub that returns a list filtered to show only the correct idps for journey. Also updated sign in controller to use separate new endpoint in hub for full list of idps to maintain current expected behaviour. Pair: @CharlesIC @rachelthecodesmith


rachelthecodesmith:
Dont merge yet, found an issue 🤕

app/controllers/application_controller.rb
```diff
@@ -173,8 +173,12 @@ def selected_identity_provider
     end
   end
 
-  def current_identity_providers
-    CONFIG_PROXY.get_idp_list(current_transaction_entity_id).idps
+  def current_identity_providers_for_loa
+    CONFIG_PROXY.get_idp_list_for_loa(session[:transaction_entity_id], session[:requested_loa]).idps
```
MichaelWalker: I think we already have methods for `session[:transaction_entity_id]` and `session[:requested_loa]`.
If we do, we should use them to reduce the number of magic strings in this file!

MichaelWalker: same for the rest of the changes in this file if we have a sensible method!

app/controllers/application_controller.rb
```diff
@@ -189,15 +193,22 @@ def hide_feedback_link
     @hide_feedback_link = true
   end
 
-  def select_viewable_idp(entity_id)
-    for_viewable_idp(entity_id) do |decorated_idp|
+  def select_viewable_idp_for_sign_in(entity_id)
```
MichaelWalker: Is this just trying to do
```
visible_idps
  .find {#matching IDP}
  .map { #to decorated_idp }
```
??

Or am I missing something - it confused me a little! You've not made it worse though, so perhaps not worth doing anything about??

app/controllers/choose_a_certified_company_about.rb
```diff
@@ -1,10 +1,10 @@
 module ChooseACertifiedCompanyAbout
   def about
     simple_id = params[:company]
-    matching_idp = current_identity_providers.detect { |idp| idp.simple_id == simple_id }
+    matching_idp = current_identity_providers_for_loa.detect { |idp| idp.simple_id == simple_id }
```
MichaelWalker: genuine question - is it easier to understand if the method was called `current_identity_providers_for_registration`?

If I'm understanding this correctly, signIn / registration is the conceptual divide in the front end?
On the other hand, literally it's the `loa` that makes the difference, so naming it after that also makes sense

app/models/config_endpoints.rb
```diff
@@ -1,13 +1,17 @@
 module ConfigEndpoints
   PATH = '/config'.freeze
   PATH_PREFIX = Pathname(PATH)
-  IDP_LIST_SUFFIX = 'idps/idp-list'.freeze
+  IDP_LIST_SUFFIX = 'idps/idp-list/%s/%s'.freeze
```
MichaelWalker: not required, but a (maybe) nice ruby feature I found the other day was that you can make these parameters named. ie `foo = idps/idp-list/%<transaction_name>s/%<loa>s`

Then you can call it with `foo % {transaction_name: name, loa: 'LEVEL_2'}`

spec/controllers/sign_in_controller_spec.rb
```diff
@@ -5,7 +5,7 @@
 
 describe SignInController do
   before(:each) do
-    stub_api_idp_list([{ 'simpleId' => 'stub-idp-one',
+    stub_api_idp_list_for_sign_in([{ 'simpleId' => 'stub-idp-one',
```
MichaelWalker: tiny OCD thing (which you can ignore if you like) but the alignment of this is now wrong!

spec/features/user_visits_confirm_your_identity_page_spec.rb
```diff
@@ -105,22 +110,23 @@ def set_up_session(idp_entity_id)
   describe 'and the journey hint cookie is invalid in some way' do
     it 'should redirect to sign in page when the journey cookie is not set' do
       set_session_and_session_cookies!
-      stub_api_idp_list
+      stub_api_idp_list_for_sign_in
```
MichaelWalker: quite a lot of these. Could they be moved to the `before` step?
We also never assert that they actually get called - is that important? (probably not but I thought I'd check)

Final question - how come we need the stubs now when we didn't before? We aren't making any extra requests for this page are we?

MichaelWalker:
All looks great! No interesting comments at all - just a couple of tiny things!
Feel free to push any further changes to here or just close this and push whatever you like to master! 😄

MichaelWalker:
(oops - meant to close the comment window not the PR - sorry!)

stub/api/stub_api.rb
```diff
@@ -25,11 +25,12 @@ class StubApi < Sinatra::Base
     status 200
     "{
       \"simpleId\":\"test-rp\",
-      \"loaList\":[\"#{level_of_assurance}\"]
+      \"loaList\":[\"#{level_of_assurance}\"],
+      \"serviceHomepage\":\"www.example.com\"
```
MichaelWalker: Sorry! Thanks for fixing!

app/controllers/application_controller.rb
rachelthecodesmith: Yeah, I honestly didnt investigate this, just kept what was there for the most part, can have a look though

app/controllers/choose_a_certified_company_about.rb
rachelthecodesmith: So in hub we have it as sign in and loa, HE and I discussed it and decided to do it like that for reasons that I now cant really remember but I think it was one of those no completely good choice things. For now I think we should stick to sign in and loa to match with the hub

app/controllers/choose_a_certified_company_loa1_controller.rb
```diff
@@ -35,12 +34,4 @@ def only_one_uk_doc_selected
   def interstitial_question_flag_enabled_for(decorated_idp)
     IDP_FEATURE_FLAGS_CHECKER.enabled?(:show_interstitial_question_loa1, decorated_idp.simple_id)
   end
-
-  def is_onboarding?(idp)
-    LOA1_ONBOARDING_IDPS.include?(idp.simple_id)
-  end
-
-  def is_test_rp?
```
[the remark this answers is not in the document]
rachelthecodesmith: Yeah... ops. May have missed that one.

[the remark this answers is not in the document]
rachelthecodesmith: This is odd, I cant see this on my branch, seems I did delete it... confused as to why is it still showing up for you.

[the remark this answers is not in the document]
rachelthecodesmith: I am a moron. This was just a comment on the fact it's removed. Totally spaced on that

spec/features/user_visits_confirm_your_identity_page_spec.rb
rachelthecodesmith: How do you mean didnt need them before? So we always need to call for an idp list, sometimes its the one for sign in and sometimes it the one for loa. Before, as there was only one list it just had one call to one endpoint which could go in the before blocks. Now that we have a different call for sign in it makes it harder to separate it out in the tests in their current order and structure. I did mean to go back and see if they could be reorganised to dry it out a bit as I know it's not ideal repeating. Will have a look. Let me know if that other stuff doesnt make sense

app/controllers/choose_a_certified_company_about.rb
MichaelWalker: fine by me! sounds right that we match the hub

spec/features/user_visits_confirm_your_identity_page_spec.rb
MichaelWalker: oh I see, yeah that makes sense!

app/controllers/choose_a_certified_company_loa1_controller.rb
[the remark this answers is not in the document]
MichaelWalker: yep - I was trying to be friendly! I should have known that would be confusing - sorry

(also looking back it was quite badly worded!)

app/models/config_endpoints.rb
rachelthecodesmith: Done, that's so much nicer

spec/controllers/select_phone_controller_spec.rb
```diff
@@ -18,23 +18,23 @@
     subject { post :select_phone, params: { locale: 'en', select_phone_form: valid_phone_evidence } }
 
     it 'redirects to choose certified company page when eligible IDPs exist' do
-      stub_api_idp_list([{ 'simpleId' => 'stub-idp-one',
-                           'entityId' => 'http://idcorp.com',
-                           'levelsOfAssurance' => %w(LEVEL_2) }])
+      stub_api_idp_list_for_loa([{ 'simpleId' => 'stub-idp-one',
+                                  'entityId' => 'http://idcorp.com',
+                                  'levelsOfAssurance' => %w(LEVEL_2) }], 'LEVEL_1')
```
jakubmiarka: would it be better to keep using LOA2 in this instance (and other below), since select phone controller is not used within LOA1 journey anyway?

rachelthecodesmith: Good point, I'll amend, thanks Jakub

PR #433, opened 2018-01-04
Use raw

MichaelWalker:
'raw' is a rails method wrapping 'html_safe' and is defined as

def raw(stringish)
  stringish.to_s.html_safe
end

For use in views, it has 2 advantages over html_safe

1) Its nil safe.
html_safe is only defined on strings. Therefore nil.html_safe causes a
NoSuchMethod error to be thrown.
raw(nil) is fine however, as nil.to_s returns an empty string, which you
can then call html_safe on (again to return an empty string).
This means that using nil acts as a defensive measure against variables
in our template unexpectedly being nil.

2) Its a clearer name.
'html_safe' is potentially confusing, as it can sound like it is making
the string safe in some way (ie sanitising the input). In reality, its
marking the string as safe - ie telling rails that it doesn't need to
escape the string at all.
Raw is a little clearer in this respect.

Solo: @michaelwalker


jakubmiarka:
Looks good! Should this reference the actual kaizen ticket [TT-1756](https://govukverify.atlassian.net/browse/TT-1756) rather than just BAU?

MichaelWalker:
Thanks Jakub - Good though - I'll update the comment, and re-push it!
(there is also a failing check, so I'll take a look at that!)

PR #442, opened 2018-02-06
BAU: disable the Rails IP spoofing check

hugh-emerson:
It can cause errors but gives no benefit. From Rails release notes:
"The fact that Rails checks for IP spoofing can be a nuisance for sites that do heavy traffic with cell phones, because their proxies don't generally set things up right."
In our case, we think it’s some misconfiguration at the Welsh Assembly causing errors in the frontend.

Solo: @hugh-emerson

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -35,6 +35,9 @@ module VerifyFrontend
 
     config.exceptions_app = self.routes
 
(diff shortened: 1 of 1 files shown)


kornysietsma:
Hmm - this seems like it might be of value? @jakubmiarka @rachelthecodesmith  - does this make any sense?

jakubmiarka:
Hugh said it was something he saw while yakking, so might be worth merging.

PR #447, opened 2018-02-23
HUB-63: Tear down AB test to remove performance management

CharlesIC:
Reverse changes made in HUB-52
c6e9d07862ff94a88a070b819a48729d6891ea64

Author: @CharlesIC


jakubmiarka:
Nice work, looks good to me, don't forget the fed and app config updates too. In order not to break joint, this has to go first and then the other ones.

CharlesIC:
Pushing the configs now...

PR #451, opened 2018-03-07
Hub 71

rachelthecodesmith:


app/controllers/choose_a_certified_company_loa2_variant_controller.rb
```diff
@@ -1,42 +1,24 @@
...
-    end
-  end
-
-  def only_one_uk_doc_selected
-    (%i[passport driving_licence] & selected_evidence).size == 1
-  end
-
-  def interstitial_question_flag_enabled_for(decorated_idp)
-    IDP_FEATURE_FLAGS_CHECKER.enabled?(:show_interstitial_question, decorated_idp.simple_id)
+  def warning_page
```
jakubmiarka: isn't this redundant? it could be in the select_idp method directly, no need for this to be a private function

config/main_routes.rb
```diff
@@ -47,43 +48,43 @@ def add_routes(routes_name)
...
-  get 'idp_wont_work_for_you_one_doc', to: 'redirect_to_idp_question_loa2#idp_wont_work_for_you', as: :idp_wont_work_for_you_one_doc
   get 'confirmation', to: 'confirmation_loa2#index', as: :confirmation
   get 'about', to: 'about_loa2#index', as: :about
   get 'about_certified_companies', to: 'about_loa2#certified_companies', as: :about_certified_companies
   get 'about_identity_accounts', to: 'about_loa2#identity_accounts', as: :about_identity_accounts
-  get 'about_choosing_a_company', to: 'about_loa2#choosing_a_company', as: :about_choosing_a_company
+  get 'redirect_to_idp_question', to: 'redirect_to_idp_question_loa2#index', as: :redirect_to_idp_question
+  post 'redirect_to_idp_question', to: 'redirect_to_idp_question_loa2#continue', as: :redirect_to_idp_question_submit
+  get 'idp_wont_work_for_you_one_doc', to: 'redirect_to_idp_question_loa2#idp_wont_work_for_you', as: :idp_wont_work_for_you_one_doc
+  # Add back in after AB test HUB-71
```
jakubmiarka: my preference would be keep the routes in order and just comment them out.. we have tried to keep the routes in the order which they appear in the hub

jakubmiarka:
No tests for the new content? Should we be testing whether the variant actually display the variant text. Especially if this is going to be re-iterated we might want to have some test for the variant so we are confident throught the iterations.

jakubmiarka:
Is the acceptance criteria outdated on the ticket? It says:
"If I am a user in the variant accessing a LOA2 service (excluding UC) then I will see the blue slides as normal and at the end of the blue slides I will then see the picker page."

But the variant has one more page between the slides and the picker (/about-choosing-a-company).

config/locales/en.yml
```diff
@@ -165,6 +165,13 @@ en:
         </ul>
         <p>However, each company uses different methods for each of these steps.</p>
         <p>So we’ll now ask you some questions to check which companies can verify you. We’ll then send you to the website of the company you choose.</p>
+      content_variant_html: |
```
jakubmiarka: Welsh version missing. There will be no content on the page if the user has switchted to the the welsh language.

PR #452, opened 2018-03-08
HUB-71 feature tests and routing fix

swilson96:
Fix the routing so that users can't go off the rails and see the questions (e.g. by clicking a link or a bookmark, or url-hacking)

Fix a bug in the welsh yaml, which was spotted by the new tests (yay tests)

Adding feature tests for the two affected pages

### DIFF
--- a/spec/features/user_visits_about_choosing_a_company_page_spec.rb
+++ b/spec/features/user_visits_about_choosing_a_company_page_spec.rb
@@ -26,34 +26,4 @@ RSpec.describe 'When the user visits the about choosing a company page' do
     click_link 'Continue'
     expect(page).to have_current_path(will_it_work_for_me_path)
   end
-
-  # HUB-71 Delete with test teardown
-  context 'in the no questions test variant' do
-    before(:each) do
-      set_session_and_ab_session_cookies!('no_questions' => 'no_questions_variant')
-    end
-
-    it 'will display variant content in Welsh' do
-      visit '/am-ddewis-cwmni'
-
-      expect(page).to have_content 'Dod o hyd i’r cwmni iawn i’ch dilysu chi'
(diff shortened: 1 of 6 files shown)


spec/features/user_visits_about_choosing_a_company_page_spec.rb
```diff
@@ -25,4 +26,34 @@
     click_link 'Continue'
     expect(page).to have_current_path(will_it_work_for_me_path)
   end
+
+  # HUB-71 Delete with test teardown
```
jakubmiarka: It would be better if this was in a separate file, as a variant feature test. Like we did with the unit tests. Less we touch the control files, the better ;)

spec/features/user_visits_choose_a_certified_company_page_spec.rb
```diff
@@ -174,4 +174,170 @@
       end
     end
   end
+
+  # HUB-71 Delete with test teardown
```
jakubmiarka: Same as above, a separate variant test file

spec/features/user_visits_choose_a_certified_company_page_spec.rb
```diff
@@ -174,4 +174,170 @@
...
+
+      it 'displays the page in Welsh' do
+        visit '/dewis-cwmni-ardystiedig'
+
+        expect(page).to have_title t('hub.choose_a_certified_company.title', locale: :cy)
+        expect(page).to have_css 'html[lang=cy]'
+      end
+    end
+
+    context 'user is from an LOA1 service' do
```
jakubmiarka: This isn't required as LOA1 doesn't have the controllers and AB test is not running on LOA1.

PR #453, opened 2018-03-12
HUB-79 show interstitial idp question when 0 docs

swilson96:
We are adding 0 docs segments to an IDP with an interstitial question.

We want the question to show up for 0-docs users as well as for 1-doc
users.

Also removed a dead method from the loa1 version of the controller.

### DIFF
--- a/app/controllers/choose_a_certified_company_loa2_controller.rb
+++ b/app/controllers/choose_a_certified_company_loa2_controller.rb
@@ -32,7 +32,7 @@ private
   end
 
   def only_one_uk_doc_selected
-    (%i[passport driving_licence] & selected_evidence).size == 1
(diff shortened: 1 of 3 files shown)


app/controllers/choose_a_certified_company_loa2_controller.rb
```diff
@@ -32,7 +32,7 @@ def warning_or_question_page(decorated_idp)
   end
 
   def only_one_uk_doc_selected
```
jakubmiarka: do we want to rename the function to be more accurate that it's not just 'only one uk doc'?

swilson96: oh yeah i did do that then reverted my changes, thanks for spotting

PR #473, opened 2018-05-11
Hub-111 Remove unnecessary Piwik reporting

rylangooch:
Removes duplicated piwik reporting

Co-Authored-By: Callum Knights <callum.knights@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/other_identity_documents_controller.rb
+++ b/app/controllers/other_identity_documents_controller.rb
@@ -6,7 +6,6 @@ class OtherIdentityDocumentsController < ApplicationController
   def select_other_documents
     @form = OtherIdentityDocumentsForm.new(params['other_identity_documents_form'] || {})
     if @form.valid?
-      report_to_analytics('Other Documents Next')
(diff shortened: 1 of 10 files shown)


willp-bl:
Isn't this reporting for when users have javascript off, or it failed to load?

willp-bl:
It would be better to leave it in server side, and remove the client side reporting if duplicates are an issue

jakubmiarka:
No, page view reporting happens even when the visitors have javascript switched off (using img tag). But these 'actions' (which are being removed) are already being reported by the actual pageview (JS or ), they don't need a separate/duplicate server-side reporting. Especially when the data is not used by anyone (neither or us perfomance analysts) and potentially can cause race conditions.

Interesting fact: about 1.6% (Nov 2017) of visitors on Verify have JavaScript disabled (they need it for IDP part of the journey anyway). And client-side reporting is more useful as we can track visitors' screen resolution, scrolling behaviour etc...

PR #476, opened 2018-05-17
Hub-117 Accept different journey hint values

z-gooch:
Adds redirecting of signing in users and registering users using the 'journey_hint' value

Co-Authored-By: Callum Knights <callum.knights@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/authn_request_controller.rb
+++ b/app/controllers/authn_request_controller.rb
@@ -74,11 +74,12 @@ private
     elsif check_journey_hint('sign_in')
       redirect_to begin_sign_in_path
     else
+      Rails.logger.info "journey_hint value: #{params['journey_hint']}"
       redirect_to confirm_your_identity_path
     end
   end
 
   def check_journey_hint(path)
-    params['journey_hint'] == path
+    params['journey_hint'].downcase == path
   end
 end
--- a/spec/features/redirects_with_journey_hint_authn_request_spec.rb
+++ b/spec/features/redirects_with_journey_hint_authn_request_spec.rb
@@ -8,6 +8,12 @@ describe 'pages redirect with journey hint parameter', type: :request do
     expect(response).to redirect_to begin_registration_path
   end
 
+  it 'will redirect the user to registration path when journey hint parameter is set to registration and is case insensitive' do
+    stub_session_creation
+    post '/SAML2/SSO', params: { 'SAMLRequest' => 'my-saml-request', 'RelayState' => 'my-relay-state', 'journey_hint' => 'RegiStraTion' }
+    expect(response).to redirect_to begin_registration_path
+  end
+
   it 'will redirect the user to sign-in path when journey hint parameter is set to sign_in' do
     stub_session_creation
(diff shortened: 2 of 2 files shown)


app/controllers/authn_request_controller.rb
```diff
@@ -11,7 +11,7 @@ def rp_request
     AbTest.set_or_update_ab_test_cookie(current_transaction_simple_id, cookies)
 
     if params['journey_hint'].present?
-      redirect_to confirm_your_identity_path
+      follow_journey_hint
```
jakubmiarka: nice one, I like that

app/controllers/authn_request_controller.rb
```diff
@@ -67,4 +67,18 @@ def set_requested_loa(levels_of_assurance)
   def set_transaction_homepage(transaction_homepage)
     session[:transaction_homepage] = transaction_homepage
   end
+
+  def follow_journey_hint
+    if check_journey_hint('registration')
+      redirect_to begin_registration_path
+    elsif check_journey_hint('sign_in')
+      redirect_to sign_in_path
```
jakubmiarka: we should create a new route (similar to begin_registration_path). When user goes to the sign-in we need to set the Piwik custom variable (JOURNEY_TYPE) to SIGN_IN. Landing directly on the sign-in page will not update the custom variable.

app/views/test_saml/index.html.erb
```diff
@@ -26,6 +26,18 @@
   <%= hidden_field_tag('journey_hint', 'my-journey-hint') %>
   <div><%= submit_tag 'saml-post-journey-hint' %></div>
 <% end -%>
+<%= form_tag('/SAML2/SSO') do -%>
+  <%= hidden_field_tag('SAMLRequest', 'my-saml-request') %>
+  <%= hidden_field_tag('RelayState', 'my-relay-state') %>
+  <%= hidden_field_tag('journey_hint', 'registration') %>
+  <div><%= submit_tag 'saml-post-registration' %></div>
```
jakubmiarka: could we rename the button to say `saml-post-journey-hint-registration` ? so it's clearer it's using the journey hint parameter, same below with sign-in

spec/features/user_sends_authn_request_spec.rb
```diff
@@ -60,6 +60,27 @@
       expect(page.get_rack_session['transaction_supports_eidas']).to eql false
     end
 
+    it 'will redirect the user to /begin_registration when journey hint is set to registration' do
```
jakubmiarka: the user doesn't get redirected to `/begin-registration` as it's a virtual route (the requests just goes through the begin_registration_path). It should say `/about` as that's the page which the user ends up on

spec/features/user_sends_authn_request_spec.rb
```diff
@@ -60,6 +60,27 @@
...
+      stub_session_creation
+
+      visit('/test-saml')
+      click_button 'saml-post-registration'
+
+      expect(page).to have_title t('hub.about.title')
+      expect(page.get_rack_session['transaction_supports_eidas']).to eql false
+    end
+
+    it 'will redirect the user to /sign_in when journey hint is set to sign_in' do
```
jakubmiarka: small thing, it's `/sign-in` not `/sign_in`

PR #481, opened 2018-05-31
Hub 125 IDP Reminder Backwards Compatibility

rylangooch:
Ensures users with old version of the verify-front-journey-hint cookie have
the same IDP hint behaviour as previously, whilst maintaining updated IDP
hint behaviour for those with the new version of the cookie.

### DIFF
--- a/app/controllers/partials/journey_hinting_partial_controller.rb
+++ b/app/controllers/partials/journey_hinting_partial_controller.rb
@@ -16,12 +16,27 @@ module JourneyHintingPartialController
     journey_hint.nil? ? nil : journey_hint[status]
   end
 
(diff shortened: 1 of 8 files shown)


jakubmiarka:
DO NOT MERGE - to be discussed

jakubmiarka:
Discussed with Tom Hall and agreed not to go ahead.
Reasons:

- less accurate hinting for people with the old cookie (as they might have failed)
- confusing user experience for people with the old cookie (if they see the 'old' hint, select IDP, go back, the hint will be gone)
- skewed reporting (as we'll see all the hints but won't be able to distinguish which are the new (more accurate ones) and which are the old ones for attempts only)

PR #483, opened 2018-06-05
Hub 114 report attempts registration

rachelthecodesmith:
First release to test new string for analytics on one segment on successful registration journey. If this does not conflict with existing reporting and if it provides useful information we will roll this out for sign in and all registration journeys else the code will be removed.

### DIFF
--- a/app/controllers/partials/analytics_partial_controller.rb
+++ b/app/controllers/partials/analytics_partial_controller.rb
@@ -33,4 +33,32 @@ private
     @piwik_custom_variables_img_tracker =
       current_transaction_custom_variable.merge(loa_requested_custom_variable)
   end
+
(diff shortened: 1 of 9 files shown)


app/controllers/partials/analytics_partial_controller.rb
```diff
@@ -33,4 +33,33 @@ def custom_variables_for_img_tracker
...
+    end
+  end
+
+  def segment_should_be_reported?
+    return false if session[:user_segments].nil?
+    session[:user_segments].include?(SEGMENT_TO_BE_TESTED) || session[:user_segments].include?(SEGMENT_FOR_TEST)
+  end
+
+  def journey_should_be_reported?
+    session[:journey_type] == JOURNEY_TYPE
```
jakubmiarka: where does the `session[:journey_type]` get set?

PR #487, opened 2018-06-07
EID-457: New eIDAS country/scheme picker

CharlesIC:
- added local vars to .env to point to flags and scheme logos
- new choose_a_country controller, html template and CSS stylesheet
- country display decorator now adds schemes to countries
- added display data, repository and decorator for eidas_scheme objetcts
- added a method to yaml_loader to load files as a hash with filename as simple_id so we can map schemes to their logos
- added a method to repository_factory to load display_data without translation as eIDAS scheme names don't need that
- made new picker responsive by using divs and percentage widths


willp-bl:
What's happening with the commits in this PR?  Seems to be about 9 stories.  Also it would be good to have some useful descriptive text in the summary.

CharlesIC:
I merged master into the branch and I guess git picked it up. But the changes detected are only the ones in my own commit..... go figure!

I've copied the description from my commit 👍

app/views/choose_a_country/choose_a_country.html.erb
```diff
@@ -1,19 +1,33 @@
...
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <h1 class="heading-large"><%= t('hub.choose_country.heading') %></h1>
+    <p><%= t 'hub.choose_country.description' %></p>
+  </div>
+</div>
+
+<div class="grid-row">
+  <div class="column-full">
+    <table class="country-picker">
```
jakubmiarka: I'd prefer to use divs instead of table. Tables are not responsive.

app/models/display/country_display_decorator.rb
```diff
@@ -1,24 +1,27 @@
...
     end
 
   private
 
-    def correlate_display_data(country)
-      return not_viewable(country) unless !country.nil? && country.enabled == true
+    def correlate_display_data(country, schemes)
+      return not_viewable(country) unless !country.nil? && country.enabled == true && !schemes.nil? && !schemes.empty?
       simple_id = country.simple_id
+      flag_path = File.join(@flags_directory, "#{simple_id}.svg")
```
jakubmiarka: .svg is not supported in IE8, just FYI. IE8 [is still required](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in) for a functional support so missing flag might be OK?

CharlesIC: Thanks for the reference, Jakub. The page indeed says that IE 8-10 is assigned the 'functional' level of support and that's defined as 'your service might not look perfect in that browser but must still be usable without errors'. That's the case if the flags are not showing so it's fine to leave them as svg.

app/assets/stylesheets/pages/_choose-a-country.scss
```diff
@@ -1,4 +1,34 @@
...
-  margin-bottom: 6px;
+  tr {height: 75px;}
+  tr:last-child td {border: none;}
+  td {vertical-align: middle;}
+}
+
+.country-flag {
+  width: 70px;
+  padding: 0;
+  img {max-width: 40px; max-height: 30px;
```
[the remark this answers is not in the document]
CharlesIC: Done 👍

app/views/choose_a_country/choose_a_country.html.erb
```diff
@@ -1,19 +1,53 @@
 <%= page_title 'hub.choose_country.title' %>
+<% content_for :feedback_source, 'CHOOSE_A_CERTIFIED_COMPANY_PAGE' %>
```
jakubmiarka: this should say `CHOOSE_A_COUNTRY_PAGE`?

CharlesIC: Yes, I've already fixed it but didn't commit 🤦‍♀️ 
Re-pushing....

app/controllers/redirect_to_country_controller.rb
```diff
@@ -4,10 +4,42 @@ class RedirectToCountryController < ApplicationController
...
+  end
+
+private
+
+  def decorated_country(country_simple_id)
+    countries = POLICY_PROXY.get_countries(session[:verify_session_id]).countries.select do |country|
+      country.simple_id == country_simple_id
+    end
+
+    return Display::NotViewableCountry.new unless countries.count == 1
```
jakubmiarka: What is the purpose of the `unless` condition?

jakubmiarka: Oh I see now

CharlesIC: Just a check to make sure only one country in the collection matches the simple id

app/views/choose_a_country/choose_a_country.html.erb
```diff
@@ -1,19 +1,53 @@
...
-    <p><a href="javascript:history.back()">Try another way to sign in.</a></p>
+  <div class="column-full js-redirect-to-country" data-location="<%= url_for(controller: 'redirect_to_country', action: 'choose_a_country_submit_ajax') %>">
+    <div class="country-picker" id="country-picker">
+      <% @countries.each do |country| %>
+        <div class="country-option">
+          <% country.schemes.each do |scheme| %>
+            <div class="scheme-option">
+              <div class="scheme-info">
+                <div class="country-flag">
+                  <img src="<%= country.flag_path %>" alt="<%= country.display_name %> flag">
```
jakubmiarka: use Rails img tag, e.g. `<%= image_tag country.flag_path, alt: country.display_name %>` ... might help with the failing travis

CharlesIC: Done 👍

PR #488, opened 2018-06-11
Translations from config service - CTV-201/CTV-204

karlbaker02:
This PR adds stubs for the endpoints in the config service which will serve translations to the frontend. It also adds in some infrastructure that is necessary to get translations from an external source rather than from local YAML files.

I'll tidy up the commit history later.


app/models/config_proxy.rb
```diff
@@ -10,6 +10,21 @@ def get_transaction_details(transaction_entity_id)
...
+    translations_for_locale = TransactionTranslationResponse.validated_response(response)
+
+    {
+      name: translations_for_locale.name,
+      rp_name: translations_for_locale.rp_name,
+      analytics_description: translations_for_locale.analytics_description,
+      other_ways_text: translations_for_locale.other_ways_text,
+      other_ways_description: translations_for_locale.other_ways_description,
+      tailored_text: translations_for_locale.tailored_text,
+      taxon_name: translations_for_locale.taxon_name
```
jakubmiarka: there are more properties than just these... some of the other ones (used by just specific RPs) are for example:
`custom_fail_heading`, `custom_fail_what_next_content`, `custom_fail_try_another_text` etc... See `app/models/display/rp_display_data.rb`

Also, could this be a more dynamic? If we add a new property, we'll have to add it twice. Here and also at the model `transaction_translation_response.rb` and whenever it will be used.

karlbaker02:
Hi @jakubmiarka. Thanks for your feedback! I've updated this PR with the ticket now (CTV-201).

The point of the epic under which this piece of work falls is that these translations will no longer live in FFC - instead they'll be moved to the HFC and served by the config service. Therefore, you'll only need to update translations as part of a HFC release, removing the need for us to release the frontend whenever we onboard a new RP.

app/models/config_proxy.rb
karlbaker02: Thanks for pointing this out, I'll have a look at what other properties exist - but I take it is only those listed at https://github.com/alphagov/verify-frontend/blob/master/app/models/display/rp_display_data.rb ?

jakubmiarka: Yep that should be it!

Yeah I know, that's why I'm giving my feedback early ;)

jakubmiarka:
It will remove the need to do two releases when onboarding a new RP, but will add one extra release for the frontend team, when changing the schema/adding new properties/or just updating the text. If we update/change the translation text, what is the plan for the frontend to pick up the change? If I'm reading the PR correctly, the translations are still loaded on the app startup. So if there's a change to the config we will have to manually restart the frontend app or is there a different approach being taken?

karlbaker02:
The plan is that the frontend will make ongoing requests to the config service for translations whenever translations are needed, in much the same way that the config service is used by the frontend currently. If the config service is unavailable for any reason, then the frontend will fallback to use the translations it has cached.

In practical terms, this means that for any changes to existing translations and/or any entirely new translations added, no work will be required by the frontend team. For any updates to the schema, there will need to be a release, but there would need to be a release currently in any case - the difference at the moment is that there are two files that need changing instead of one (I'm going to look into this as mentioned in one of the previous comments).

karlbaker02:
Tests are currently failing because translations are attempting to be pulled when the app starts up (via the Rakefile).

karlbaker02:
**Update:**

The above comments are somewhat out-of-date. 

This PR now starts-up the frontend app without RP translations loaded. When a request is received, the app checks whether translations are present and if not, loads translations for each RP for the current locale. Translations are updated on a continuous basis, whenever a method is called on an instance of `RpDisplayData` that retrieves one of the pre-defined fields.

There are some performance considerations to be made with this PR, notably that a single request may need multiple field translations to render the response, each of which will currently make a call to the config service. The plan is to load test this code and determine if/how response times are affected, after which we'll be in a better position to determine if there are any optimisations that need to be made (to either the config service or to the frontend).

This PR has been reviewed by @jakubmiarka on 22/06.

PR #490, opened 2018-06-13
Sentry/Front/194: Delete from selected answers safely

vixus0:
Should fix https://sentry-hub-prod-a-dmz.ida.digital.cabinet-office.gov.uk/sentry/front/group/194/

See the following for more information on the safe navigation operator: http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

Author: @vixus0


richardTowers:
Nice. I didn't know about the safe navigation operator in ruby. Love it in C# though. Will be littering all future code with `&.`s.

PR #493, opened 2018-06-14
Hub 114 report attempts and outcomes

rachelthecodesmith:


app/models/analytics/federation_reporter.rb
```diff
@@ -54,19 +54,24 @@ def report_sign_in_journey_hint_shown(current_transaction, request, idp_display_
...
         request,
-        "ATTEMPT_#{attempt_number}: #{journey_type} | #{transaction_simple_id} | #{idp_name} | #{user_segments}"
+        report
       )
     end
 
-    def report_user_idp_outcome(journey_type:, attempt_number:, current_transaction:, request:, idp_name:, user_segments:, transaction_simple_id:, response_status:)
+    def report_user_idp_outcome(journey_type:, attempt_number:, current_transaction:, request:, idp_name:, user_segments:, transaction_simple_id:, hint_followed:, response_status:)
+      report = "OUTCOME_#{attempt_number}| #{journey_type} | #{transaction_simple_id} | #{idp_name} | #{user_segments} |"
+      report << journey_hint_details(hint_followed)
```
jakubmiarka: I wonder if we should think about the scenario when someone tries to login (with hint) but then does a registration - the hint will show up on the registration attempt

app/controllers/redirect_to_idp_controller.rb
```diff
@@ -13,11 +13,13 @@ def register
 
   def sign_in
     request_form
+    increase_attempt_number
     if session[:user_followed_journey_hint].nil?
       FEDERATION_REPORTER.report_sign_in_idp_selection(current_transaction, request, session[:selected_idp_name])
     else
       FEDERATION_REPORTER.report_sign_in_idp_selection_after_journey_hint(current_transaction, request, session[:selected_idp_name], session[:user_followed_journey_hint])
     end
+    report_user_idp_attempt_to_piwik
```
jakubmiarka: shouldn't this be before the FEDERATION_REPORTER in order to keep the current reporting intact? e.g. for the analytics team to use the transitions from "Sign in - <IDP>" to "SUCCESS" ? Or is it now irrelevant?

app/controllers/partials/idp_selection_partial_controller.rb
```diff
@@ -28,18 +32,18 @@ def ajax_idp_redirection_registration_request(recommended)
...
-        request: request,
-        idp_name: session[:selected_idp_name],
-        user_segments: session[:user_segments],
-        transaction_simple_id: session[:transaction_simple_id],
-        attempt_number: session[:attempt_number],
-        journey_type: session[:journey_type]
-      )
-    end
+  def report_user_idp_attempt_to_piwik(hint = nil)
+    save_journey_hint_details(hint)
```
jakubmiarka: for a readibility sake, I would move the `unless hint` condition here. So when reading the code you don't need to go to the function to see what it does if the hint is nil.

PR #501, opened 2018-06-21
Update verify-front-journey-hint tracking cookie for Country Responses too

minhngocd:
rearrange the cookie tracking specs to allow for overwriteable variables so we can add the same for the country response endpoints

### DIFF
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -20,11 +20,7 @@ class AuthnResponseController < SamlController
   end
 
   def country_response
-    if params['RelayState'].nil?
-      if session[:transaction_supports_eidas]
-        params['RelayState'] = session[:verify_session_id]
-      end
-    end
+    params['RelayState'] ||= session[:verify_session_id] if session[:transaction_supports_eidas]
     raise_error_if_session_mismatch(params['RelayState'], session[:verify_session_id])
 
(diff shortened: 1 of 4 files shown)


app/controllers/authn_response_controller.rb
```diff
@@ -34,34 +30,42 @@ def country_response
...
+    raise Errors::WarningLevelError, error_message if relay_state != session_id
   end
 
   def idp_response_handlers
     handlers = {
-      SUCCESS => ->(response) { analytics_reporters[SUCCESS].call(response); idp_selection_reporters[SUCCESS].call; redirecters[SUCCESS].call(response) },
-      CANCEL => ->(response) { analytics_reporters[CANCEL].call(response); idp_selection_reporters[CANCEL].call; redirecters[CANCEL].call(response) },
-      FAILED_UPLIFT => ->(response) { analytics_reporters[FAILED_UPLIFT].call(response); idp_selection_reporters[FAILED_UPLIFT].call; redirecters[FAILED_UPLIFT].call(response) },
-      PENDING => ->(response) { analytics_reporters[PENDING].call(response); idp_selection_reporters[PENDING].call; redirecters[PENDING].call(response) }
+      SUCCESS => ->(response) { idp_response_handler_for SUCCESS, response },
```
jakubmiarka: can we use the brackets around the parameters which are being passed in? Just to stay consistent with the wider codebase (and the country handlers is using brackets too e.g. `country_response_handler_for(SUCCESS, response)`)

PR #504, opened 2018-06-26
HUB-97: IDP reminder on Verify start page

jakubmiarka:
In order to make it easier for users to re-use their IDP, we're moving the
IDP reminder from the sign-in page to the start page. A new view has been
 created to render a different 'hinted' view when the cookie is present.

Solo: @jakubmiarka


jakubmiarka:
To merge and release on a Monday morning to make it easier to analyze for the performance team. Discuss & agree with the performance team which Monday

jakubmiarka:
There have been some changes/fixes to how we report hint success. After we have enough data and evaluated them, we can release this.

jakubmiarka:
Targeting merge & deploy on 6th August, to allow for performance team make necessary changes to their reporting pipeline.

app/models/analytics/federation_reporter.rb
```diff
@@ -100,7 +100,7 @@ def report_sign_in_idp_selection_after_journey_hint(current_transaction, request
       report_action(
         current_transaction,
         request,
-        "Sign In - #{idp_display_name} - Hint #{hint_followed ? 'Followed' : 'Ignored'}"
+        "Sign In - #{idp_display_name} - #{hint_followed ? 'Followed' : 'Ignored'} Hint"
```
galund: As discussed, apart from this change, 👍

app/assets/stylesheets/pages/_slides.scss
```diff
@@ -71,6 +71,10 @@
     }
   }
 
+  .other-options {
+    font-size: 18px;
```
galund: This is "smaller than small" (i.e. smaller than something immediately below called "small_details") -- so I guess we should be sure that having yet another size of text is really the right thing to do?

app/views/start/start_with_hint.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= page_title 'hub.start.title', {}, "with hint (#{@suggested_idp[0].display_name})" %>
```
galund: Why does the "with hint" thing actually go into the user-facing title of the page? This seems useful for debugging but is it really supposed to be seen by an end user?

jakubmiarka: It doesn't go the user-facing title of the page. It goes only to the Piwik reporting page title. Same like Welsh page titles are shown to users in Welsh but reported in English.

app/assets/stylesheets/pages/_slides.scss
jakubmiarka: I think this might have been based on a designer's prototype. Removing the new class then and sticking with the existing one.

PR #507, opened 2018-06-27
EID-638 Handle new journey hint values

kerrr:
- We used to specify an `eidas_journey` parameter to
  frontend to skip to the country picker.
- We now have a `eidas_sign_in` value for the `journey_hint`
  parameter which we have replaced it with.
- We have also renamed `sign_in` to `uk_idp_sign_in` to
  differentiate it from the eIDAS version.
- We also now explictly support `unspecified` which is
  the same as if the journey hint is missing.
- Previously we would raise an error if the journey hint was set to
  country-sign-in and the RP had eidas disabled. We will now treat this
  as undefined and redirect to the start page.
- Some tests relied on using a non standard journey_hint to redirect to
  subject_confirmation flow. This is no longer valid. Tests and setup
  changed to reflect this.

### DIFF
--- a/app/controllers/authn_request_controller.rb
+++ b/app/controllers/authn_request_controller.rb
@@ -13,17 +13,9 @@ class AuthnRequestController < SamlController
 
     AbTest.set_or_update_ab_test_cookie(current_transaction_simple_id, cookies)
 
-    if params['journey_hint'].present?
-      logger.info "journey_hint value: #{params['journey_hint']}"
-      follow_journey_hint
-    elsif params['eidas_journey'].present?
-      raise StandardError, 'Users session does not support eIDAS journeys' unless session[:transaction_supports_eidas]
-      redirect_to choose_a_country_path
-    elsif session[:transaction_supports_eidas]
-      redirect_to prove_identity_path
-    else
-      redirect_to start_path
-    end
+
+    journey_hint = params['journey_hint'].present? ? params['journey_hint'] : 'unspecified'
+    follow_journey_hint journey_hint
   end
 
 private
@@ -74,17 +66,26 @@ private
     session[:transaction_homepage] = transaction_homepage
   end
 
-  def follow_journey_hint
-    if check_journey_hint('registration')
+  def follow_journey_hint(hint)
+    case hint
+    when 'registration'
       redirect_to begin_registration_path
-    elsif check_journey_hint('sign_in')
+    when 'uk_idp_sign_in'
       redirect_to begin_sign_in_path
-    else
+    when 'eidas_sign_in'
+      if session[:transaction_supports_eidas]
+        redirect_to choose_a_country_path
+      else
+        redirect_to start_path
+      end
+    when 'submission_confirmation'
       redirect_to confirm_your_identity_path
+    else
+      if session[:transaction_supports_eidas]
+        redirect_to prove_identity_path
+      else
+        redirect_to start_path
+      end
     end
-  end
-
-  def check_journey_hint(path)
-    params['journey_hint'].downcase == path
   end
 end
--- a/app/views/test_saml/index.html.erb
+++ b/app/views/test_saml/index.html.erb
@@ -27,6 +27,12 @@
   <div><%= submit_tag 'saml-post-journey-hint' %></div>
 <% end -%>
 <%= form_tag('/SAML2/SSO') do -%>
+    <%= hidden_field_tag('SAMLRequest', 'my-saml-request') %>
+    <%= hidden_field_tag('RelayState', 'my-relay-state') %>
+    <%= hidden_field_tag('journey_hint', 'submission_confirmation') %>
+    <div><%= submit_tag 'saml-post-journey-hint_submission_confirmation' %></div>
+  <% end -%>
+<%= form_tag('/SAML2/SSO') do -%>
   <%= hidden_field_tag('SAMLRequest', 'my-saml-request') %>
   <%= hidden_field_tag('RelayState', 'my-relay-state') %>
   <%= hidden_field_tag('journey_hint', 'registration') %>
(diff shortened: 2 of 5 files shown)


app/controllers/authn_request_controller.rb
```diff
@@ -13,17 +13,9 @@ def rp_request
...
-      raise StandardError, 'Users session does not support eIDAS journeys' unless session[:transaction_supports_eidas]
-      redirect_to choose_a_country_path
-    elsif session[:transaction_supports_eidas]
-      redirect_to prove_identity_path
-    else
-      redirect_to start_path
-    end
+
+    journey_hint = params['journey_hint'].present? ? params['journey_hint'] : 'unspecified'
+    follow_journey_hint journey_hint
```
jakubmiarka: If journey hint is not present we shouldn't call the `follow_journey_hint` method, in terms of readability it doesn't make sense. Could we either rename the method name or *follow* the hint only if there is one to follow?

app/controllers/authn_request_controller.rb
```diff
@@ -74,17 +66,26 @@ def set_transaction_homepage(transaction_homepage)
...
+    when 'uk_idp_sign_in'
       redirect_to begin_sign_in_path
-    else
+    when 'eidas_sign_in'
+      if session[:transaction_supports_eidas]
+        redirect_to choose_a_country_path
+      else
+        redirect_to start_path
+      end
+    when 'submission_confirmation'
```
jakubmiarka: have we confirmed HMRC are actually using this value for their non-repudiation journey? We have placed logging in there to confirm, which is being deleted with this PR.

app/controllers/authn_request_controller.rb
```diff
@@ -74,17 +66,26 @@ def set_transaction_homepage(transaction_homepage)
     session[:transaction_homepage] = transaction_homepage
   end
 
-  def follow_journey_hint
-    if check_journey_hint('registration')
+  def follow_journey_hint(hint)
+    case hint
+    when 'registration'
```
jakubmiarka: feels way too complex, one method having 3 `when`s, 2 `if`s and 2 `else`s ... I think we should be able to get rid of the `else`s at least...

app/controllers/authn_request_controller.rb
```diff
@@ -74,17 +66,26 @@ def set_transaction_homepage(transaction_homepage)
...
       redirect_to begin_registration_path
-    elsif check_journey_hint('sign_in')
+    when 'uk_idp_sign_in'
```
jakubmiarka: this now becomes a case-sensitive, so if the RP sends this value in uppercase, it will go through the normal start path and not the sign-in one

spec/features/redirects_with_journey_hint_authn_request_spec.rb
```diff
@@ -8,28 +8,22 @@
     expect(response).to redirect_to begin_registration_path
   end
 
-  it 'will redirect the user to registration path when journey hint parameter is set to registration and is case insensitive' do
+  it 'will redirect the user to sign-in path when journey hint parameter is set to uk_idp_sign_in' do
```
jakubmiarka: this test is no longer testing for the casing of the string. It's in an input from many different RPs so we should be able to deal with little variations such as casing.

app/views/test_saml/index.html.erb
```diff
@@ -26,6 +26,12 @@
   <%= hidden_field_tag('journey_hint', 'my-journey-hint') %>
   <div><%= submit_tag 'saml-post-journey-hint' %></div>
 <% end -%>
+<%= form_tag('/SAML2/SSO') do -%>
+    <%= hidden_field_tag('SAMLRequest', 'my-saml-request') %>
+    <%= hidden_field_tag('RelayState', 'my-relay-state') %>
+    <%= hidden_field_tag('journey_hint', 'submission_confirmation') %>
+    <div><%= submit_tag 'saml-post-journey-hint_submission_confirmation' %></div>
```
jakubmiarka: personally I would call this a non-repudation journey, instead of submission_confirmation - for new comers it might be unclear what it is. We also use it in Test RP [here](https://github.com/alphagov/verify-test-rp/blob/b9c6894b00cb696cef3ac3fa55b73ed1f3564fba/src/main/resources/uk/gov/ida/rp/testrp/views/landingPage.jade#L46)

app/controllers/authn_request_controller.rb
kerrr: The thinking here is that we want to follow the exact same code path for the no hint case and the unknown hint case. So we thought it was simpler to set the no hint case to a value and then follow that case.

app/controllers/authn_request_controller.rb
kerrr: In discussion with our PM and tech arc we decided that we would be strict about the valid allowed values for journey hints. This included any casing issues.

app/views/test_saml/index.html.erb
kerrr: I don't have a strong feeling on this. In this particular case I thought it was more obvious that the name matched the hint value. I can change it.

spec/features/redirects_with_journey_hint_authn_request_spec.rb
kerrr: As above, in discussion with our PM and tech arc we decided that we would be strict about the valid allowed values for journey hints. This included any casing issues.

app/controllers/authn_request_controller.rb
jakubmiarka: my main concern is that the parameter will be passed/set from multiple places, i.e. GOV.UK pages, emails, RP themselves etc. In many cases using CMS so it will be difficult for us to be strict and enforce across all the actors at the same time. But if this is a group decision...

app/controllers/authn_request_controller.rb
```diff
@@ -74,17 +66,29 @@ def set_transaction_homepage(transaction_homepage)
...
       redirect_to begin_sign_in_path
-    else
+    when 'eidas_sign_in'
+      do_eidas_sign_in_redirect
+    when 'submission_confirmation'
       redirect_to confirm_your_identity_path
+    else
+      logger.info "journey_hint value: #{hint}"
```
jakubmiarka: this will run for all the RP requests now, flooding the logs with `journey_hint value: unspecified`. We wanted to see the RPs who are currently sending the journey_hint, what it is they're sending. If we're sure that for non-repudiation journey (confirm_your_identity_path), they're sending 'submission_confirmation' then we can remove the logging. As of now, the logic is pretty much if journey_hint present, then do the non-repudiation journey.

app/controllers/authn_request_controller.rb
jakubmiarka: any plans how to monitor or follow up on any issues? Because it will affect the users who might be redirected to a wrong place, not really affecting RPs

app/controllers/authn_request_controller.rb
kerrr: We specifically want to remove that an unknown value redirects to the non-repudiation flow. I take the point that this will log too much. I've put in a check to handle the not present case separately and log anything else.

jakubmiarka: Cool, nice!

PR #508, opened 2018-06-27
HUB-190: Hide feedback form from integration.

callumknights:
The reason behind this is that since the IP whitelist has been removed the integration environment
has been visible to the public, including the feedback form, which on integration lacks any
zendesk config and merely throws an error when submitted. We decided to modify the frontend so that
when in integration if the feedback form is accessed a gov.uk styled error is shown to the user
stating that feedback is disabled in this environment.

To make these changes I have added a new view to the feedback folder, and made some changes to the
respective controller, that now checks the value of a new environment variable named
FEEDBACK_DISABLED, if true it renders the new view with the error message, if false renders the
feedback form as normal.

I have also added some unit tests for the additional functionality in the
controller, as well as a feature test to check the correct page is shown when feedback is disabled.

### DIFF
--- a/app/views/feedback/disabled.html.erb
+++ b/app/views/feedback/disabled.html.erb
@@ -1,0 +1,17 @@
+<%= page_title 'hub.feedback_disabled.title' %>
+<% content_for :show_to_search_engine, true %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+
+    <div class="error-summary" role="alert" aria-labelledby="error-summary-heading-example-1" tabindex="-1">
+
+      <h1 class="heading-medium error-summary-heading" id="error-summary-heading-example-1">
+        <%= t 'hub.feedback_disabled.heading' %>
+      </h1>
+
+      <p> <%= t 'hub.feedback_disabled.error_content' %> </p>
+
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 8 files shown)


app/views/feedback/disabled.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+
+    <%= render 'shared/form-errors', errors: flash[:errors] %>
+
+    <div class="error-summary" role="alert" aria-labelledby="error-summary-heading-example-1" tabindex="-1">
+
+      <h1 class="heading-medium error-summary-heading" id="error-summary-heading-example-1">
+        Feedback is disabled on this environment
```
jakubmiarka: Can we move the text to the locales files? Apart from the static pages no views should have a content hardcoded.

app/views/feedback/disabled.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      </h1>
+
+      <p>You are currently using a non-production version of Verify, where feedback has been disabled. Please use the regular <a href="https://www.signin.service.gov.uk/feedback">feedback form</a>.</p>
```
jakubmiarka: Has it been agreed to point users to the production feedback form? I assumed the original issue was that the RP testers and developers were using the form. Surely they should use their channels and not the user's channel for providing a feedback.

PR #517, opened 2018-07-18
BAU: Remove warnings about already initialized constants

jakubmiarka:
The tests re-initialize the global RP_TRANSLATION_SERVICE and CONFIG_PROXY
instances resulting in warnings in the results.


jakubmiarka:
The travis build failed because of the timeout, the other one passed 🤷‍♂️

PR #520, opened 2018-07-20
EID 776: Add uk_idp_start journey hint

FelixMarcusMillne:


app/controllers/authn_request_controller.rb
```diff
@@ -75,10 +76,7 @@ def redirect_for_journey_hint(hint)
       do_eidas_sign_in_redirect
     when 'submission_confirmation'
       redirect_to confirm_your_identity_path
-    when 'unspecified'
```
jakubmiarka: If you're removing the 'unspecified' value, then [RFC](https://github.com/alphagov/verify-architecture/pull/34) should be updated/replaced

FelixMarcusMillne: Hadn't seen the RFC. Thanks for pointing it out.

FelixMarcusMillne: Talked with Kerr about it, and decided to leave the option for unspecified in (allowing an option that does nothing). Have added the logging back in though, to make a note of any unrecognised hints.

PR #521, opened 2018-07-20
EID-637: Refactor Authn Response Controller for readability

minhngocd:
Previously there were a lot of lambdas being passed around and called so
tracing through the code was a bit difficult. Since these lambdas are
all called immediately after fetching anyways, they've been moved into
named methods. Hopefully this looks more readable and easier to trace
through a journey.

Bigger changes with this commit include:
	* Removing OTHER status - This means changes to current behaviour of:
		* when receiving unknown status, send FAILED to piwik
rather than OTHER. The special 'OTHER' case always maps to FAILED
scenarios anyways, except piwik. Changing this will make the code much
easier to generalise.
	* CANCEL country journey sets tracking cookie to CANCEL rather
than FAILED. nothing really uses the tracking cookie for country yet.
Previous value of FAILED probably doesn't reflect actual status.
	* separating redirects for IDP and Country journeys. Even though
it looks similar, they are subtly different, and the previous way of
forcing them to use the same map caused more complexity. Having a
seperate map for IDP and Country will allow us in the next step to
direct Country signin failures to another page without affecting the IDP
flow.

Solo: @minhngocd

### DIFF
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -52,7 +52,6 @@ private
   end
 
   def handle_country_response(status, response)
-    set_journey_status(session[:selected_country], status)
     redirect_to country_redirects(status, response)
   end
 
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -1,22 +1,30 @@
+# frozen_string_literal: true
+
 require 'partials/user_cookies_partial_controller'
 
 class AuthnResponseController < SamlController
   protect_from_forgery except: %i[idp_response country_response]
 
-  SIGNING_IN_STATE = 'SIGN_IN_WITH_IDP'.freeze
-  REGISTERING_STATE = 'REGISTER_WITH_IDP'.freeze
-  SUCCESS = 'SUCCESS'.freeze
-  CANCEL = 'CANCEL'.freeze
-  FAILED = 'FAILED'.freeze
-  FAILED_UPLIFT = 'FAILED_UPLIFT'.freeze
-  PENDING = 'PENDING'.freeze
-  OTHER = 'OTHER'.freeze
+  SIGNING_IN_STATE = 'SIGN_IN_WITH_IDP'
+  REGISTERING_STATE = 'REGISTER_WITH_IDP'
+  SUCCESS = 'SUCCESS'
+  CANCEL = 'CANCEL'
+  FAILED = 'FAILED'
+  FAILED_UPLIFT = 'FAILED_UPLIFT'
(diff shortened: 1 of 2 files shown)


app/controllers/authn_response_controller.rb
```diff
@@ -34,83 +45,57 @@ def raise_error_if_session_mismatch(relay_state, session_id)
...
+    set_journey_status(session[:selected_idp], status)
+    redirect_to idp_redirects(status, response)
   end
 
-  def idp_response_handler_for(status, response, report_status = status)
-    analytics_reporters[status].call(response)
-    selection_reporters(:selected_idp, report_status).call
-    redirecters[status].call(response)
+  def handle_country_response(status, response)
+    set_journey_status(session[:selected_country], status)
```
jakubmiarka: We shouldn't be storing the country in the hint cookie. When we read the value we expect the `entity_id` of the IDP, not country. Also the hint is designed to work with UK IDPs, not countries.

simonwo: Understood. This is used to for hinting the IDP to the user when they come to Verify again, and that code isn't designed to handle countries. So, we should remove this call.

app/controllers/authn_response_controller.rb
```diff
@@ -1,30 +1,41 @@
...
-  PENDING = 'PENDING'.freeze
-  OTHER = 'OTHER'.freeze
+  SIGNING_IN_STATE = 'SIGN_IN_WITH_IDP'
+  REGISTERING_STATE = 'REGISTER_WITH_IDP'
+  SUCCESS = 'SUCCESS'
+  CANCEL = 'CANCEL'
+  FAILED = 'FAILED'
+  FAILED_UPLIFT = 'FAILED_UPLIFT'
+  PENDING = 'PENDING'
+  OTHER = 'OTHER'
```
FelixMarcusMillne: Do we still need this? Looks like it isn't used anywhere.

app/controllers/authn_response_controller.rb
```diff
@@ -1,30 +1,41 @@
+# frozen_string_literal: true
+
 require 'partials/user_cookies_partial_controller'
```
FelixMarcusMillne: This is required on application_controller already. Does it need to be defined here?

simonwo: Given that we specifically call `set_journey_hint_by_status`, would it better to leave it here for explicitness?

app/controllers/authn_response_controller.rb
simonwo: Agreed. Will be removed in next PR.

PR #522, opened 2018-07-23
EID-637: Adding eIDAS specific error page

simonwo:
This adds an eIDAS-specific error page for users that come back from an EU proxy node with a failed status.

There are still a bunch of TODOs that are blocked, so we're going to merge this now and deal with them later.
- The grammar for the Welsh translations isn't quite correct because sometimes the country names change when prefixed with "in". Content design is working on an alternative.
- Still need a translation for the URL. Welsh language unit have been contacted.

### DIFF
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -80,21 +80,23 @@ private
   end
 
   def idp_redirects(status, response)
+    is_registration = response.is_registration
     {
-      SUCCESS => response.is_registration ? confirmation_path : response_processing_path,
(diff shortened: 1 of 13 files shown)


kgarwood:
We've just rebased against master and the merge conflicts should disappear

simonwo:
We're taking the decision to ignore Codacy in this case because it's not possible to shorten the lines without sacrificing readability.

jakubmiarka:
Can you do a release when you merge it? So we don't build up to large releases (even if eIDAS is not live yet)

PR #529, opened 2018-08-09
EID-795: Revise error pages

CharlesIC:
Redirect users to either the Verify/eIDAS picker or the service error page depending on whether the journey is a regular Verify (IDP) journey or an eIDAS one
To do achieve this, we need to ensure that only either selected_idp or selected_country is set at any one time

### DIFF
--- a/app/controllers/choose_a_country_controller.rb
+++ b/app/controllers/choose_a_country_controller.rb
@@ -6,6 +6,7 @@ class ChooseACountryController < ApplicationController
   before_action :setup_countries
 
   def choose_a_country
+    session.delete(:selected_idp)
     @other_ways_description = current_transaction.other_ways_description
   end
 
--- a/app/controllers/choose_a_country_controller.rb
+++ b/app/controllers/choose_a_country_controller.rb
@@ -6,7 +6,6 @@ class ChooseACountryController < ApplicationController
   before_action :setup_countries
 
   def choose_a_country
-    session.delete(:selected_idp)
     @other_ways_description = current_transaction.other_ways_description
   end
 
--- a/app/controllers/response_processing_controller.rb
+++ b/app/controllers/response_processing_controller.rb
@@ -35,12 +35,9 @@ class ResponseProcessingController < ApplicationController
 
   def render_error_page(feedback_source)
     @hide_available_languages = false
-
-    @display_text = { other_ways_description: current_transaction.other_ways_description,
-                      other_ways_text: current_transaction.other_ways_text }
-
-    if session[:selected_country] then @redirect_path = prove_identity_path end
-    if session[:selected_idp] then @redirect_path = redirect_to_service_error_path end
+    @other_ways_text = current_transaction.other_ways_text
+    @other_ways_description = current_transaction.other_ways_description
+    @redirect_path = session[:selected_country] ? prove_identity_path : redirect_to_service_error_path
 
     render 'matching_error', status: 500, locals: { error_feedback_source: feedback_source }
   end
--- a/app/controllers/response_processing_controller.rb
+++ b/app/controllers/response_processing_controller.rb
@@ -35,8 +35,13 @@ class ResponseProcessingController < ApplicationController
 
   def render_error_page(feedback_source)
     @hide_available_languages = false
-    @other_ways_description = current_transaction.other_ways_description
-    @other_ways_text = current_transaction.other_ways_text
+
+    @display_text = { other_ways_description: current_transaction.other_ways_description,
+                      other_ways_text: current_transaction.other_ways_text }
+
+    if session[:selected_country] then @redirect_path = prove_identity_path end
+    if session[:selected_idp] then @redirect_path = redirect_to_service_error_path end
+
     render 'matching_error', status: 500, locals: { error_feedback_source: feedback_source }
   end
 end
--- a/app/controllers/start_controller.rb
+++ b/app/controllers/start_controller.rb
@@ -8,7 +8,6 @@ class StartController < ApplicationController
   before_action :set_device_type_evidence
 
   def index
-    session.delete(:selected_country)
     entity_id = entity_id_of_journey_hint_for('SUCCESS')
     @suggested_idp = entity_id.nil? ? [] : retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_sign_in, entity_id)
     if @suggested_idp.empty? || !defined?(@suggested_idp)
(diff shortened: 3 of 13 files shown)


app/controllers/response_processing_controller.rb
```diff
@@ -35,8 +35,13 @@ def index
...
   def render_error_page(feedback_source)
     @hide_available_languages = false
-    @other_ways_description = current_transaction.other_ways_description
-    @other_ways_text = current_transaction.other_ways_text
+
+    @display_text = { other_ways_description: current_transaction.other_ways_description,
+                      other_ways_text: current_transaction.other_ways_text }
+
+    if session[:selected_country] then @redirect_path = prove_identity_path end
+    if session[:selected_idp] then @redirect_path = redirect_to_service_error_path end
```
jakubmiarka: You could simplify the two IFs to a single one by doing something like:
```
@redirect_path = redirect_to_service_error_path
@redirect_path = prove_identity_path unless session[:selected_idp]
```
This way we also have a default redirect if for some reason none of the session variables were set.

Other way of doing this redirect could be to keep this as it is, and in the view change the `href` to point to a new method which would have the redirect logic. E.g. `redirect_to_error_or_start_path` and the method/route could then either redirect to `prove_your_identity_path` or `redirect_to_service_error_path` once clicked.

app/controllers/choose_a_country_controller.rb
```diff
@@ -6,6 +6,7 @@ class ChooseACountryController < ApplicationController
   before_action :setup_countries
 
   def choose_a_country
+    session.delete(:selected_idp)
```
minhngocd: why don't we reset the selected IDP at the same time we set the selected country and vice versa? this leaves a gap when nothing's selected, and user may decide to back track instead of choosing the country (submit)

spec/features/user_visits_response_processing_page_spec.rb
```diff
@@ -26,4 +26,23 @@
     visit '/prosesu-ymateb'
     expect(page).to have_css 'html[lang=cy]'
   end
+
+  it 'should redirect to prove-identity page on matching error for an eIDAS journey' do
+    page.set_rack_session selected_country: 'stub-country'
+    stub_matching_outcome MatchingOutcomeResponse::SHOW_MATCHING_ERROR_PAGE
+
+    visit '/response-processing'
+    click_on 'Other ways to prove your identity online'
```
jakubmiarka: instead of hardcoding the text, could we use the `t()` function to get the translation. We try to do that with other tests - if we change text, it's not gonna fail the tests.

spec/features/user_visits_start_page_spec.rb
```diff
@@ -113,6 +113,15 @@
     expect(page).to have_css('meta[name="robots"][content="noindex"]', visible: false)
   end
 
+  it 'will delete selected_country from session' do
```
jakubmiarka: Would we want to also test if it doesn't delete `selected_idp` ?

spec/features/user_visits_choose_a_country_page_spec.rb
```diff
@@ -40,6 +40,14 @@ def when_i_choose_to_continue
     click_button t('navigation.continue')
   end
 
+  it 'should delete selected_idp from session when choose_a_country page is rendered' do
```
jakubmiarka: This probably should be a controller (unit) test instead of a feature one. Similar with the start one below.

app/controllers/response_processing_controller.rb
```diff
@@ -35,8 +35,13 @@ def index
 
   def render_error_page(feedback_source)
     @hide_available_languages = false
-    @other_ways_description = current_transaction.other_ways_description
-    @other_ways_text = current_transaction.other_ways_text
+
+    @display_text = { other_ways_description: current_transaction.other_ways_description,
```
minhngocd: I feel like the hash is extra structure we don't need at this point

app/controllers/response_processing_controller.rb
minhngocd: we can probably replace this with a ternary? since there's always gotta be a link - either eIDAS or otherwise

app/controllers/start_controller.rb
```diff
@@ -8,6 +8,7 @@ class StartController < ApplicationController
   before_action :set_device_type_evidence
 
   def index
+    session.delete(:selected_country)
```
minhngocd: same as above - can we reset the country where it's being set? sign_in_controller, i believe

app/controllers/choose_a_country_controller.rb
jakubmiarka: Good idea, although we have to do for signin and registration I believe, so multiple places?

Anyway, it's fine when nothing is selected, it should only be set when user leaves for IDP/country or just came back. When browsing in hub it should be empty...

CharlesIC: I discussed this with @jakubmiarka and he advised it was better to reset those variables at the point of selection of an eIDAS or Verify journey instead of later on. Actually, there's only one scenario where you could set both and that's only if the user goes through the eIDAS/Verify picker - so it makes sense to reset them also when you click one of the green buttons on that page. Also, you wouldn't have a gap as you will receive a new value later on anyway (when selecting an IDP or country), and it's normal to not have any value selected at that stage of the journey anyway, so it's fine.

app/controllers/response_processing_controller.rb
jakubmiarka: or as Minno pointed out, ternary could be an option too..

app/controllers/response_processing_controller.rb
CharlesIC: Actually, it's a Ruby 'rule' (or code quality recommendation) that you should have a maximum of two instance variables in a view - one a hash for anything to share with the view and one for user info. We already had two instance variables in the view and when I added the redirect path it exceeded the limit, so I had to make this into a hash instead. I'm not sure how strictly this rule is enforced within Verify but you do get a warning

app/controllers/start_controller.rb
CharlesIC: _see my comment above_

app/controllers/response_processing_controller.rb
CharlesIC: We could  but this is the most robust solution in case none or two values are set (which should never happen but can't be too careful, I guess?)

CharlesIC: I did the two IFs because it's the most robust solution in case none or two values are set (which should never happen but can't be too careful, I guess?)
It defaults to the redirect_to_service_error_path in any of these two cases

jakubmiarka:
It would be good if we could get rid of the two session variables (`selected_idp` and `selected_country`) and just use one. At the point we store the selected idp or country, we could indicate it within the object. It would save us deleting and setting different session variables.

PR #530, opened 2018-08-09
HUB-216: Build List of Services Refactor

chrisclayson:
Change service to read from new Config service endpoint that includes the entityId

### DIFF
--- a/spec/api_test_helper.rb
+++ b/spec/api_test_helper.rb
@@ -19,6 +19,10 @@ module ApiTestHelper
     config_api_uri('/config/transactions/enabled')
   end
 
+  def api_transactions_for_service_list_endpoint
+    config_api_uri('/config/transactions/for-service-list')
+  end
+
   def api_translations_endpoint(simple_id, locale)
     config_api_uri("/config/transactions/#{simple_id}/translations/#{locale}")
   end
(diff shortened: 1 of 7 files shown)


spec/api_test_helper.rb
```diff
@@ -29,13 +33,20 @@ def api_countries_endpoint(session_id)
...
+    ]
+    transactions_for_service_list = [
         { 'simpleId' => 'test-rp',      'entityId' => 'some-entity-id', 'serviceHomepage' => 'http://localhost:50130/test-rp', 'loaList' => ['LEVEL_2'] },
         { 'simpleId' => 'test-rp-noc3', 'entityId' => 'some-other-entity-id', 'serviceHomepage' => 'http://localhost:50130/test-rp-noc3', 'loaList' => ['LEVEL_2'] },
         { 'simpleId' => 'headless-rp',  'entityId' => 'some-entity-id', 'serviceHomepage' => 'http://localhost:50130/headless-rp', 'loaList' => ['LEVEL_2'] },
         { 'simpleId' => 'test-rp-with-continue-on-fail', 'entityId' => 'some-entity-id', 'serviceHomepage' => 'http://localhost:50130/test-rp-with-continue-on-fail', 'loaList' => ['LEVEL_2'] }
     ]
 
     stub_request(:get, api_transactions_endpoint).to_return(body: transactions.to_json, status: 200)
+    stub_request(:get, api_transactions_for_service_list_endpoint).to_return(body: transactions_for_service_list.to_json, status: 200)
```
jakubmiarka: this should be in a separate method. The method `stub_transactions_list` is called within the test which are expected to call the transactions_endpoint. This will now stub too much.

spec/api_test_helper.rb
```diff
@@ -29,13 +33,20 @@ def api_countries_endpoint(session_id)
 
   def stub_transactions_list
     transactions = [
+        { 'simpleId' => 'test-rp', 'serviceHomepage' => 'http://localhost:50130/test-rp', 'loaList' => ['LEVEL_2'] },
+        { 'simpleId' => 'test-rp-noc3', 'serviceHomepage' => 'http://localhost:50130/test-rp-noc3', 'loaList' => ['LEVEL_2'] },
+        { 'simpleId' => 'headless-rp', 'serviceHomepage' => 'http://localhost:50130/headless-rp', 'loaList' => ['LEVEL_2'] },
+        { 'simpleId' => 'test-rp-with-continue-on-fail', 'serviceHomepage' => 'http://localhost:50130/test-rp-with-continue-on-fail', 'loaList' => ['LEVEL_2'] }
+    ]
+    transactions_for_service_list = [
```
jakubmiarka: shouldn't it also include the taxon?

chrisclayson: This stubs what can't back from the new config service endpoint, which doesn't include the taxon. The taxon comes from the translations endpoint.

spec/spec_helper.rb
```diff
@@ -104,6 +104,7 @@
   config.before(:each) do |test|
     unless test.metadata[:skip_before]
       stub_transactions_list
+      stub_transactions_for_service_list
```
jakubmiarka: does this have to be called before each test? It should be sufficient to include this only in the tests testing this specific feature (which calls that endpoint)

jakubmiarka:
Also just realized, since the new endpoint is merged, we need to also add it to the stub, so frontend can still run in the stub mode:
https://github.com/alphagov/verify-frontend/blob/master/stub/api/stub_api.rb

PR #538, opened 2018-08-29
bau: upgrade selenium and capybara

chrisholmes:
Changes:
- Upgrade selenium and capybara
- Switch to using headless chrome via chromedriver
- Fixes to tests that were failing due to Chrome honoring our CSP rules
- Updated .travis.yml to use Chrome

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 language: ruby
 before_script:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-  - sleep 3 # give xvfb some time to start
+  # - "export DISPLAY=:99.0"
+  # - "sh -e /etc/init.d/xvfb start"
+  # - sleep 3 # give xvfb some time to start
 cache:
   directories:
     - $HOME/.phantomjs 
+sudo: required #this is needed for the chromedriver to work
 addons:
-  firefox: "45.9.0esr"
+  apt:
+    packages:
+      - google-chrome-stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,4 @@
 language: ruby
-before_script:
-  # - "export DISPLAY=:99.0"
-  # - "sh -e /etc/init.d/xvfb start"
-  # - sleep 3 # give xvfb some time to start
 cache:
   directories:
     - $HOME/.phantomjs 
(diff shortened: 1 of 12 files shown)


chrisholmes:
Chromium is available on the CI server so this should be ready to merge

.travis.yml
```diff
@@ -1,10 +1,13 @@
...
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-  - sleep 3 # give xvfb some time to start
+  # - "export DISPLAY=:99.0"
+  # - "sh -e /etc/init.d/xvfb start"
+  # - sleep 3 # give xvfb some time to start
 cache:
   directories:
     - $HOME/.phantomjs 
+sudo: required #this is needed for the chromedriver to work
```
rachelthecodesmith: I had a google around to see if this might introduce any security issues, can't see anything particular but is it something we'd want to run past sec ops?

spec/feature_helper.rb
```diff
@@ -7,33 +7,15 @@
 
 RACK_COOKIE_DATE_FORMAT = "%a, %d %b %Y".freeze
 
-if ENV['HEADLESS'] == 'true'
```
jakubmiarka: given we're no longer using the HEADLESS env variable, should it be also removed from `jenkins.sh`?

.travis.yml
jakubmiarka: I think the only side-effect is a longer boot time: https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

spec/feature_helper.rb
chrisholmes: yes

.travis.yml
chrisholmes: It's running in an isolated container on Travis and it's not part of the critical path so there's no issues.

.travis.yml
```diff
@@ -1,10 +1,13 @@
 language: ruby
 before_script:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-  - sleep 3 # give xvfb some time to start
+  # - "export DISPLAY=:99.0"
+  # - "sh -e /etc/init.d/xvfb start"
+  # - sleep 3 # give xvfb some time to start
```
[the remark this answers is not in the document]
chrisholmes: accidental

PR #539, opened 2018-08-29
EID-885: Revise error pages for failed sign-in and UAC

CharlesIC:
For eIDAS journeys, the new pages have links that take the user to the Verify/eIDAS picker if they end up in a error state after an AuthFailure or user account creation failure. The new policy endpoint is used to reset the user session back to SessionStartedState and enable them to retry.

Regular Verify users will be redirected to the start page on AuthFailure response from the IDP

To do achieve the above, we need to ensure that only either `selected_idp` or `selected_country `is set at any one time

[View the story on Jira](https://govukverify.atlassian.net/browse/EID-885)

### DIFF
--- a/app/controllers/partials/user_session_partial_controller.rb
+++ b/app/controllers/partials/user_session_partial_controller.rb
@@ -57,14 +57,4 @@ module UserSessionPartialController
   def store_selected_country_for_session(selected_country)
     session[:selected_provider] = SelectedProviderData.new(JourneyType::EIDAS, selected_country)
   end
-
-  def switch_to_verify_journey(selected_idp)
-    session.delete(:selected_country)
-    session[:selected_idp] = selected_idp
-  end
-
-  def switch_to_eidas_journey(selected_country)
-    session.delete(:selected_idp)
-    session[:selected_country] = selected_country
-  end
 end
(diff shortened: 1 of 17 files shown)


app/controllers/partials/user_session_partial_controller.rb
```diff
@@ -36,4 +36,14 @@ def selected_country
 
     Country.from_session(selected_country)
   end
+
+  def switch_to_verify_journey(selected_idp)
```
jakubmiarka: instead of the switching and deleting, could we not merge the custom variable to be one? e.g. `session[:selected_provider]` ? and when setting it, we could have the structure something like:
```
session[:selected_provider] = {
  type: 'country',
  selected: { entityId: 'country-whatever' ... }
}
```
and use the `type` to distinguish between Verify and eIDAS user?

CharlesIC: I guess we could but then anywhere we use the provider we'd need extra logic. Which is fine in itself but probably not in the scope of this story. @minhngocd - what do you think?

minhngocd: @jakubmiarka there are a lot of places that use :selected_idp and :selected_country in the session, so switching to use selected_provider will mean us having to make a decision whether or not the provider is an idp or country - I'm not sure whether that'd be easier.

jakubmiarka: well `session[:selected_idp]` is used across only 3 files (5 lines in total). It's not being manipulated or anything, it gets set on IDP selection and then read on return from IDP. So I think it's actually less work then making sure we call the switch function at the right time and the right places and delete the other session variable.

PR #540, opened 2018-08-29
HUB-217: Frontend accepts initiating request from idp

rachelthecodesmith:
-Created post endpoint 'initiate-single-idp'
-Accepts params: 'transaction_id', 'entity_idp_id' and 'uuid'
-Checks uuid of correct length, format and characters
-Checks transaction_id against front fed config and redirects to url from front fed config
-Checks entity_idp_id correct against front fed config
-Creates single idp cookie if all params were vaild
-Redirects to service list if any params invalid and does not save
cookie
-Sends action to piwik and sets journey type cvar to SINGLE_IDP

solo: @rachelthecodesmith


app/controllers/initiate_single_idp_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      redirect_to(rp_url)
+    else
+      redirect_to verify_services_path
+    end
+  end
+
+private
+
+  def valid_uuid?(uuid)
+    /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/.match?(uuid.to_s.downcase)
```
jakubmiarka: should we actually do `to_s.downcase` on the param itself, so we store the lowercased uuid in the cookie instead of just lowercasing when validating it?

app/controllers/partials/user_cookies_partial_controller.rb
```diff
@@ -33,8 +33,11 @@ def set_journey_hint_by_status(idp_entity_id, status)
     journey_hint_by_status_value = journey_hint_value || {}
     journey_hint_by_status_value[status] = idp_entity_id
 
-    cookies.encrypted[CookieNames::VERIFY_FRONT_JOURNEY_HINT] = { value: journey_hint_by_status_value.to_json,
-                                                                  expires: 18.months.from_now }
+    cookies.encrypted[CookieNames::VERIFY_FRONT_JOURNEY_HINT] = { value: journey_hint_by_status_value.to_json, expires: 18.months.from_now }
+  end
+
+  def set_single_idp_journey_cookie(data)
+    cookies.encrypted[CookieNames::SINGLE_IDP_JOURNEY] = { value: data.to_json }
```
jakubmiarka: According to the ticket the cookie should expire in 90 minutes.

app/models/cookie_names.rb
```diff
@@ -2,6 +2,7 @@ module CookieNames
   SESSION_COOKIE_NAME = '_verify-frontend_session'.freeze
   SESSION_ID_COOKIE_NAME = 'x_govuk_session_cookie'.freeze
   VERIFY_FRONT_JOURNEY_HINT = 'verify-front-journey-hint'.freeze
+  SINGLE_IDP_JOURNEY = 'single-idp-journey'.freeze
```
jakubmiarka: we could be consistent with other cookies and have the cookie name to begin with `verify-` maybe?

out/test/verify-frontend/controllers/redirect_to_rp_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
```
jakubmiarka: `/out` directory? 😕

spec/controllers/initiate_single_idp_controller_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      expect(subject).to redirect_to(SINGLE_IDP_ENABLED_RP_LIST_MOCK[VALID_TEST_RP]['url'])
+      expect(stub_piwik_request).to have_been_made.once
+      expect(cookies.encrypted[CookieNames::SINGLE_IDP_JOURNEY])
+        .to match({
+          transaction_id: VALID_TEST_RP, entity_idp_id: VALID_STUB_IDP, uuid: UUID_ONE
+        }.to_json)
+    end
+  end
+
+  context 'idp hits post without providing valid parameters' do
```
jakubmiarka: we should also have a test with missing parameters

config/routes.rb
```diff
@@ -52,6 +52,7 @@ def add_routes(routes_name)
   get '/SAML2/metadata/idp', to: 'metadata#identity_providers', as: :identity_provider_metadata
   if SINGLE_IDP_FEATURE
     get '/get-available-services', to: 'metadata#service_list', as: :services
+    post 'initiate-single-idp', to: 'initiate_single_idp#redirect_from_idp'
```
jakubmiarka: according to the tech spec, it should be `/initiate-single-idp-journey`

app/controllers/initiate_single_idp_controller.rb
```diff
@@ -0,0 +1,46 @@
+require 'partials/user_cookies_partial_controller'
+
+class InitiateSingleIdpController < ApplicationController
+  skip_before_action :validate_session
+  skip_before_action :set_piwik_custom_variables
+
+  def redirect_from_idp
+    transaction_id = params['transaction_id']
+    entity_idp_id = params['entity_idp_id']
+    uuid = params['uuid']
```
jakubmiarka: according to the tech spec, the three parameters should be called:

- `serviceId`
- `idpEntityId`
- `singleIdpJourneyIdentifier`

PR #541, opened 2018-08-30
bau: switch to firefox headless

chrisholmes:
There's been some issues with running chromium headless on Jenkins so
this changes the browser back to Firefox but with its new headless
functionality and geckodriver support.


chrisholmes:
That's fair. I'll raise a PR later with some configuration for that.

PR #546, opened 2018-08-31
EID-895 Tidy up stub IDP's that are not being used.

andrewchong:
-Changed references to stub-idp-demo to stub-idp-demo-one, including yml file name.


JHjava:
Can't find stub-idp-demo-one logo so is failing the frontend tests. If Karol is right and it isolated from the hub journey then maybe it won't be needed.

FelixMarcusMillne:
Fails on: 
       No route matches [GET] "/stub-logos/stub-idp-demo-one.png"

Guessing that the file in question is still stub-idp-demo.png in https://github.com/alphagov/verify-frontend/tree/master/public/stub-logos.

CharlesIC:
The frontend has its own "fake" hub-fed-config, called stub-api so it can be self-contained and run its tests without relying on anything outside the repo. It's fine to leave it as it is.

PR #549, opened 2018-09-03
HUB-213: Handover page for single idp journey

jakubmiarka:
- added a new view for the single IDP path (based on design from HUB-224)
- renamed the single IDP controller to be also used for redirect to IDP
- added the UUID to be passed on to the IDPs (from HUB-222)
- added a new piwik reporting
- updated the tests (switched to entity id instead of simple id)
- left some TODO comments which are dependant on other work in config service


app/views/shared/_continue_to_idp_form.html.erb
```diff
@@ -2,5 +2,6 @@
     <input name=SAMLRequest type=hidden>
     <input name=RelayState type=hidden>
     <input name=registration type=hidden>
+    <input name=singleIdpJourneyIdentifier type=hidden>
```
chrisclayson: I'm wondering if it might be best to include this only if the UUID is present (and feature flag enabled)? Could we potentially 'break' an IDP if we send them a form parm they're not expecting? Depends on their implementation, of course! Happy if you think it's fine!

PR #550, opened 2018-09-04
EID-885: New selected_provider storage variable in session

CharlesIC:
The new SelectedIdentityProviderData struct will contain information on the current journey type as well as the selected identity provider - either a country or a Verify IDP.

This is release 1/2 - in the first step we're writing to the new variable. In the second release, we can start reading from the new variable and remove the current selected_idp and selected_country


app/controllers/partials/user_session_partial_controller.rb
```diff
@@ -36,4 +52,18 @@ def selected_country
 
     Country.from_session(selected_country)
   end
+
+  def switch_to_verify_journey(selected_idp)
```
jakubmiarka: could we rename this to something like `select Verify IDP` or whatever makes sense (store_selected_idp?) so it's clear that it's required every time a user selects an IDP (or a country)? People with no context or knowledge of eIDAS might not know why they need to do a **switch**. It makes perfect sense to be switching when working within the eIDAS flow but not if we're dealing with one isolated/journey only. Just semantics ;)

app/controllers/partials/viewable_idp_partial_controller.rb
```diff
@@ -1,14 +1,14 @@
...
     end
   end
 
   def select_viewable_idp_for_loa(entity_id)
     for_viewable_idp(entity_id, current_identity_providers_for_loa) do |decorated_idp|
-      session[:selected_idp] = decorated_idp.identity_provider
+      switch_to_verify_journey(decorated_idp.identity_provider)
       yield decorated_idp
     end
   end
```
jakubmiarka: there's a new method on master which will be missing the update. Could you rebase or raise a new PR to fix it afterwards?

CharlesIC: Rebased 👍

app/controllers/partials/user_session_partial_controller.rb
CharlesIC: Renamed to `store_selected_idp_for_session`

app/models/selected_provider_data.rb
```diff
@@ -0,0 +1,10 @@
+SelectedProviderData = Struct.new(
```
chrisholmes: I think it miight help to make this a delegator to the identity_provider so that you don't have to call `decorated_idp.identity_provider` in places when you need to ready IDP data later on. You could use [DelegateClass](https://stackoverflow.com/a/38809382) for this or [SimpleDelegator](https://hashrocket.com/blog/posts/using-simpledelegator-for-your-decorators). I think it would help make the code less flexible and you won't need to change things in so many places going forward.

.rubocop.yml
```diff
@@ -5,3 +5,9 @@ Metrics/BlockLength:
     - '**/*.rake'
     - 'spec/**/*.rb'
     - 'config/initializers/**/*.rb'
+
```
chrisholmes: Can you raise this a separate PR

PR #554, opened 2018-09-05
BAU: Ensure we know what the working directory is in startup.sh

rhowe-gds:
startup.sh relies on relative paths. Ensure that we know what the current
working directory is.

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+cd "$(dirname "$0")"
+
 ./kill-service.sh
 
 if [ "$1" == '--stub-api' ]


startup.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+cd "$(dirname "$0")"
```
CharlesIC: What's the value of `"$0"` going to be when the script is run?

rhowe-gds: From `bash(1)`

```
If arguments remain after option processing, and neither the -c nor the -s option has been supplied, the first argument is assumed to be the name of a file containing shell commands. If bash is invoked in this fashion, $0 is set to the name of the file, and the positional parameters are set to the remaining arguments. 
```

So, the filename of the script as invoked by the user (could be a relative path, so dirname will turn it into an absolute one)

CharlesIC: That makes sense! 👍

PR #555, opened 2018-09-06
EID-895: Add stub-idp-demo-one and two to local frontend stub config

CharlesIC:
Following the removal of a bunch of obscure IDPs from frontend-federation-config, the frontend stub config is now out of sync and doesn't always show valid IDPs when running locally. Adding the two new IDPs - stub-idp-demo-one and new stub-idp-demo-two - will fix this


rhowe-gds:
What about other references to `stub-idp-demo`?

```
verify-frontend $ git grep stub-idp-demo
spec/api_test_helper.rb:      { 'simpleId' => 'stub-idp-demo', 'entityId' => 'demo-entity-id', 'levelsOfAssurance' => %w(LEVEL_2) },
stub/identity_providers.yml:single_idp_enabled: ['http://idcorp.com', 'http://stub_idp.acme.org/stub-idp-demo-one/SSO/POST']
stub/idp-rules/stub-idp-demo.yml:simpleIds: ["stub-idp-demo"]
```
and presumably if `simpleId` changes, federation config will need updating

CharlesIC:
These _should_ be fine as they're used in the stub mode and for the tests only - but I will double check. And anyway, when you run via the hub-startup script in ida-hub-acceptance-tests then it sets the config to the one in frontend-federation-config. But if you run without that script, then it uses the default local .env file which points to the stub config. So most of the time it's not an issue. And the tests don't need this config as they use the "fake" stub config.

PR #556, opened 2018-09-06
EID-795: Switch to using the new selected_provider session var

CharlesIC:
- Deleted the previously used `selected_idp` and `selected_country`
- Made the new `SelectedProviderData` object delegate to `identity_provider`, meaning that the new object can be used in virtually the same way in most cases and fewer code changes are required. The `identity_provider` attribute can either be an `IdentityProvider` or a `Country` - the new var will morph to act like the object it's initialised with
- Allow tests to set and read the new variable

### DIFF
--- a/app/controllers/partials/user_session_partial_controller.rb
+++ b/app/controllers/partials/user_session_partial_controller.rb
@@ -26,7 +26,6 @@ module UserSessionPartialController
   def current_selected_provider_data
     selected_provider_data = SelectedProviderData.from_session(session[:selected_provider])
     raise(Errors::WarningLevelError, 'No selected identity provider data in session') if selected_provider_data.nil?
-
     selected_provider_data
   end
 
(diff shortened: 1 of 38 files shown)


spec/feature_helper.rb
```diff
@@ -104,6 +104,18 @@ def set_loa_in_session(loa)
     )
   end
 
+  def set_selected_idp_in_session(selected_idp)
```
chrisholmes: 👍

spec/controller_helper.rb
```diff
@@ -10,3 +10,11 @@ def set_session_and_cookies_with_loa(loa_requested, transaction_simple_id = 'tes
   cookies[CookieNames::SESSION_COOKIE_NAME] = 'my-session-cookie'
   cookies[CookieNames::SESSION_ID_COOKIE_NAME] = 'my-session-id-cookie'
 end
+
+def set_selected_idp(selected_idp)
```
chrisholmes: 👍

app/models/identity_provider.rb
```diff
@@ -28,7 +28,7 @@ def hash
 
   def self.from_session(object)
```
chrisholmes: I think we could move these factory methods over to the delegator and it might read better. What do you think?

app/models/selected_provider_data.rb
```diff
@@ -1,7 +1,17 @@
...
+  def initialize(journey_type, identity_provider)
+    super(identity_provider)
+
+    @journey_type = journey_type
+    @identity_provider = identity_provider
+  end
+
+  def as_json(options = nil)
+    { journey_type: @journey_type.as_json(options), identity_provider: @identity_provider.as_json(options) }
+  end
```
chrisholmes: I think some helpers like #is_select_country? and `#is_selected_companies` might help with readability.

app/controllers/partials/user_session_partial_controller.rb
```diff
@@ -40,32 +40,20 @@ def user_journey_type
...
+    raise(Errors::WarningLevelError, 'No selected IDP in session') if user_journey_type != JourneyType::VERIFY
+    IdentityProvider.from_session(current_selected_provider_data.identity_provider)
   end
 
   def selected_country
-    selected_country = session[:selected_country]
-    raise(Errors::WarningLevelError, 'No selected Country in session') if selected_country.nil?
-
-    Country.from_session(selected_country)
+    raise(Errors::WarningLevelError, 'No selected Country in session') if user_journey_type != JourneyType::EIDAS
```
chrisholmes: I think you could moved this error handling behaviour into the new Delegator

app/models/selected_provider_data.rb
CharlesIC: Good idea, added!

PR #571, opened 2018-09-17
HUB-175: User can hit pause page without session and be redirected to rp

rachelthecodesmith:
Add view for users without valid session but with the journey hint
cookie state to allow them to proceed to the RP service start page.

solo: @rachelthecodesmith

### DIFF
--- a/app/controllers/partials/retrieve_federation_data_partial_controller.rb
+++ b/app/controllers/partials/retrieve_federation_data_partial_controller.rb
@@ -1,0 +1,35 @@
+module RetrieveFederationDataPartialController
+  def get_selected_rp_from_entity_id(list, entity_id)
+    return nil if list.nil?
+    list.select { |hash| hash.fetch('entityId') == entity_id }.first
+  end
+
+  def get_service_choice_url(list, transaction_id)
+    selected_rp = get_selected_rp_from_entity_id(list, transaction_id)
+    get_rp_homepage(selected_rp)
+  end
+
+  def get_rp_homepage(selected_rp)
+    return nil if selected_rp.nil?
+    selected_rp.fetch('serviceHomepage', nil)
+  end
+
+  def get_idp_choice(list, idp_entity_id)
+    return false if list.nil?
+    list.select { |idp| idp.entity_id == idp_entity_id }.first
+  end
+
+  def valid_transaction?(list, transaction_id)
+    list.detect { |hash| hash.fetch('entityId') == transaction_id }
+  end
+
+  def valid_idp_choice?(list, idp_entity_id)
+    return false if list.nil?
+    list.detect { |idp| idp.entity_id == idp_entity_id }
+  end
+
+  def get_transaction_entity_id(selected_rp)
+    return nil if selected_rp.nil?
+    selected_rp.fetch('entityId', nil)
+  end
+end
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -1,11 +1,14 @@
 require 'partials/user_cookies_partial_controller'
 require 'partials/journey_hinting_partial_controller'
 require 'partials/viewable_idp_partial_controller'
+require 'partials/retrieve_federation_data_partial_controller'
+
 
 
 class PausedRegistrationController < ApplicationController
   include JourneyHintingPartialController
   include ViewableIdpPartialController
+  include RetrieveFederationDataPartialController
 
   # Validate the session manually within the action, as we don't want the normal 'no session' page.
   skip_before_action :validate_session, except: :resume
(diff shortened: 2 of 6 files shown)


app/controllers/single_idp_journey_controller.rb
```diff
@@ -44,9 +46,15 @@ def redirect_from_idp
...
     idp_entity_id = params['idpEntityId']
     uuid = params['singleIdpJourneyIdentifier'].to_s.downcase
-    rp_url = get_service_choice_url(transaction_id)
 
-    if !rp_url.nil? && valid_idp_choice?(idp_entity_id, transaction_id) && valid_uuid?(uuid)
+    rp_url = get_service_choice_url(get_service_list, transaction_id)
+
+    if rp_url.nil?
+      redirect_to verify_services_path
+      return
```
chrisclayson: Could this explicit `return` be avoided? Maybe with an `elsif` below?

rachelthecodesmith: It already has an elsif below that it used to be part of but I agree this is also a bit ugly. I just figured this at least broke us out of things as early as possible. Will have another look

app/controllers/partials/retrieve_federation_data_partial_controller.rb
```diff
@@ -0,0 +1,39 @@
+module RetrieveFederationPartialController
```
jakubmiarka: the file is called retrieve_federation_*data*_partial_controller but the controller is called RetrieveFederationPartialController

app/controllers/paused_registration_controller.rb
```diff
@@ -1,22 +1,25 @@
...
   layout 'slides', except: :index
 
   def index
-    if session_is_valid
-      set_transaction_from_session
-      @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider)
-      render :with_user_session
+    if session_is_valid?
+      with_session
+    elsif journey_hint_cookie_state_present?
```
jakubmiarka: should this be true only if the last state is PENDING? we could then use `is_last_status?(status)` method and get rid off the `journey_hint_cookie_state_present?` below?

app/controllers/partials/retrieve_federation_data_partial_controller.rb
chrisclayson: Well spotted! I missed that in my review!

spec/api_test_helper.rb
```diff
@@ -37,10 +37,10 @@ def api_select_country_endpoint(session_id, country_code)
 
   def stub_transactions_list
     transactions = [
-        { 'simpleId' => 'test-rp', 'serviceHomepage' => 'http://localhost:50130/test-rp', 'loaList' => ['LEVEL_2'] },
-        { 'simpleId' => 'test-rp-noc3', 'serviceHomepage' => 'http://localhost:50130/test-rp-noc3', 'loaList' => ['LEVEL_2'] },
-        { 'simpleId' => 'headless-rp', 'serviceHomepage' => 'http://localhost:50130/headless-rp', 'loaList' => ['LEVEL_2'] },
-        { 'simpleId' => 'test-rp-with-continue-on-fail', 'serviceHomepage' => 'http://localhost:50130/test-rp-with-continue-on-fail', 'loaList' => ['LEVEL_2'] }
+        { 'simpleId' => 'test-rp', 'entityId' => 'http://www.test-rp.gov.uk/SAML2/MD', 'serviceHomepage' => 'http://localhost:50130/test-rp', 'loaList' => ['LEVEL_2'] },
```
jakubmiarka: does the endpoint really return the entityId?

app/controllers/paused_registration_controller.rb
```diff
@@ -34,21 +37,63 @@ def resume
...
+    @transaction = {
+        name: selected_rp,
+        homepage: get_rp_homepage(selected_rp)
+    }
+  end
+
+  def get_cookie_rp_from_service_list
+    last_rp_value = last_rp
+    return nil if last_rp_value.nil?
+    get_selected_rp_from_simple_id(get_service_list, last_rp_value)
```
jakubmiarka: the last_rp_value is an entityId if I'm not mistaken? not simpleId

app/controllers/paused_registration_controller.rb
```diff
@@ -34,21 +37,63 @@ def resume
...
     @transaction = {
         name: current_transaction.rp_name,
         homepage: current_transaction_homepage
     }
   end
+
+  def set_transaction_from_cookie(selected_rp)
+    @transaction = {
+        name: selected_rp,
+        homepage: get_rp_homepage(selected_rp)
```
jakubmiarka: could we replace this method with `selected_rp.fetch('serviceHomepage', nil)` directly? that way we don't need the separate method to wich basically just wraps the fetch?

app/controllers/paused_registration_controller.rb
```diff
@@ -34,21 +37,63 @@ def resume
...
   def set_transaction_from_session
     @transaction = {
         name: current_transaction.rp_name,
         homepage: current_transaction_homepage
     }
   end
+
+  def set_transaction_from_cookie(selected_rp)
+    @transaction = {
+        name: selected_rp,
```
jakubmiarka: does `selected_rp` represent an object? If so, should it be the `selected_rp.name`?

PR #573, opened 2018-09-19
HUB-264: Add Headless Redirect Links

chrisclayson:
Alter pause page so button redirects to RP's headless start page, if present.

### DIFF
--- a/app/controllers/authn_request_controller.rb
+++ b/app/controllers/authn_request_controller.rb
@@ -69,7 +69,7 @@ private
   def redirect_for_journey_hint(hint)
     if !cookies.encrypted[CookieNames::VERIFY_SINGLE_IDP_JOURNEY].nil? && SINGLE_IDP_FEATURE
       redirect_to continue_to_your_idp_path
-    elsif PAUSE_AND_RESUME_FEATURE && is_last_status?('PENDING') && hint != 'submission_confirmation'
+    elsif PAUSE_AND_RESUME_FEATURE && is_last_status?(PENDING_STATUS) && hint != 'submission_confirmation'
       redirect_to resume_registration_path
     else
       case hint
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -18,7 +18,7 @@ class PausedRegistrationController < ApplicationController
   def index
     if session_is_valid?
       with_session
-    elsif is_last_status?('PENDING')
+    elsif is_last_status?(PENDING_STATUS)
       with_cookie
     else
       render :without_user_session
@@ -44,7 +44,7 @@ private
   def get_idp_from_cookie
     last_idp_value = last_idp
     unless last_idp_value.nil?
-      return retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_sign_in, last_idp_value).first
+      retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_sign_in, last_idp_value).first
     end
   end
 
@@ -55,7 +55,7 @@ private
   end
 
   def with_cookie
-    selected_rp = get_rp_details
+    selected_rp = get_rp_details(last_rp)
     set_transaction_from_cookie(selected_rp)
     enabled_idp_list = get_idp_list(last_rp)
     idp = get_idp_choice(enabled_idp_list, last_idp)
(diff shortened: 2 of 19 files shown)


app/controllers/authn_request_controller.rb
```diff
@@ -66,10 +67,14 @@ def set_transaction_homepage(transaction_homepage)
     session[:transaction_homepage] = transaction_homepage
   end
 
+  def set_transaction_headless_startpage(transaction_headless_homepage)
+    session[:transaction_headless_startpage] = transaction_headless_homepage
```
jakubmiarka: I don't like storing it in the session :( ... This means we increase the session size for all the users despite then being used by a handful of pausing users. Also people without a session, but with a cookie won't have the headless start page.

app/controllers/paused_registration_controller.rb
```diff
@@ -66,14 +66,16 @@ def with_cookie
...
+      start_page: preferred_start_page
     }
   end
 
   def set_transaction_from_cookie(selected_rp)
     @transaction = {
       name: get_translated_service_name(selected_rp.simple_id),
-      homepage: selected_rp.transaction_homepage
+      homepage: selected_rp.transaction_homepage,
+      start_page: preferred_start_page
```
jakubmiarka: this will break, as there is no current session when retrieving the stuff from the cookie?

spec/features/user_sends_authn_response_spec.rb
```diff
@@ -99,6 +99,7 @@
 
   it 'will redirect the user to /paused-registration on pending response' do
```
jakubmiarka: I know it's not your change, but it should say `/paused`

app/views/test_journey_hint_cookie/index.html.erb
```diff
@@ -3,5 +3,6 @@ Current value: <%= @current_cookie %>
   <%= text_field_tag('entity-id') %>
   <%= text_field_tag('locale') %>
   <%= text_field_tag('status') %>
+  <%= text_field_tag('rp-entity-id') %>
```
jakubmiarka: handy! thanks for adding that, I did forget

spec/feature_helper.rb
```diff
@@ -132,11 +132,13 @@ def query_params
     current_uri.query ? CGI::parse(current_uri.query) : {}
   end
 
-  def set_journey_hint_cookie(entity_id, status = nil, locale = 'en')
+  def set_journey_hint_cookie(entity_id, status = nil, locale = 'en', rp_entity_id = nil)
     visit '/test-journey-hint'
     fill_in 'entity-id', with: entity_id
     fill_in 'status', with: status
     fill_in 'locale', with: locale
```
jakubmiarka: we should revisit this and figure out why we're setting the locale.. it shouldn't make any difference to the cookie...

It would make the method simpler if we didn't have to keet passing the 'en' in there and we never pass 'cy' anyway 🤐

spec/features/user_visits_paused_registration_page_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    context 'and verifying for a service WITHOUT a headless start page' do
+      it 'goes to renders page with a button linking to the services headless start page' do
+        stub_transaction_details(headlessStartpage: nil)
+        visit '/paused'
+
+        expect(page).to have_link(t('hub.paused_registration.with_session.continue_verifying', idp_name: idp_display_name), href: 'http://www.test-rp.gov.uk/')
+      end
+    end
+  end
+end
```
jakubmiarka: would be worth testing a welsh version of the page? just to check it renders properly... maybe with expectation for the service name to be in welsh too?

spec/features/user_visits_paused_registration_page_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    context 'and is verifying for a service WITH headless start page' do
+      it 'goes to renders page with a button linking to the services headless start page' do
+        stub_transaction_details
+        visit '/paused'
+
+        expect(page).to have_link(t('hub.paused_registration.with_session.continue_verifying', idp_name: idp_display_name), href: 'http://www.test-rp.gov.uk/success')
+      end
+    end
+    context 'and verifying for a service WITHOUT a headless start page' do
+      it 'goes to renders page with a button linking to the services headless start page' do
```
jakubmiarka: should the name of the test be `... linking to services start page` and not headless?

spec/features/user_visits_paused_registration_page_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        stub_transaction_details
+        set_session_and_session_cookies!
+        visit '/paused'
+
+        expect(page).to have_link(t('hub.paused_registration.with_session.continue_verifying', idp_name: idp_display_name), href: 'http://www.test-rp.gov.uk/success')
+      end
+    end
+    context 'and verifying for a service WITHOUT a headless start page' do
+      it 'goes to renders page with a button linking to the services headless start page' do
```
jakubmiarka: ...~~headless~~...

PR #574, opened 2018-09-19
BAU: Bug fix for rp display name

rachelthecodesmith:
Fixed error where I had put the endpoint rather than the method for the
endpoint for the rp name.

solo: @rachelthecodesmith

### DIFF
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -84,7 +84,7 @@ private
   end
 
   def get_translated_service_name(simple_id)
-    CONFIG_PROXY.transaction_translation_data_endpoint(simple_id, params[:locale])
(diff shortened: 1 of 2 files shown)


app/controllers/paused_registration_controller.rb
```diff
@@ -84,7 +84,7 @@ def get_rp_details
   end
 
   def get_translated_service_name(simple_id)
-    CONFIG_PROXY.transaction_translation_data_endpoint(simple_id, params[:locale])
+    CONFIG_PROXY.get_transaction_translations(simple_id, params[:locale]).name
```
jakubmiarka: `CONFIG_PROXY.get_transaction_translations(simple_id, params[:locale]).fetch(:name, nil)`

PR #578, opened 2018-09-24
HUB-293: Update reporting for P&R journey

jakubmiarka:
Updating the reporting to add the custom variable for the resume page.
Changed the authn response reporting to report RESUME_WITH_IDP instead of SIGN_IN_WITH_IDP.
Switched the Resume button to start using resume method and reporting instead of the sign-in logic.

### DIFF
--- a/app/controllers/partials/analytics_partial_controller.rb
+++ b/app/controllers/partials/analytics_partial_controller.rb
@@ -12,6 +12,13 @@ module AnalyticsPartialController
     custom_variables_for_img_tracker
   end
 
+  def set_additional_piwik_custom_variable(key, value)
+    @piwik_custom_variables.push(Analytics::CustomVariable.build_for_js_client(key, value))
+
+    @piwik_custom_variables_img_tracker =
+      @piwik_custom_variables_img_tracker.merge(Analytics::CustomVariable.build(key, value))
+  end
+
   def delete_new_visit_flag
     http_redirect = 302
     http_see_other = 303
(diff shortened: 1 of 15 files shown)


app/controllers/partials/analytics_partial_controller.rb
```diff
@@ -12,6 +12,13 @@ def set_piwik_custom_variables
     custom_variables_for_img_tracker
   end
 
+  def set_additional_piwik_custom_variable(key, value)
```
javindo: As a general note, it's best to avoid methods like this which are doing nothing other than applying side effects to external references. Not requesting a change here as we're fairly guilty of it throughout the rails app but we should have a discussion about this offline, could be a good firebreak refactor for example.

app/controllers/paused_registration_controller.rb
```diff
@@ -72,7 +88,8 @@ def set_transaction_from_session
     }
   end
 
-  def set_transaction_from_cookie(selected_rp)
+  def set_transaction_from_cookie
+    selected_rp = get_rp_details(last_rp)
```
javindo: What will happen if get_rp_details returns nil when last_rp is nil? Will the homepage setter below just set it to nil or will it throw an exception?

jakubmiarka: to be fair, I just moved this to be more consistent with the other method (set_transaction_from_session), I haven't made any changes... But last_rp shouldn't be nil at this point, as there is a check on the upstream `index` method `is_last_status?(PENDING_STATUS)`

javindo: SGTM, thanks

PR #581, opened 2018-09-25
HUB-294: Delete single idp cookie after any SAML response.

rachelthecodesmith:
Add method to delete single idp journey cookie after all saml responses.

AC's:
The single IDP cookie is deleted when the user is on a single-idp journey and we receive any SAML response (SUCCESS, FAIL, PENDING ... ) from an IDP

The single IDP cookie is deleted when the user chooses a different IDP and we receive any SAML response from that IDP (SUCCESS, FAIL, PENDING ... ) from an IDP

The SIDP cookie is retained when the user closes the browser during the IDP portion of the journey

The SIDP cookie is retained when the user clicks Continue on the handover page, then clicks back from the IDP journey

solo: @rachelthecodesmith

### DIFF
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -50,6 +50,7 @@ private
   def handle_idp_response(status, response)
     analytics_reporters(status, response)
     set_journey_status(status)
+    clear_single_idp_cookie
     redirect_to idp_redirects(status, response)
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/authn_response_controller.rb
```diff
@@ -109,4 +110,8 @@ def country_redirects(status, response)
       FAILED => is_registration ? failed_registration_path : failed_country_sign_in_path
     }.fetch(status)
   end
+
+  def clear_single_idp_cookie
+    cookies.delete CookieNames::VERIFY_SINGLE_IDP_JOURNEY
```
javindo: Will this method be ok if the cookie doesn't exist?

rachelthecodesmith: Yep, I added a test for that just in case controllers/authn_response_controller_spec.rb:142

PR #583, opened 2018-09-27
CVQ2-225: Update content for empty segment

karlbaker02:
This PR updates the content that we show for an empty segment in the frontend. It also renames the page and endpoint from `no-mobile-phone` to `verify-will-not-work-for-you`.

Welsh translations are queued and should be updated in the frontend once available.

Solo: @karlbaker02


config/locales/cy.yml
```diff
@@ -15,7 +15,7 @@ cy:
     other_identity_documents: dogfennau-hunaniaeth-eraill
     select_phone: dewis-ffon
     will_it_work_for_me: a-fydd-yn-gweithio-i-mi
-    no_mobile_phone: dim-ffon-symudol
+    verify_will_not_work_for_you: dim-ffon-symudol
```
jakubmiarka: I would suggest putting the english version here, as the welsh is now 'incorrect' and makes it easier to spot the missing translation (if the welsh translations are delayed or done by other team)

karlbaker02: Ah gotcha, so it would be `verify-will-not-work-for-you` for both en and cy until cy translations come through?

karlbaker02: Would you suggest that we do this here?

jakubmiarka: Given the current _low usage_ of this page, I think we can leave it, as long as we're expecting the translations soonish... ;)

karlbaker02: Hopefully by 03/10.

PR #585, opened 2018-09-28
HUB-185: Add Resume Links to Pause and Resume

chrisclayson:
Add new endpoint to PausedRegistrationController to start a new journey based on IDP name passed in URL.

### DIFF
--- a/app/controllers/partials/journey_hinting_partial_controller.rb
+++ b/app/controllers/partials/journey_hinting_partial_controller.rb
@@ -16,6 +16,16 @@ module JourneyHintingPartialController
     journey_hint_value.nil? ? nil : journey_hint_value['SUCCESS']
   end
 
+  def resume_link?
+    journey_hint = journey_hint_value
+    !(journey_hint.nil? || journey_hint.fetch('RESUMELINK', nil).nil?)
+  end
+
+  def resume_link_idp
+    journey_hint = journey_hint_value
+    journey_hint.nil? ? nil : journey_hint.dig('RESUMELINK', 'IDP')
+  end
+
   def last_status
     journey_hint = journey_hint_value
     journey_hint.nil? ? nil : journey_hint['STATE']
@@ -43,5 +53,9 @@ module JourneyHintingPartialController
 
   def retrieve_decorated_singleton_idp_array_by_entity_id(providers, entity_id)
     IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(providers.select { |idp| idp.entity_id == entity_id })
+  end
+
+  def retrieve_decorated_singleton_idp_array_by_simple_id(providers, simple_id)
+    IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(providers.select { |idp| idp.simple_id == simple_id })
   end
 end
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -3,17 +3,19 @@ require 'partials/journey_hinting_partial_controller'
 require 'partials/viewable_idp_partial_controller'
 require 'partials/retrieve_federation_data_partial_controller'
 require 'partials/idp_selection_partial_controller'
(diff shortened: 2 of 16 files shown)


app/controllers/partials/journey_hinting_partial_controller.rb
```diff
@@ -16,6 +16,21 @@ def success_entity_id
     journey_hint_value.nil? ? nil : journey_hint_value['SUCCESS']
   end
 
+  def magic_link?
+    journey_hint = journey_hint_value
+    ! (journey_hint.nil? || journey_hint.fetch('MAGIC', nil).nil?)
```
jakubmiarka: I wouldn't use `MAGIC` but something like 'assisted_resume' maybe? I'll have a look at the rest of the PR next week ;)

spec/features/user_visits_magic_link_paused_page_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  context 'and the IDP specified in the URL is valid' do
+    it 'renders page with a list of available services' do
+      stub_transaction_details
+      set_session_and_session_cookies!
+      visit '/paused/stub-idp-one'
+
+      expect(page).to have_content(t('hub.paused_registration.from_magic.heading', idp_name: idp_display_name))
+    end
+  end
```
jakubmiarka: I would add another context if the IDP specified in the URL is invalid

app/controllers/paused_registration_controller.rb
```diff
@@ -59,9 +76,14 @@ def session_is_valid?
...
-    unless last_idp_value.nil?
-      retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_sign_in, last_idp_value).first
+    magic_link_idp_value = magic_link_idp
+    if magic_link_idp_value.nil?
+      last_idp_value = last_idp
+      unless last_idp_value.nil?
+        retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_sign_in, last_idp_value).first
+      end
+    else
+      retrieve_decorated_singleton_idp_array_by_simple_id(current_identity_providers_for_sign_in, magic_link_idp_value).first
```
jakubmiarka: just wondering if we could simplify this, getting a bit complex and less readable when we mix `if`, `unless`, `else`... maybe move the last_idp_value check to the same line with magic_link_idp_value?

spec/features/user_visits_magic_link_paused_page_spec.rb
chrisclayson: That scenario already covered in unit test ... it tests for the rendering of the different template, which contains no interpolation.

config/locales/en.yml
```diff
@@ -541,6 +542,17 @@ en:
             <li>bookmark this page</li>
           </ul>
           <p>You'll need the email or username and password you used to create an account.</p>
+      from_magic:
+        title: "Verification Paused %{idp_name}"
```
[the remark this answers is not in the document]
chrisclayson: Have posed the question to Sarah.

app/controllers/partials/journey_hinting_partial_controller.rb
```diff
@@ -16,6 +16,16 @@ def success_entity_id
     journey_hint_value.nil? ? nil : journey_hint_value['SUCCESS']
   end
 
+  def resume_link?
+    journey_hint = journey_hint_value
+    ! (journey_hint.nil? || journey_hint.fetch('RESUMELINK', nil).nil?)
```
jakubmiarka: I think codacy complained about the space between `!` and `(`

spec/controllers/paused_registration_controller_spec.rb
```diff
@@ -66,6 +67,49 @@
     end
   end
 
+  context 'user is directed to a magic link from an IDP e-mail with a valid IDP simple_id' do
```
jakubmiarka: I would drop the `magic` word from tests too, but 🤷‍♂️

PR #590, opened 2018-10-08
Add flags for registration and authentication enabled

javindo:


javindo:
Merging to check against joint. Will revert if failing.

PR #598, opened 2018-10-22
HUB-292: Show Single IDP users registration success page

chrisclayson:
Update AuthnResponseController to redirect user appropriately on Single IDP journeys.

### DIFF
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -96,7 +96,7 @@ private
   def idp_redirects(status, response)
     is_registration = response.is_registration
     {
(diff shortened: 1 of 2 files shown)


app/controllers/authn_response_controller.rb
```diff
@@ -129,4 +129,8 @@ def country_redirects(status, response)
   def clear_single_idp_cookie
     cookies.delete CookieNames::VERIFY_SINGLE_IDP_JOURNEY
   end
+
+  def single_idp?
```
jakubmiarka: There's already an existing method `def journey_type?(journey_type)` let's reuse that one.

jakubmiarka: Line 115

PR #602, opened 2018-10-26
Refactor the redirect_from_idp method

daniele-occhipinti:
This PR does two small things:
* fixes this issue: we were not checking that the request params were not nil before starting the request verification (that was causing the `valid_request?` method to fail)
* prepares the ground for adding logging, mainly by restructuring the `redirect_from_idp` method to be a "series of filters", which will be able to easily trigger a specific log message if something goes wrong.

A note:
The `redirect_from_idp` method has got a few return points - which may not be ideal - however, in this case, I feel that is making the flow more understandable and will make it easier to add specific logging w/o using nested if statements.

### DIFF
--- a/app/controllers/single_idp_journey_controller.rb
+++ b/app/controllers/single_idp_journey_controller.rb
@@ -44,18 +44,24 @@ class SingleIdpJourneyController < ApplicationController
   end
 
   def redirect_from_idp
-    transaction_id = params['serviceId']
-    idp_entity_id = params['idpEntityId']
-    uuid = params['singleIdpJourneyIdentifier'].to_s.downcase
-
-    rp_url = get_service_choice_url(get_service_list, transaction_id)
-
-    if !rp_url.nil? && valid_request?(transaction_id, idp_entity_id, uuid)
-      save_to_cookie(transaction_id, idp_entity_id, uuid)
-      FEDERATION_REPORTER.report_started_single_idp_journey(request)
-      redirect_to(rp_url)
-    else
+    if params_are_missing(%w(serviceId idpEntityId singleIdpJourneyIdentifier))
       redirect_to verify_services_path
+    else
+      transaction_id = params['serviceId']
+      idp_entity_id = params['idpEntityId']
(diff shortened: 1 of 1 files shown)


app/controllers/single_idp_journey_controller.rb
```diff
@@ -44,19 +44,32 @@ def continue_ajax
...
+    if rp_url.nil?
+      redirect_to verify_services_path
+      return
+    end
+
+    is_valid_request = valid_request?(transaction_id, idp_entity_id, uuid)
+
+    if !is_valid_request
       redirect_to verify_services_path
+      return
```
jakubmiarka: `redirect_to verify_services_path`
`return`
is duplicated 3 times in this method...  Let's try to avoid doing `return` and use `else`. It should make it more readable too.

app/controllers/single_idp_journey_controller.rb
```diff
@@ -44,19 +44,32 @@ def continue_ajax
   end
 
   def redirect_from_idp
+    if !verify_params_are_not_null(%w(serviceId idpEntityId singleIdpJourneyIdentifier))
```
jakubmiarka: could we rename the method name so that we don't have to use `!` ?

app/controllers/single_idp_journey_controller.rb
```diff
@@ -44,19 +44,32 @@ def continue_ajax
...
-      save_to_cookie(transaction_id, idp_entity_id, uuid)
-      FEDERATION_REPORTER.report_started_single_idp_journey(request)
-      redirect_to(rp_url)
-    else
+    if rp_url.nil?
+      redirect_to verify_services_path
+      return
+    end
+
+    is_valid_request = valid_request?(transaction_id, idp_entity_id, uuid)
```
jakubmiarka: do we need to assign it to the variable? could we not just call the `valid_request?` directly in the if statement before?

daniele-occhipinti:
@jakubmiarka I think I have addressed all your points.

PR #603, opened 2018-10-28
BAU: Add various configuration defaults

rhowe-gds:
Currently we configure a bunch of things externally, but they are the same in
all environments. Make those things have a sane default.

### DIFF
--- a/config/initializers/00_configuration.rb
+++ b/config/initializers/00_configuration.rb
@@ -1,17 +1,19 @@
 require 'configuration'
 
+FED_CONFIG_DIR = '/usr/share/verify-frontend-federation-config'.freeze
+
 CONFIG = Configuration.load! do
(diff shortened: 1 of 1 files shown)


config/initializers/00_configuration.rb
```diff
@@ -1,17 +1,19 @@
 require 'configuration'
 
+FedConfigDir = '/usr/share/verify-frontend-federation-config'.freeze
```
jakubmiarka: constants should be in SCREAMING_SNAKE_CASE: https://github.com/alphagov/govuk-lint/blob/1c61c92c409f95a53d537fba1c107d099e163dd5/configs/rubocop/gds-ruby-styleguide.yml#L234

Linter is disabled for this file, so that's why it didn't get highlighted.

rhowe-gds: OK, but if this pushes me over the line length rubocop check, I'm coming for you!

PR #605, opened 2018-10-29
BAU: Supply a default for `WHITE_LOGO_DIRECTORY`

rhowe-gds:
We configure this externally, but it always has the same value. Supply a
sensible default.


rhowe-gds:
Superseded by https://github.com/alphagov/verify-frontend/pull/606

PR #608, opened 2018-10-31
BAU: Enable the single IDP feature by default

rhowe-gds:
This is enabled everywhere except the performance testing environment and our
vagrant test boxes. Enable it everywhere by default.


rhowe-gds:
I'm quite against configuration for configuration's sake - if the default in frontend is true and we want to make it false in some environment we can add that configuration then, when we need it.

jakubmiarka:
I see your point. My concern is that it's no longer just a config/feature flag which anyone can toggle if there is an urgent need, but will need to 'implement' the config in puppet first (or toggle it for all environments).

PR #612, opened 2018-11-05
EID-1053: Fixing bug in eIDAS journey when user presses back after authentication fails.

tomrosier:
Added new variable to the session so we can identify when the user has reached the failed to authenticate page.

When returning to return-to-country view it resets the session if we have failed before
Fixed the test runner to not upgrade firefox as there is currently a bug stopping it from function.

### DIFF
--- a/app/controllers/redirect_to_country_controller.rb
+++ b/app/controllers/redirect_to_country_controller.rb
@@ -5,6 +5,11 @@ class RedirectToCountryController < ApplicationController
   before_action :ensure_session_eidas_supported
 
   def choose_a_country_submit
+    if session[:failed_country_sign_in]
+      restart_journey
+      session.delete(:failed_country_sign_in)
+    end
+
     decorated_country = decorated_country(params[:country])
     if decorated_country.viewable?
       select_country(decorated_country)
--- a/app/controllers/failed_sign_in_controller.rb
+++ b/app/controllers/failed_sign_in_controller.rb
@@ -4,6 +4,7 @@ class FailedSignInController < ApplicationController
   end
 
   def country
+    session[:failed_country_sign_in] = true
     @entity = COUNTRY_DISPLAY_DECORATOR.decorate(selected_country)
     @other_ways_description = current_transaction.other_ways_description
     @other_ways_text = current_transaction.other_ways_text
--- a/app/controllers/partials/user_session_partial_controller.rb
+++ b/app/controllers/partials/user_session_partial_controller.rb
@@ -1,6 +1,7 @@
 module UserSessionPartialController
   def validate_session
     validation = session_validator.validate(cookies, session)
+
     unless validation.ok?
       logger.info(validation.message)
       render_error(validation.type, validation.status)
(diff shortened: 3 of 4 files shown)


app/controllers/redirect_to_country_controller.rb
```diff
@@ -5,6 +5,11 @@ class RedirectToCountryController < ApplicationController
   before_action :ensure_session_eidas_supported
 
   def choose_a_country_submit
+    if session[:failed_country_sign_in]
+      restart_journey
+      session[:failed_country_sign_in] = false
```
jakubmiarka: I'd prefer `session.delete(:failed_country_sign_in)` ... there's no need keeping it in the session

CharlesIC:
@jakubmiarka What do you think of this fix?
This addresses the case when we receive the authn failed response from a country and we show the user the "Your identity scheme in [country] couldn't verify you." If they then click the back button instead of clicking the journey restart link and go back to the country picker page and select a country again, they will receive the "Sorry something went wrong" page due to session state mismatch.
So when selecting a country, we need to have a way of knowing whether the user has already been on the authentication failed page. Hence the new session variable.

jakubmiarka:
It feels a bit of edge case to me. When the user goes back from hub back to IDP/country, their session should be invalid in the IDP/country. Ideally it should be implemented in a same way we check this for UK IDP journeys.

CharlesIC:
@jakubmiarka Where do we check that for IDP journeys? What's the mechanism for that?

jakubmiarka:
@CharlesIC I believe it might be policy allowing it?

CharlesIC:
@jakubmiarka  IDP journeys allow users to go back and sign in again without seeing any errors. This is implemented using the `IdpSelectingState` interface - so even `IdpSelectedState` and `AuthnFailureState` implement that interface and therefore users can select an IDP again even if they've already done it. So for countries, we're manually resetting the session to `SessionStartedState` so we can select again. But we do actually have the `EidasCountrySelectedState` so we could use that in the same way for this scenario as well. @jakubmiarka do you think that's better than doing that in the frontend? (it would be an easy change in Policy)

jakubmiarka:
I think the policy is the place where this logic should go. It's policy's job to enforce these transitions. Adding the logic to frontend doesn't feel right, especially when we have a working pattern for these journeys for the UK IDPs.

CharlesIC:
Yep, agree. We'll implement this in Policy. We can still merge the acceptance tests fix from this PR, though. This will fix https://trello.com/c/5tdw3vFh/135-verify-acceptance-tests-are-broken-with-the-latest-firefox-63

jakubmiarka:
It's not a real fix, if people already updated, this will not fix it - it's too late 👿 . Also, the issue will appear in the `verify-acceptance-tests` as well. We can merge the 'workaround' in the meantime for sure, and remove once we find the right solution.

CharlesIC:
Yep, I meant a workaround. Good for now I guess

CharlesIC:
@tomrosier Can we force a specific version of Firefox? `62.0.3` is is the latest known working version

tomrosier:
Abandoning, Changes should be made in hub policy rather than in the front end.

PR #614, opened 2018-11-09
EID-1100: Handle non standard HTTP methods.

tomrosier:


lib/store_session_id.rb
```diff
@@ -4,10 +4,15 @@ def initialize(app)
   end
 
   def call(env)
+    if ["POST", "GET"].exclude? env['REQUEST_METHOD']
```
jakubmiarka: linter is going to shout because of the array syntax. Plus we also use PUT method for our ajax calls...

tomrosier: Good shout 👍

jakubmiarka: there might a better way of doing it. It doesn't feel right listing the 'allowed' methods.. Why we don't just capture the unsupported ones?

jakubmiarka: In that case I would be tempted to do the whitelist on the server (nginx) and not even let the request to reach the app? 🤷‍♂️

tomrosier: @jakubmiarka would this be preferred? https://github.com/alphagov/verify-puppet/pull/1460/files

tomrosier:
Abandoned, in-favour of nginx change.

PR #616, opened 2018-11-21
redirect for non matching journey

Rosa-Fox:
Co-authored-by: Olakunle Jegede <olakunle.jegede@digital.cabinet-office.gov.uk>
Co-authored-by: Amrit Sidhu <amrit.sidhu@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -76,7 +76,7 @@ private
   def analytics_message(status, response)
     {
       SUCCESS => "Success - #{user_state(response)} at LOA #{response.loa_achieved}",
-      MATCHING_JOURNEY_SUCCESS => "Success - #{user_state(response)} at LOA #{response.loa_achieved}",
+      MATCHING_JOURNEY_SUCCESS => "Success Matching Journey - #{user_state(response)} at LOA #{response.loa_achieved}",
       NON_MATCHING_JOURNEY_SUCCESS => "Success Non Matching Journey - #{user_state(response)} at LOA #{response.loa_achieved}",
       CANCEL => "Cancel - #{user_state(response)}",
       FAILED_UPLIFT => "Failed Uplift - #{user_state(response)}",
@@ -101,12 +101,16 @@ private
     is_registration || journey_type?(SINGLE_IDP_JOURNEY_TYPE) ? confirmation_path : response_processing_path
   end
 
+  def path_for_success_non_matching(is_registration)
+    is_registration || journey_type?(SINGLE_IDP_JOURNEY_TYPE) ? confirmation_non_matching_journey_path : redirect_to_service_signing_in_path
+  end
+
   def idp_redirects(status, response)
     is_registration = response.is_registration
     {
       SUCCESS => path_for_success(is_registration),
       MATCHING_JOURNEY_SUCCESS => path_for_success(is_registration),
-      NON_MATCHING_JOURNEY_SUCCESS => path_for_success(is_registration),
+      NON_MATCHING_JOURNEY_SUCCESS => path_for_success_non_matching(is_registration),
       CANCEL => is_registration ? cancelled_registration_path : start_path,
       FAILED_UPLIFT => failed_uplift_path,
       PENDING => paused_registration_path,
--- a/app/controllers/authn_response_controller.rb
+++ b/app/controllers/authn_response_controller.rb
@@ -10,6 +10,8 @@ class AuthnResponseController < SamlController
   RESUMING_STATE = 'RESUME_WITH_IDP'
   SINGLE_IDP_STATE = 'SINGLE_IDP'
   SUCCESS = 'SUCCESS'
+  MATCHING_JOURNEY_SUCCESS = 'MATCHING_JOURNEY_SUCCESS'
+  NON_MATCHING_JOURNEY_SUCCESS = 'NON_MATCHING_JOURNEY_SUCCESS'
   CANCEL = 'CANCEL'
   FAILED = 'FAILED'
   FAILED_UPLIFT = 'FAILED_UPLIFT'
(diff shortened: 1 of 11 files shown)


app/controllers/authn_response_controller.rb
```diff
@@ -74,6 +76,8 @@ def set_journey_status(status)
   def analytics_message(status, response)
     {
       SUCCESS => "Success - #{user_state(response)} at LOA #{response.loa_achieved}",
+      MATCHING_JOURNEY_SUCCESS => "Success Matching Journey - #{user_state(response)} at LOA #{response.loa_achieved}",
+      NON_MATCHING_JOURNEY_SUCCESS => "Success Non Matching Journey - #{user_state(response)} at LOA #{response.loa_achieved}",
```
alan-gds: Have we checked with the analytics team that they're okay with us changing what we put in the analytics logs?  It's fine for us to be renaming the SUCCESS constant internally, but using a different message here is a separate issue; I think we should probably be sticking with "Success - #{user_state(response)} at LOA #{response.loa_achieved}".  It might also be worth talking to the analytics team about what message we should be using in non-matching journeys.

app/controllers/authn_response_controller.rb
```diff
@@ -93,10 +97,16 @@ def user_state(response)
...
+    is_registration || journey_type?(SINGLE_IDP_JOURNEY_TYPE) ? confirmation_path : response_processing_path
+  end
+
   def idp_redirects(status, response)
     is_registration = response.is_registration
     {
-      SUCCESS => is_registration || journey_type?(SINGLE_IDP_JOURNEY_TYPE) ? confirmation_path : response_processing_path,
+      SUCCESS => path_for_success(is_registration),
+      MATCHING_JOURNEY_SUCCESS => path_for_success(is_registration),
+      NON_MATCHING_JOURNEY_SUCCESS => path_for_success(is_registration),
```
alan-gds: This is so much neater than what was there before!

app/models/matching_outcome_response.rb
```diff
@@ -7,6 +7,7 @@ class MatchingOutcomeResponse < Api::Response
   SEND_USER_ACCOUNT_CREATED_RESPONSE_TO_TRANSACTION = 'SEND_USER_ACCOUNT_CREATED_RESPONSE_TO_TRANSACTION'.freeze
   SHOW_MATCHING_ERROR_PAGE = 'SHOW_MATCHING_ERROR_PAGE'.freeze
   USER_ACCOUNT_CREATION_FAILED = 'USER_ACCOUNT_CREATION_FAILED'.freeze
+  SEND_SUCCESSFUL_NON_MATCHING_RESPONSE_TO_TRANSACTION = 'SEND_SUCCESSFUL_NON_MATCHING_RESPONSE_TO_TRANSACTION'.freeze
```
alan-gds: Ditto - please put this with the other success ones.

app/models/matching_outcome_response.rb
```diff
@@ -17,6 +18,7 @@ class MatchingOutcomeResponse < Api::Response
     SEND_USER_ACCOUNT_CREATED_RESPONSE_TO_TRANSACTION,
     SHOW_MATCHING_ERROR_PAGE,
     USER_ACCOUNT_CREATION_FAILED,
+    SEND_SUCCESSFUL_NON_MATCHING_RESPONSE_TO_TRANSACTION
```
alan-gds: Ditto - please put this with the other success ones.

app/controllers/authn_response_controller.rb
alan-gds: Rosa and Ola spoke to Anne and she confirmed that the analytics team aren't using this, so we're free to log this however we want.

With that in mind, I agree that it's worth logging the matching and non-matching journey successes separately.  I'm very tempted to say we should do the opposite of what I originally suggested and say "Success Matching Journey" for the new MATCHING_JOURNEY_SUCCESS constant (as it will make the future logs clearer and Anne has confirmed that we don't care about staying compatible with the old logs).  I don't want to keep messing you around though!  You could ask Andy whether he thinks it's worth changing the message again.

app/controllers/confirmation_loa1_controller.rb
```diff
@@ -3,9 +3,20 @@ class ConfirmationLoa1Controller < ApplicationController
   layout 'slides'
 
   def index
```
jakubmiarka: I'd rename the index to something like `matching_journey` and have the other method be something like `non_matching_journey` to make it more readable that there are now two _index/default_ methods

app/controllers/authn_response_controller.rb
andy-paine: Given the perf team aren't using this then yeah we have freedom. In all likelyhood it would end up being us that used this to have a look at relative balances of users in each journey but given this will actually be configured on a per-RP basis we could probably just use Billing & Audit. If we want to record anything then making them actively different as above is fine 👍

app/controllers/confirmation_loa1_controller.rb
```diff
@@ -3,9 +3,20 @@ class ConfirmationLoa1Controller < ApplicationController
...
+    report_to_analytics('Outcome - Matching Not Used By Service')
+  end
+
+private
+
+  def journey_confirmation(matching)
     @idp_name = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider).display_name
     @transaction_name = current_transaction.name
-
+    @matching_status_path = matching ? response_processing_path : redirect_to_service_signing_in_path
```
andy-paine: I think we could name this something more generic like `@redirect_path` - it means the page template doesn't need to know about the concept of matching or not, just where the button should redirect to

app/controllers/confirmation_loa1_controller.rb
andy-paine: Agreed - someone coming to this new would probably get a bit confused and we don't want to continue to promote the idea that matching is the status quo and non-matching is new

app/controllers/confirmation_loa2_controller.rb
```diff
@@ -3,9 +3,20 @@ class ConfirmationLoa2Controller < ApplicationController
...
+    report_to_analytics('Outcome - Matching Not Used By Service')
+  end
+
+private
+
+  def journey_confirmation(matching)
     @idp_name = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider).display_name
     @transaction_name = current_transaction.name
-
+    @matching_status_path = matching ? response_processing_path : redirect_to_service_signing_in_path
```
andy-paine: Same comments from loa1 controller 👍

PR #618, opened 2018-12-03
HUB-376: Cookie Based Journey Hinting - Add New Endpoint

chrisclayson:
https://govukverify.atlassian.net/browse/HUB-376
Add new 'initiate-journey' route and controller

### DIFF
--- a/app/controllers/initiate_journey_controller.rb
+++ b/app/controllers/initiate_journey_controller.rb
@@ -1,0 +1,43 @@
+require 'partials/user_errors_partial_controller'
+
+class InitiateJourneyController < ApplicationController
+  include UserErrorsPartialController
+
+  skip_before_action :validate_session
+  skip_before_action :set_piwik_custom_variables
+
+  def index
+    reset_session
+    simple_id_value = params.fetch('transaction_simple_id', nil)
+    journey_hint_value = params.fetch('journey_hint', nil)
+
+    transaction = CONFIG_PROXY.get_transaction_by_simple_id(simple_id_value)
+    headless_start_page = transaction.nil? ? nil : transaction.fetch('headlessStartpage')
+
+    if !headless_start_page.nil? && valid_journey_hint?(journey_hint_value)
+      session[:journey_hint] = journey_hint_value
+      session[:journey_hint_rp] = simple_id_value
+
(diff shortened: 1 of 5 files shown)


app/controllers/initiate_journey_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    transaction = CONFIG_PROXY.get_transaction_by_simple_id(simple_id)
+    headless_start_page = transaction.nil? ? nil : transaction.fetch('headlessStartpage')
+
+    if !headless_start_page.nil? && valid_journey_hint?(journey_hint_value)
+      session[:journey_hint] = journey_hint_value
+      session[:journey_hint_rp] = simple_id
+
+      return redirect_to merge_query_params(headless_start_page, journey_hint_value)
+    end
+    redirect_to redirect_to_service_error_path
```
jakubmiarka: I'm a little bit unsure about using `redirect_to_service` path in here... it's a different context to this one...

app/controllers/initiate_journey_controller.rb
```diff
@@ -0,0 +1,35 @@
+class InitiateJourneyController < ApplicationController
+  skip_before_action :validate_session
+  skip_before_action :set_piwik_custom_variables
+
+  def index
+    reset_session
+    simple_id = params.dig('transaction_simple_id')
```
jakubmiarka: wouldn't `fetch` achieve the same thing? Dig is more useful in multi-level hashes...

chrisclayson: I used `dig` due to its default behaviour of returning `nil` if index isn't found (rather than throw exception). With `fetch` you have to use the two parameter form to return a default of `nil` ... happy to change it though, I suppose it does look strange on a single dimensional array.

app/controllers/initiate_journey_controller.rb
chrisclayson: You're right, I meant to go back and review the use of this path.

app/controllers/initiate_journey_controller.rb
jakubmiarka: I'm easy, but as you said, it looks strange. That's why it caught my eye. In other places when we deal with params we just use `fetch` with the two arguments. Consistency maybe? Up to you :)

PR #620, opened 2018-12-05
BAU Throw 400 if registering with disconnected IDP

Wynndow:
See this Zendesk ticket: https://govuk.zendesk.com/agent/tickets/3474574

We had some 500s from users trying to register with IDPs that were no
longer offering registrations, via the `/redirect-to-idp-warning` page.
They probably used a bookmark, or a direct link after visiting the sign
in page to set one of the invalid suppliers in selected in their
session.

### DIFF
--- a/app/controllers/redirect_to_idp_warning_controller.rb
+++ b/app/controllers/redirect_to_idp_warning_controller.rb
@@ -2,6 +2,7 @@ require 'partials/idp_selection_partial_controller'
 
 class RedirectToIdpWarningController < ApplicationController
   include IdpSelectionPartialController
+  include ViewableIdpPartialController
 
   SELECTED_IDP_HISTORY_LENGTH = 5
   helper_method :user_has_no_docs_or_foreign_id_only?, :other_ways_description
@@ -9,7 +10,9 @@ class RedirectToIdpWarningController < ApplicationController
   def index
     @idp = decorated_idp
     @service_name = current_transaction.name
-    if @idp.viewable?
+    if !idp_is_providing_registrations?(@idp)
+      something_went_wrong("IDP with entity id: #{@idp.entity_id} is not providing registrations", :bad_request)
+    elsif @idp.viewable?
       render 'redirect_to_idp_warning'
     else
       something_went_wrong("Couldn't display IDP with entity id: #{@idp.entity_id}")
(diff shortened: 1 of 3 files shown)


spec/controllers/redirect_to_idp_warning_controller_spec.rb
```diff
@@ -29,6 +29,25 @@
 
       expect(subject).to render_template('errors/something_went_wrong')
     end
+
+    it 'error page when idp not providing registrations', :chris do
```
jakubmiarka: `:chris`? 😆

Wynndow: Whoops! This is still a bit of WIP.

app/controllers/redirect_to_idp_warning_controller.rb
```diff
@@ -2,14 +2,17 @@
...
 
   SELECTED_IDP_HISTORY_LENGTH = 5
   helper_method :user_has_no_docs_or_foreign_id_only?, :other_ways_description
 
   def index
     @idp = decorated_idp
     @service_name = current_transaction.name
-    if @idp.viewable?
+    if !idp_is_providing_registrations?(@idp)
+      something_went_wrong("IDP with entity id: #{@idp.entity_id} is not providing registrations", :bad_request)
```
jakubmiarka: I wonder if we should be more graceful with the failure... Maybe just delete the selected IDP from the session and redirect them to the `start` page or the `choose-a-certified-company`? Might be trickier for the user to recover from the something went wrong page...

Wynndow: Yep, could do. However I think this is a very thin edge case. If a user is arriving at this page with an incorrect idp in their session, they've done something they probably shouldn't have - like hacking the url. Might be that just returning a 400 is sufficient?

jakubmiarka: or they could just have bookmarked the page 🤷‍♂️ - no hacking skills neccesary 😆 ... I agree it's a thin edge case nevertheless...

Wynndow:
I'm going to close the PR for now. The branch will still be here if it's decided it's the right approach.

PR #624, opened 2018-12-07
ZD-3495915: Update IDP temporarily unavailable message

CharlesIC:
New content reflects IDP being temporarily unavailable

As per the [updated copy](https://docs.google.com/document/d/1-qIcWplfYefRWAEOqyWwzDafd8kUasHI8QB20mNgOkI/edit#)

### DIFF
--- a/app/views/shared/_disconnected_idp_list.erb
+++ b/app/views/shared/_disconnected_idp_list.erb
@@ -2,9 +2,7 @@
   <div class="column-two-thirds">
     <h2 class="heading-medium"><%= t 'hub.certified_companies_disconnected.title' %></h2>
     <p><%=
-      t('hub.certified_companies_disconnected.verify_another_company_html',
-        link: link_to(t('hub.certified_companies_disconnected.verify_another_company_link'), choose_a_certified_company_path)
-      )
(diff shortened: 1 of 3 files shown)


app/views/shared/_disconnected_idp_list.erb
```diff
@@ -2,9 +2,7 @@
   <div class="column-two-thirds">
     <h2 class="heading-medium"><%= t 'hub.certified_companies_disconnected.title' %></h2>
     <p><%=
-      t('hub.certified_companies_disconnected.verify_another_company_html',
-        link: link_to(t('hub.certified_companies_disconnected.verify_another_company_link'), choose_a_certified_company_path)
```
jakubmiarka: are we losing the link to choose a different company on purpose?

CharlesIC: Yes - the users are already on the sign-in picker page.... We don't need to send them anywhere. Also, that link used to point to `choose-a-certified-company` which is the registration journey. It didn't make sense, or I'm missing something.

jakubmiarka: ah, yeah.. well spotted actually.. didn't realize it was a link to the registration...

PR #626, opened 2018-12-10
EID-1005: Fix mobile layout of country picker

simonwo:
- Previously the country scheme logos would wrap
  onto a new line when the layout was narrow.
  This was because they took up a fixed width.
- Mitigate this by not specifying a manual width
  which works for all layouts.

With:
screen shot 2018-12-10 at 15 31 02

Without:
screen shot 2018-12-10 at 15 31 12

### DIFF
--- a/app/assets/stylesheets/pages/_choose-a-country.scss
+++ b/app/assets/stylesheets/pages/_choose-a-country.scss
@@ -24,7 +24,7 @@
   }
 
   .scheme-logo {
-    width: 50%;
+    width: 40%;
     height: 100%;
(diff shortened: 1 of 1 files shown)


CharlesIC:
This change shifts the logos to the left, though. It should look like this:

screen shot 2018-12-11 at 10 38 30

But it looks like this:

screen shot 2018-12-11 at 10 38 24

simonwo:
@CharlesIC updated. It turns out reducing the width is sufficient for all reasonable mobile devices.

PR #628, opened 2018-12-11
HUB-363: Show user session timeout message based on NotOnOrAfter

jakubmiarka:
With HUB-348 we exposed the NotOnOrAfter expiry date of the assertions sent by an IDP. This is to present this information to the user when they're close (<5mins) to the expiry time. And also prevent users to submits Cycle3 attributes with an expired assertion.

It's using amended `GOVUK.modalDialog` component to maintain accessibility of the modal box. The functionality has been changed to satisfy our requirements (triggered by a hard limit rather by inactivity).

It might be easier to review the PR commit by commit.

### DIFF
--- a/app/controllers/further_information_controller.rb
+++ b/app/controllers/further_information_controller.rb
@@ -1,32 +1,32 @@
 class FurtherInformationController < ApplicationController
   def index
+    return redirect_to further_information_timeout_path if !session[:assertion_expiry].nil? && expired?
+
     session_id = session[:verify_session_id]
     @cycle_three_attribute = FURTHER_INFORMATION_SERVICE.get_attribute_for_session(session_id).new({})
     @transaction_name = current_transaction.name
+    @idp_name = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider).display_name
   end
 
   def submit
-    if !session[:assertion_expiry].nil? && expired?
-      redirect_to further_information_timeout_path
+    return redirect_to further_information_timeout_path if !session[:assertion_expiry].nil? && expired?
+
+    session_id = session[:verify_session_id]
+    cycle_three_attribute_class = FURTHER_INFORMATION_SERVICE.get_attribute_for_session(session_id)
+    @cycle_three_attribute = cycle_three_attribute_class.new(params['cycle_three_attribute'])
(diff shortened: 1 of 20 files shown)


app/controllers/further_information_controller.rb
```diff
@@ -37,4 +47,10 @@ def submit_null_attribute
       something_went_wrong('Unexpected submission to Cycle3 Null Attribute endpoint', :forbidden)
     end
   end
+
+private
+
+  def expired?
+    Time.parse(session[:assertion_expiry]) < Time.now
```
chrisclayson: Technically shouldn't this be `<=` ... not going to make huge difference!

jakubmiarka: Technically yes

app/views/further_information/index.html.erb
```diff
@@ -1,6 +1,10 @@
 <%= page_title 'hub.further_information.title', cycle_three_name: @cycle_three_attribute.field_name %>
 <% content_for :feedback_source, 'CYCLE_3_PAGE' %>
 
+<% if !session[:assertion_expiry].nil? %>
```
chrisclayson: `unless` instead of `if !`?

jakubmiarka: Hm, codacy has a new name :D

config/routes.rb
```diff
@@ -11,7 +11,7 @@ def add_routes(routes_name)
     instance_eval(File.read(Rails.root.join("config/#{routes_name}.rb")))
   end
 
-  get 'initiate-journey/:transaction_simple_id', to: 'initiate_journey#index'
+  get 'initiate-journey/:transaction_simple_id', to: 'initiate_journey#index', as: 'initiate_journey'
```
chrisclayson: Isn't convention here to use symbols for the `as` rather than literals?

app/controllers/further_information_controller.rb
jakubmiarka: fixed

config/routes.rb
jakubmiarka: it indeed is. fixed

app/views/further_information/index.html.erb
jakubmiarka: fixed

PR #636, opened 2018-12-14
BAU: Fix Invalid Journey Hint Handling

chrisclayson:
Change the joruney hint handler so that it still routes to RP, but discards hint, rather than throwing error.

### DIFF
--- a/spec/controllers/initiate_journey_controller_spec.rb
+++ b/spec/controllers/initiate_journey_controller_spec.rb
@@ -32,11 +32,11 @@ describe InitiateJourneyController do
   end
 
   context 'is called with invalid parameters' do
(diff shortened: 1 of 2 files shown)


spec/controllers/initiate_journey_controller_spec.rb
```diff
@@ -33,10 +33,10 @@
 
   context 'is called with invalid parameters' do
     it 'should display error page if invalid journey hint value passed' do
```
jakubmiarka: the name of the test is no longer accurate

chrisclayson: Well spotted 👍

PR #638, opened 2018-12-17
EID-1170: Work around for oddly shaped focus selectors

tomrosier:
The issue is caused by the css element line 46 `govuk_frontend_toolkit-8.1.0/app/assets/stylesheets/_shims.scss` where it creates an `:after` element which has the content of `""`, if the content is set to `none` rather than `""` the focus selection box is the correct shape.


jakubmiarka:
Have we identified the underlying issue? 

The library is longer maintained, there's a tech debt card to migrate to GOV.UK Frontend: https://trello.com/c/qBjNuAeH/98-migrate-frontend-to-use-govuk-frontend-instead-of-elements

tomrosier:
Updated the description with a brief description of what I think the underlying problem is.

PR #639, opened 2018-12-17
EID-1178: Update rails

vixus0:
Probably a good time to update rails.


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 ruby '2.4.2'
 
-gem 'rails', '5.0.2'
+gem 'rails', '5.0.7.1'
```
willp-bl: any reason why this can't go to the latest version of rails?

config/environments/production.rb
```diff
@@ -61,7 +61,7 @@
 
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
-  config.i18n.fallbacks = true
+  config.i18n.fallbacks = [I18n.default_locale]
```
willp-bl: is it worth making this `[:en]` as done in the initializer below ⬇️

Gemfile
vixus0: Think I tried going to 5.2.2 before and there were some breaking feature changes. Don't really have time to fix them so this will have to do for now.

config/environments/production.rb
vixus0: Could do, I did this following the instructions given by the message following the `i18n` gem upgrade. Who knows, we might want to make `:cy` the default someday :trollface:

Gemfile
jakubmiarka: ➡️ tech debt trello?

tomrosier: + 👍

PR #644, opened 2018-12-24
HUB-405: Recommendation Model v1 A/B test

jakubmiarka:
Variant version will present recommended IDPS more prominently and not
recommended ones a bit less to test user interaction
if the recommendation model would be deployed.

Config [here](https://github.com/alphagov/verify-frontend-federation-config/pull/108)

### DIFF
--- a/app/controllers/choose_a_certified_company_loa2_variant_controller.rb
+++ b/app/controllers/choose_a_certified_company_loa2_variant_controller.rb
@@ -10,6 +10,12 @@ class ChooseACertifiedCompanyLoa2VariantController < ApplicationController
     @non_recommended_idps = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(suggestions[:unlikely])
     session[:user_segments] = suggestions[:user_segments]
     FEDERATION_REPORTER.report_number_of_idps_recommended(current_transaction, request, @recommended_idps.length)
+    # Hack this in here to make this AB test work ...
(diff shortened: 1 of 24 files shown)


config/locales/en.yml
```diff
@@ -327,6 +327,12 @@ en:
         title: About %{display_name}
         information_provided_by: Information provided by %{display_name}
         close: close
+      ab_test:
+        intro: 'Based on your answers, these companies are most likely to work for you:'
```
jakubmiarka: this still requires confirmation from Claire

jakubmiarka: confirmed

app/controllers/choose_a_certified_company_loa2_variant_controller.rb
```diff
@@ -10,6 +10,12 @@ def index
     @non_recommended_idps = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(suggestions[:unlikely])
     session[:user_segments] = suggestions[:user_segments]
     FEDERATION_REPORTER.report_number_of_idps_recommended(current_transaction, request, @recommended_idps.length)
+    # Hack this in here to make this AB test work ...
+    top_three = { "post-office-stub" => "post-office-stub", "digidentity" => "digidentity", "experian" => "experian" }
```
jakubmiarka: this is an open source repo and we should not name real IDPs!

jakubmiarka:
Superseeded by https://github.com/alphagov/verify-frontend/pull/710

PR #648, opened 2019-01-10
BAU: Do not show onboarding services on the SIDP endpoint

jakubmiarka:
To allow testing against test services we can enable them in the config. However,
then they will appear in the get-available-services endpoint along with the test ones.
This change filters out the test ones using the name_and_homepage list. In production this is the list: https://github.com/alphagov/verify-frontend-federation-config/blob/ece7721b9893d86be8964becd442adb0d3342ffa/configuration/relying_parties.yml#L25

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -20,7 +20,9 @@ class MetadataController < ApplicationController
   end
 
   def service_list
-    transactions = transactions_for_service_list
+    transaction_list = transactions_for_service_list
+    displayed_transactions = RPS_NAME_AND_HOMEPAGE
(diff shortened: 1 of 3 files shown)


jakubmiarka:
Kind of... It's still using the SIDP list, but basically what I'm trying to achieve is to have a service enabled on SIDP list, but not visible when the list is used on the endpoint. When using to validate it should still be there and valid.

chrisclayson:
Potential better (but more complex) solution is to have new flag in config to show/hide from SIDP list! (whichever way round you want the default)?

jakubmiarka:
Not required for now - so closing the PR.

Btw, I still prefer this option. Using the same list for 'display' config rather than having another flag which would be incosistent with the logic how we control displayed items on the `/verify-services` page.

PR #649, opened 2019-01-21
Add copy for IDP disconnect

bjgill:
Royal Mail and Citizensafe come to the end of their continued assurance period on March 1st. Therefore, we need to have the copy in place to explain to users why they can no longer use those IDPs to log in any more.

I've not done any testing of this PR, and we need to get the Welsh translation before merging.

See https://trello.com/c/vi7RV11M

### DIFF
--- a/app/views/shared/_disconnected_idp_list.erb
+++ b/app/views/shared/_disconnected_idp_list.erb
@@ -3,7 +3,7 @@
     <h2 class="heading-medium"><%= t 'hub.certified_companies_disconnected.title' %></h2>
     <p><%=
       t('hub.certified_companies_disconnected.verify_another_company_html',
-        link: link_to(t('hub.certified_companies_disconnected.verify_another_company_link'), choose_a_certified_company_path)
+        link: link_to(t('hub.certified_companies_disconnected.verify_another_company_link'), begin_registration_path, id: 'begin-registration-route')
       )
     %></p>
   </div>
--- a/app/views/shared/_disconnected_idp_list.erb
+++ b/app/views/shared/_disconnected_idp_list.erb
@@ -2,7 +2,9 @@
   <div class="column-two-thirds">
     <h2 class="heading-medium"><%= t 'hub.certified_companies_disconnected.title' %></h2>
     <p><%=
-      t 'hub.certified_companies_disconnected.verify_another_company_html'
+      t('hub.certified_companies_disconnected.verify_another_company_html',
+        link: link_to(t('hub.certified_companies_disconnected.verify_another_company_link'), choose_a_certified_company_path)
+      )
     %></p>
   </div>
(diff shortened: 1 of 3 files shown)


bjgill:
@RacingHippo - can you confirm that this is as you expected? If so, I'll send this off for translation.

config/locales/en.yml
```diff
@@ -499,8 +499,9 @@ en:
       verify_another_company_html: To access government services, you must %{link}.
       verify_another_company_link: verify your identity with another certified company
     certified_companies_disconnected:
-      title: These companies are temporarily unavailable with GOV.UK Verify
-      verify_another_company_html: You can try again later or verify your identity with another certified company.
+      title: The following companies are no longer providing identity services as part of GOV.UK Verify.
```
bjgill: Probably remove the full stop.

bjgill:
There is now some question as to whether the copy is correct in the first place. Ruth is going to check about it in the comms meeting this morning (?).

In the interim, I've tested this manually. The "The following companies are no longer providing identity services as part of GOV.UK Verify." section appearerd on the login page as expected.

However, I noticed that the `choose_a_certified_company_path` link links straight to http://localhost:50300/choose-a-certified-company. In other words, it appears to bypass all the eligibility questions and dump you in a page with a 'choice' of only Experian.

Is there another link we should be sending them to instead? Maybe http://localhost:50300/about, or http://localhost:50300/about-choosing-a-company perhaps?

We'll probably need to answer this after Ruth has returned with an answer to her question - assuming we don't need to throw the whole thing out and start again...

app/views/shared/_disconnected_idp_list.erb
```diff
@@ -2,7 +2,9 @@
   <div class="column-two-thirds">
     <h2 class="heading-medium"><%= t 'hub.certified_companies_disconnected.title' %></h2>
     <p><%=
-      t 'hub.certified_companies_disconnected.verify_another_company_html'
+      t('hub.certified_companies_disconnected.verify_another_company_html',
+        link: link_to(t('hub.certified_companies_disconnected.verify_another_company_link'), choose_a_certified_company_path)
```
jakubmiarka: I would suggest linking this to `begin_registration_path` ... It makes sure the user gets reported as a switching to the the registration journey and to the eligibility question when required.

bjgill: Great - thanks. I've updated the link.

bjgill:
We've got new copy:

>The following companies are no longer part of GOV.UK Verify
> If you had an identity account with one of these companies, you'll need to verify your identity with a different  company.

bjgill:
Oops - accidentally pushed a commit to the wrong branch. Have now backed out - this should be ready to go.

PR #652, opened 2019-01-25
ZD-3576498 avoid logging an error for a Single IDP journey interruption.

galund:
(Plan to do another PR with the piwik change, but I can't get it to work yet :/ )

 - these events are often (tho not always) a legitimate change of
   journey for the user, so we shouldn't log them as errors.

 - the information may still be useful in diagnosing cases where the
   user did not intend to change from one transaction to another,
   which is also a separate issue:
   https://govuk.zendesk.com/agent/tickets/3559335

https://trello.com/c/Xymb17Ua/39-single-idp-journey-interruptions-should-not-log-an-error
https://govuk.zendesk.com/agent/tickets/3576498

### DIFF
--- a/app/controllers/single_idp_journey_controller.rb
+++ b/app/controllers/single_idp_journey_controller.rb
@@ -138,7 +138,9 @@ private
     return false if cookie_value_is_missing(%w(idp_entity_id transaction_id uuid))
 
     unless cookie_matches_session?(transaction_id)
-      logger.error "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]} for transaction_id #{transaction_id}" + referrer_string
+      # TODO: we think this should be noted in Piwik; probably does't need to be in the error log at all
+      logger.info "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]}"\
+                      " for transaction_id #{transaction_id}" + referrer_string
       return false
     end
 
--- a/app/controllers/single_idp_journey_controller.rb
+++ b/app/controllers/single_idp_journey_controller.rb
@@ -140,7 +140,7 @@ private
     unless cookie_matches_session?(transaction_id)
       # TODO: we think this should be noted in Piwik; probably does't need to be in the error log at all
       logger.info "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]}"\
-                      " for transaction_id #{transaction_id}" + referrer_string
+                      " for transaction_id #{transaction_id} with uuid #{uuid}"
       return false
     end
 
--- a/spec/controllers/single_idp_journey_controller_spec.rb
+++ b/spec/controllers/single_idp_journey_controller_spec.rb
@@ -214,7 +214,7 @@ describe SingleIdpJourneyController do
         uuid: UUID_ONE
       }
       cookies.encrypted[CookieNames::VERIFY_SINGLE_IDP_JOURNEY] = single_idp_cookie.to_json
-      expect(Rails.logger).to receive(:error).with(/The value of the Single IDP cookie does not match the session value of http:\/\/www.test-rp.gov.uk\/SAML2\/MD for transaction_id disabled-rp/)
+      expect(Rails.logger).to receive(:info).with(/The value of the Single IDP cookie does not match the session value of http:\/\/www.test-rp.gov.uk\/SAML2\/MD for transaction_id disabled-rp/)
       expect(subject).to redirect_to(start_path)
     end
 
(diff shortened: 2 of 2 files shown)


app/controllers/single_idp_journey_controller.rb
```diff
@@ -138,7 +138,8 @@ def valid_selection?
     return false if cookie_value_is_missing(%w(idp_entity_id transaction_id uuid))
 
     unless cookie_matches_session?(transaction_id)
-      logger.error "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]} for transaction_id #{transaction_id}" + referrer_string
+      # TODO we think this should be noted in Piwik; probably does't need to be in the error log at all
+      logger.info "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]} for transaction_id #{transaction_id}" + referrer_string
```
jakubmiarka: I would use `warn` level

jakubmiarka:
It would be good to get some numbers. How often this happens and do some analysis what the timeframe between initiating the SIDP and coming from a difference service actually is. This is not a real fix. The real fix could be decreasing the lifetime of the SIDP cookie.

app/controllers/single_idp_journey_controller.rb
galund: I truly don't think this is even a warning – this turns out to be expected behaviour. It's *really* frequent. 10 an hour at least – these Piwik logs are kinda helpful to see what's going on: https://docs.google.com/document/d/1dChpTrvVeWUs8-5DhY8SsSENX0No6tIqRtr8c3U3sl0/edit#heading=h.ohlmf2ehft3z

galund:
These journey "interruptions" are happening sometimes in the middle of another journey.

e.g. RP1 -> Verify -> Post Office -> Single IDP marketing -> new RP (SIDP) -> Verify -> Post Office -> marketing -> RP1 (interruption to SIDP journey)

You can adjust the cookie to the lowest possible amount of time for it to still actually work, and people will still decide actually they want to go back to the place they were originally trying to get to, not to some other service they got distracted into looking at!

app/controllers/single_idp_journey_controller.rb
jakubmiarka: The piwik logs don't show any of the scenarios which would trigger this. They're the opposite - starting a non-SIDP journey and switching to a SIDP, which is fine and will not trigger anything. The problem is when users start a SIDP first and then go on to a non-SIDP journey.

jakubmiarka:
No, you can't reduce the time of the cookie to minimum, because you have to give them some time to read the GOV.UK start page information etc... 

I just don't like "fixing" stuff by hiding the logs. Another solution could be to reduce the cookie lifetime after they went to the IDP (so they can still click back and continue on SIDP) but after a certain time (i.e. 1 minute) their cookie is gone and when they come back with a different RP they will be normal users.

galund:
I don't think the fix can have anything to do with cookie lifetime. The single IDP cookie *will* sometimes be present when a user decides to go to a different service?

One question: if I already have a single IDP cookie, and I click on *another* single IDP link – do I expect to have my cookie updated? If not then definitely it's certainly easy for the user to end up with a conflicted SIDP cookie compared to the service they go to in the end.

Happy to talk through next week.

Don't particularly want to land+release this now if it's controversial: better to just mute things on Sentry until we're decided, maybe.

jakubmiarka:
Yes, if you have existing SIDP cookie and click on another SIDP prompt (even with the same IDP and different RP), the cookie will get updated.

I think it has everything to do with the lifetime of the cookie. The cookie is hanging around for too long so when we see the user coming back to hub again they still have the cookie but they might be on a completely different journey (e.g. it can be an hour later).

app/controllers/single_idp_journey_controller.rb
```diff
@@ -138,7 +138,9 @@ def valid_selection?
     return false if cookie_value_is_missing(%w(idp_entity_id transaction_id uuid))
 
     unless cookie_matches_session?(transaction_id)
-      logger.error "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]} for transaction_id #{transaction_id}" + referrer_string
+      # TODO: we think this should be noted in Piwik; probably does't need to be in the error log at all
+      logger.info "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]}"\
+                      " for transaction_id #{transaction_id}" + referrer_string
```
jakubmiarka: referrer will be the headless RP page so it will match the value in the `session[:transaction_entity_id]`... I would suggest adding the UUID here instead, so we can link the initialization step.

phantas: done

PR #653, opened 2019-01-28
HUB-7 Add Piwik event for unexpected single IDP journey

bjgill:
Add a new event type, and fire it when a user's cookie value isn't as expected. Make it easier to see when users are encountering this rather odd journey from the Piwik data.

Co-authored-by: John Watts <john.watts@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/single_idp_journey_controller.rb
+++ b/app/controllers/single_idp_journey_controller.rb
@@ -138,7 +138,8 @@ private
     return false if cookie_value_is_missing(%w(idp_entity_id transaction_id uuid))
 
     unless cookie_matches_session?(transaction_id)
-      # TODO: we think this should be noted in Piwik; probably does't need to be in the error log at all
+      FEDERATION_REPORTER.report_unexpected_single_idp_journey(current_transaction, request, session[:transaction_entity_id], uuid)
+      # TODO: we think this should be noted in Piwik; probably doesn't need to be in the error log at all
       logger.info "The value of the Single IDP cookie does not match the session value of #{session[:transaction_entity_id]}"\
                       " for transaction_id #{transaction_id} with uuid #{uuid}"
       return false
--- a/app/models/analytics/federation_reporter.rb
+++ b/app/models/analytics/federation_reporter.rb
@@ -41,6 +41,16 @@ module Analytics
       )
     end
 
+    def report_unexpected_single_idp_journey(current_transaction, request, session_value, uuid)
+      report_action(
+        current_transaction,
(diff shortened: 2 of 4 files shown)


lib/analytics/custom_variable.rb
```diff
@@ -7,6 +7,8 @@ module CustomVariable
       cycle_three_attribute: { name: 'CYCLE_3', index: 4 },
       idp_selection: { name: 'IDP_SELECTION', index: 5 },
       ab_test: { name: 'AB_TEST', index: 6 },
+      session_value: { name: 'SESSION_VALUE', index: 7 },
```
jakubmiarka: this is not going to work. 
```
root@analytics-1:/usr/share/matomo# ./console customvariables:info

Your Piwik is configured for 6 custom variables.
```

app/models/analytics/federation_reporter.rb
```diff
@@ -41,6 +41,16 @@ def report_started_single_idp_journey(request)
       )
     end
 
+    def report_unexpected_single_idp_journey(current_transaction, request, session_value, uuid)
+      report_action(
```
jakubmiarka: you probably want to report an event rather than a pageview (action)

app/controllers/single_idp_journey_controller.rb
```diff
@@ -138,7 +138,8 @@ def valid_selection?
     return false if cookie_value_is_missing(%w(idp_entity_id transaction_id uuid))
 
     unless cookie_matches_session?(transaction_id)
-      # TODO: we think this should be noted in Piwik; probably does't need to be in the error log at all
+      FEDERATION_REPORTER.report_unexpected_single_idp_journey(current_transaction, request, session[:transaction_entity_id], uuid)
+      # TODO: we think this should be noted in Piwik; probably doesn't need to be in the error log at all
```
jakubmiarka: that TODO can be removed now as you're doing that TODO :)

lib/analytics/custom_variable.rb
JohnWattsGDS: Am I right in thinking that we can't log this in piwik at all then?

jakubmiarka: As discussed, you can, just don't use custom variables for this kind of stuff. These custom variables apply to visit scope (so tag the whole visit, not just an action). Better to just use events. Custom dimensions are there too, but this is a great candidate for an [event](https://matomo.org/docs/event-tracking/).

PR #660, opened 2019-02-05
HUB-42: Communicate to users when IDP is having technical difficulty

bjgill:
When an IDP is temporarily down, show users a warning. Explain that the IDP is temporarily down.

Use config provided by the hub (https://github.com/alphagov/verify-hub/pull/223) to decide which IDPs are affected.

See https://trello.com/c/ly1m24h4/


app/controllers/partials/viewable_idp_partial_controller.rb
```diff
@@ -36,7 +36,11 @@ def current_identity_providers_for_loa
   end
 
   def current_identity_providers_for_sign_in
-    CONFIG_PROXY.get_idp_list_for_sign_in(session[:transaction_entity_id]).idps.select(&:authentication_enabled)
+    CONFIG_PROXY.get_idp_list_for_sign_in(session[:transaction_entity_id]).idps.select(&:authentication_enabled).reject(&:temporarily_unavailable)
```
bjgill: `\n`?

app/controllers/partials/viewable_idp_partial_controller.rb
```diff
@@ -36,7 +36,11 @@ def current_identity_providers_for_loa
...
+  end
+
+  def current_temporarily_unavailable_identity_providers_for_sign_in
+    CONFIG_PROXY.get_idp_list_for_sign_in(session[:transaction_entity_id]).idps.select(&:temporarily_unavailable)
```
bjgill: Repetition of `CONFIG_PROXY.get_idp_list_for_sign_in(session[:transaction_entity_id]).idps`

app/models/display/viewable_identity_provider.rb
```diff
@@ -11,6 +11,7 @@ module Display
     delegate :to_key, to: :identity_provider
     delegate :authentication_enabled, to: :identity_provider
     delegate :display_name, :about_content, :requirements, :special_no_docs_instructions, :no_docs_requirement, :contact_details, :interstitial_question, :interstitial_explanation, :mobile_app_installation, :tagline, to: :display_data
+    delegate :temporarily_unavailable, to: :identity_provider
```
bjgill: Refactor!!!!!!!!!!!!!!!!!!!

app/models/identity_provider.rb
```diff
@@ -1,22 +1,24 @@
...
-  attr_reader :simple_id, :entity_id, :levels_of_assurance, :authentication_enabled
+  attr_reader :simple_id, :entity_id, :levels_of_assurance, :authentication_enabled, :temporarily_unavailable
   validates_presence_of :simple_id, :entity_id, :levels_of_assurance
 
   def initialize(hash)
     @simple_id = hash['simpleId']
     @entity_id = hash['entityId']
     @levels_of_assurance = hash['levelsOfAssurance']
     @authentication_enabled = hash.has_key?('authenticationEnabled') ? hash['authenticationEnabled'] : true
+    @temporarily_unavailable = hash.has_key?('temporarilyUnavailable') ? hash['temporarilyUnavailable'] : false
```
bjgill: Should be able to replace this with the `Hash.fetch` method

app/models/identity_provider.rb
```diff
@@ -1,22 +1,24 @@
...
   end
 
   def ==(other)
     if other.is_a?(IdentityProvider)
       self.simple_id == other.simple_id &&
         self.entity_id == other.entity_id &&
         self.levels_of_assurance == other.levels_of_assurance &&
-        self.authentication_enabled == other.authentication_enabled
+        self.authentication_enabled == other.authentication_enabled &&
+        self.temporarily_unavailable == other.temporarily_unavailable
```
bjgill: Can we reduce the levels of repetition here? Can we use https://stackoverflow.com/questions/10006889/how-to-get-attributes-that-were-defined-through-attr-reader-or-attr-accessor, perhaps?

bjgill:
OK. This should now be ready for review and production. We'll open separate PRs for all the other work required for this story.

app/helpers/application_helper.rb
```diff
@@ -65,4 +65,12 @@ def button_link_to text, path, options = {}
     options[:role] = 'button'
     link_to text, path, options
   end
+
+  def unavailable_certified_company_title(unavailable_companies)
+    if unavailable_companies.length == 1
+      'hub.certified_companies_unavailable.singular_title'
+    else
+      'hub.certified_companies_unavailable.plural_title'
```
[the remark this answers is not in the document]
bjgill: I agree that this is a bit clunky. This was just something we came up with - I've not discovered anything else trying to do a similar thing. For example, disconnection always uses the plural.

Did you have an alternative? Might be worth putting into one of our follow-on PRs...

bjgill:
@jakubmiarka - this PR seems good to go. As someone more familiar with this code than anyone in optimise, would you mind giving this a review? Ideally by tomorrow morning? Thanks.

app/helpers/application_helper.rb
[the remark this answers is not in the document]
galund: It should be do-able with the right template invocation to pass in the `count` to the i18n back-end, putting the various options into sub-keys in the translation database: https://guides.rubyonrails.org/i18n.html#pluralization

... searches around...

aha here's an example: `choose_a_certified_company/choose_a_certified_company_LOA2.html.erb`

config/locales/cy.yml
```diff
@@ -499,6 +499,12 @@ cy:
       session_not_valid: Mae eich sesiwn wedi’i amseru allan.
       session_not_valid_link: ddechrau eto
       product_page: Dychwelyd i dudalen GOV.UK  Verify
+    certified_companies_unavailable:
```
tomrosier: Has a ticket been raised for theses translations? And have theses been added to the spreadsheet for translation?

app/controllers/partials/viewable_idp_partial_controller.rb
```diff
@@ -35,12 +35,20 @@ def current_identity_providers_for_loa
     CONFIG_PROXY.get_idp_list_for_loa(session[:transaction_entity_id], session[:requested_loa]).idps
   end
 
+  def current_all_identity_providers_for_sign_in
+    CONFIG_PROXY.get_idp_list_for_sign_in(session[:transaction_entity_id]).idps
+  end
+
   def current_identity_providers_for_sign_in
```
phantas: Please rename this function to make it clear why it's different from `current_all_identity_providers_for_sign_in`, in line with the other functions below. maybe `current_available...` ?

app/views/choose_a_certified_company/_idp_option.html.erb
```diff
@@ -9,16 +9,23 @@
...
+                      class: recommended ? 'button' : 'button-link',
+                      role: recommended ? nil : 'link',
+                      name: identity_provider.simple_id,
+                      id: nil,
+                      type: 'submit',
+                      value: identity_provider.display_name,
+                      'data-order': order
+        %>
+      </div>
+  <% else %>
```
phantas: Forgive my ruby n00bness, but does this mean a button is not displayed when the company in temporarily unavailable?

config/locales/cy.yml
phantas: bit nitpicking but please add `temporary unavailable` instead of `unavailable` so not to confuse with the single idp one, for example

config/locales/en.yml
```diff
@@ -493,6 +493,12 @@ en:
       session_not_valid: Your session has timed out.
       session_not_valid_link: start again
       product_page: Return to the GOV.UK Verify product page
+    certified_companies_unavailable:
```
phantas: likewise

spec/features/user_visits_choose_a_certified_company_page_spec.rb
```diff
@@ -20,6 +20,12 @@
       )
     end
 
+    it 'marks the unavailable IDP as unavailable' do
```
phantas: again, bit nitpicking: `temporarily unavailable` instead of `unavailable`

spec/features/user_visits_choose_a_certified_company_page_spec.rb
```diff
@@ -92,6 +100,12 @@
         expect(page).to have_button('Choose LOA1 Corp')
       end
     end
+
```
phantas: again, nitpicking: `temporarily unavailable`...?

app/helpers/application_helper.rb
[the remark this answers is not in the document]
bjgill: Ooh. What a nice rails feature. We'll switch to using that, then.

app/views/choose_a_certified_company/_idp_option.html.erb
bjgill: Indeed. When a company is unavailable, we do not show the button. Thus, users are unable to navigate to sign-in/register to an IDP we've marked as unavailable.

config/locales/cy.yml
bjgill: We've got a follow-up PR where we remove the old 'unavailable' feature, and then rename 'temporarily unavailable' to 'unavailable'. Given that, would it be acceptable to leave this (and the other examples below) as-is?

bjgill: Good point - I thought there was a ticket, but there doesn't seem to be.

- [ ] Sort out translation ticket

app/views/choose_a_certified_company/_idp_option.html.erb
phantas: Is this correct? What I was told is that the agreed behaviour is that the user can still sign-in/register to an IDP

config/locales/cy.yml
phantas: sure. But what do you mean by remove the old unavailable feature?!

bjgill: The commit where we propose to remove that feature is: https://github.com/alphagov/verify-frontend/commit/6b34b9ec6d7559d24820322994f96176d299aa4d

It's somewhat similar to the function we introduce in this PR. However, the config is in the frontend-fed-config, and the user journeys are somewhat different.

app/views/sign_in/index.html.erb
```diff
@@ -32,6 +32,12 @@
   <%= render partial: 'shared/continue_to_idp_form' %>
 </div>
 
+<% if !@temporarily_unavailable_identity_providers.empty? %>
```
bjgill: `.any?` instead of `!foo.empty?`?

app/controllers/paused_registration_controller.rb
```diff
@@ -80,9 +80,9 @@ def get_idp_from_cookie
     from_resume_link_idp_value = resume_link_idp
     if from_resume_link_idp_value.nil?
       last_idp_value = last_idp
-      last_idp_value.nil? ? nil : retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_sign_in, last_idp_value).first
+      last_idp_value.nil? ? nil : retrieve_decorated_singleton_idp_array_by_entity_id(current_available_identity_providers_for_sign_in, last_idp_value).first
     else
-      retrieve_decorated_singleton_idp_array_by_simple_id(current_identity_providers_for_sign_in, from_resume_link_idp_value).first
+      retrieve_decorated_singleton_idp_array_by_simple_id(current_available_identity_providers_for_sign_in, from_resume_link_idp_value).first
```
jakubmiarka: I'm a little bit hesitant on this... not allowing an IDP on the resume page will redirect users to the normal path and thus invalidating the resume functionality, not sure that's desirable?

app/views/choose_a_certified_company/_idp_option.html.erb
jakubmiarka: It was my understanding too. I thought we were only going to "warn" users that the IDP might be having issues rather then fully prevent them visiting the IDP. Otherwise we would have to be monitoring the IDP and release a config change whenever they come back.

bjgill:
OK. After a lot of fafing around, I think we're now ready to pick this back up. We're not going to change any of the process. Instead, we're just going to finish delivering this. It's not perfect, but it is a significant improvement to the user experience. We'll also be able to remove a bunch of tech debt (i.e. the config in verify-frontend-fed-config)

app/views/choose_a_certified_company/_idp_option.html.erb
bjgill: As mentioned below, we just want to get this shipped with minimal additional work. The design as implemented is no worse than what currently exists, so we're not going to make changes.

bjgill:
So the only thing left is putting in the Welsh translation that we've since had back, and I think this can be merged!

app/views/choose_a_certified_company/_idp_option.html.erb
jakubmiarka: I would argue it's making it worse in case when IDP is back up but we are not letting people to that IDP to access their service 🤔

bjgill: The problem is that there is currently no agreed process for the usage of this feature. I agree that the lack of a link to the IDP may make things worse. However, I don't think we're making anything worse. We don't _need_ to use this new feature.

I agree that there's a decision here as to what the process for using this feature will/should be. The outcome of that decision will determine the answer to questions like these. At the moment, there is no process, and this is completely undocumented.

We're therefore going to ship the cheapest possible thing (this PR with no changes) and leave the follow-on decisions to Nicole, who is better placed to answer.

jakubmiarka: To me, it feels like we're creating a tech debt - shipping _unusable_ feature with no documentation on how/when to use it. 

Anyone without this context will not know the caveats and background.

app/views/shared/_temporarily_unavailable_idp_list.erb
```diff
@@ -0,0 +1,25 @@
...
+  <div class="column-one-third company-wrapper">
+    <div class="company">
+      <div class="company-inner">
+        <div class="company-logo">
+          <div class="company-logo-inner">
+            <%= image_tag(identity_provider.logo_path, alt: idp_tagline(identity_provider)) %>
+          </div>
+        </div>
+      </div>
+      <p class="subtitle"><%= identity_provider.display_name %> is unavailable</p>
```
bjgill: This probably needs to be in a translation... 😞

bjgill:
Right. I think this should be ready to 🚢 . Re-reviews, please.

app/controllers/certified_company_unavailable_controller.rb
```diff
@@ -6,7 +6,7 @@ class CertifiedCompanyUnavailableController < ApplicationController
   def index
     simple_id = params[:company]
 
-    if UNAVAILABLE_IDPS.include?(simple_id) && !current_identity_providers_for_sign_in.map(&:simple_id).include?(simple_id)
+    if UNAVAILABLE_IDPS.include?(simple_id) && !current_available_identity_providers_for_sign_in.map(&:simple_id).include?(simple_id)
```
galund: Ignoring the weirdness of this on the ground that it's going away shortly.

bjgill:
Thanks for the reviews! I'll merge now. I'll sort out Codacy's complaints in my follow-on PR, alongside removing the old 'unavailable' feature.

PR #661, opened 2019-02-06
ly1m24h4: remove superseded 'unavailable' feature

bjgill:
Follow-on from https://github.com/alphagov/verify-frontend/pull/660

We have now introduced a new implementation of 'unavailable'. This new implementation uses `verify-hub-federation-config` and covers all user journeys. The old equivalent (improperly) uses `verify-frontend-federation-config` and only covers sign-in. This PR therefore removes the old implementation, which unblocks some tidying up of the new one.

Also add correct handling of single IDP. Now, if a user comes in on the single IDP journey for an unavailable IDP, they will be treated as if they didn't have a valid cookie. That is, they get sent to the signin page where they will see that their IDP is unavailable and have the option to choose another.

See https://trello.com/c/ly1m24h4/42-communicate-to-users-when-idp-is-having-technical-difficulty


JohnWattsGDS:
Is this PR still required?

bjgill:
Yes - it's just blocked on #660

JohnWattsGDS:
There's a conflict at the moment so I won't review anything until it's resolved.

spec/controllers/sign_in_controller_spec.rb
```diff
@@ -7,13 +7,13 @@
   before(:each) do
     stub_api_idp_list_for_sign_in([{ 'simpleId' => 'stub-idp-one',
                                      'entityId' => 'http://idcorp.com',
-                                     'levelsOfAssurance' => %w(LEVEL_1) },
+                                     'levelsOfAssurance' => %w[LEVEL_1] },
```
jakubmiarka: why are we changing the convention of using round brackets to square ones? the rest of the codebase is using `%w(...`

bjgill: Rubocop was complaining - https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/PercentLiteralDelimiters.

I suspect the right solution is actually to undo that change and instead update the rubocop config to prefer `%w(...)`

jakubmiarka: Ah, that rule had to change 🤦‍♂️ ...  I remember changing all the `[]` to `%w()` because linter said so here https://github.com/alphagov/verify-frontend/commit/f581aacb004da1faf202fdb2c47ab4002e8a05ff ... there are 183 occurences of `%w(...)`

PR #663, opened 2019-02-18
BAU: Tidy up single IDP controller a little

bjgill:
Attempt to use slightly more idiomatic Ruby patterns.


bjgill:
I agree that it's non-obvious. However, I think the ergonomic improvement justifies the cost.

PR #665, opened 2019-02-25
ZD3579186: Add error log for missing entity ID

JohnWattsGDS:
Transaction entity ID is required for the single IDP journey, and should always be available. Add log to catch any false positives of this new validation. Prepare for replacing log with validation error if there are no false positives. Commented out code will be re-instated by follow-on PR.

Co-authored-by: Benjamin Gill <benjamin.gill@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/models/session_validator.rb
+++ b/app/models/session_validator.rb
@@ -7,7 +7,8 @@ class SessionValidator
       TransactionSimpleIdPresence.new,
       SessionStartTimeValidator.new(session_duration),
       CookieSizeValidator.new,
-      RequestedLOAValidator.new
+      RequestedLOAValidator.new,
+      TransactionEntityIdPresence.new
     ]
   end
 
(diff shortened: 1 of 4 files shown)


app/models/session_validator.rb
```diff
@@ -7,7 +7,8 @@ def initialize(session_duration)
       TransactionSimpleIdPresence.new,
       SessionStartTimeValidator.new(session_duration),
       CookieSizeValidator.new,
-      RequestedLOAValidator.new
+      RequestedLOAValidator.new,
+      TransactionEntityIdPresence.new
```
jakubmiarka: isn't this equivalent the `TransactionSimpleIdPresence`? They are both being set at the same time so I can't imagine a scenario where TransactionSimpleId was present but TransactionEntityId wasn't.

JohnWattsGDS: It must happen though as there are errors in Sentry (albeit only 35).  Maybe we can log the `TransactionSimpleId` when this error occurs and see if there's one particular source of this sort of error.  Alternatively we can check inside the `TransactionSimpleIdPresence` validator instead of a separate validator but I can't see how it's `nil` when it gets to `PATH_PREFIX.join(IDP_LIST_SINGLE_IDP_SUFFIX % { transaction_name: CGI.escape(transaction_id) }).to_s` if it isn't nil inside the controller at the start of all the method calls!

PR #669, opened 2019-02-27
dCnwuq3t: Add special text for UC users on IDP disconnection

bjgill:
See https://trello.com/c/dCnwuq3t/112-use-cookie-to-remind-sign-in-users-that-those-rm-cs-will-no-longer-work

Co-authored-by: John Watts <john.watts@digital.cabinet-office.gov.uk>
Co-authored-by: Alex Newton <alex.newton@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/sign_in_controller.rb
+++ b/app/controllers/sign_in_controller.rb
@@ -22,6 +22,8 @@ class SignInController < ApplicationController
     )
 
     @disconnected_idps = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(current_disconnected_identity_providers_for_sign_in)
+    @idp_disconnected_alternative_html = current_transaction.idp_disconnected_alternative_html
+
     render :index
   end
 
--- a/app/controllers/sign_in_controller.rb
+++ b/app/controllers/sign_in_controller.rb
@@ -13,6 +13,7 @@ class SignInController < ApplicationController
     @suggested_idp = entity_id && retrieve_decorated_singleton_idp_array_by_entity_id(all_identity_providers, entity_id).first
     unless @suggested_idp.nil?
       FEDERATION_REPORTER.report_sign_in_journey_hint_shown(current_transaction, request, @suggested_idp.display_name)
+      @idp_disconnected_hint_html = get_disconnection_hint_text(@suggested_idp.display_name)
     end
 
     @identity_providers = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(current_identity_providers_for_sign_in)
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -137,7 +137,7 @@ en:
       sign_in_idp: Sign in with %{display_name}
       forgot_company: I can’t remember which company verified me
       company_no_longer_verifies_title: "%{company} is no longer a part of GOV.UK Verify"
-      company_no_longer_verifies_text: If you have an identity account with %{company}, you’ll need to %{link}.
+      company_no_longer_verifies_text: If you have an identity account with %{company}, you’ll need to %{link}%{other_ways}.
```
galund: I'm not sure `other_ways` is the right substitution name here, at least I'm not clear why it's called that. Which I think leads to a more fundamental observation:

For translation purposes, we ought to be providing complete phrases, with slots for names etc where necessary... considering variations on word order / grammar between languages, it would I think not necessarily be possible to make actually build a correct translation. Imagine (for example) that some versions of "other ways" might modify the reset of the sentence.

config/locales/cy.yml
```diff
@@ -135,7 +135,7 @@ cy:
       about_link: Dechreuwch nawr
       registration_message_html: Os nad oes gennych gyfrif hunaniaeth, gallwch %{href}.
       last_certified: "Y cwmni ardystiedig diwethaf a ddefnyddiwyd ar y ddyfais hon oedd %{company_name}"
-      any_certified_company: "Gallwch ddefnyddio cyfrif hunaniaeth a sefydlwyd gennych gydag unrhyw yn o'r cwmnïau hyn:"
```
galund: For the ultimate in pull request pedantry, I would like to recommend `o’r` instead of `o'r`.

bjgill:
This is good to merge and release once hub and hub-fed-conf releases have happened (@phantas)

app/controllers/sign_in_controller.rb
```diff
@@ -52,4 +55,12 @@ def unavailable_idps
     api_idp_simple_ids = current_identity_providers_for_sign_in.map(&:simple_id)
     UNAVAILABLE_IDPS.reject { |simple_id| api_idp_simple_ids.include?(simple_id) }
   end
+
+  def get_disconnection_hint_text(idp_name)
+    if current_transaction.idp_disconnected_hint_html.nil?
+      t('hub.signin.company_no_longer_verifies_text', company: idp_name, link: link_to(t('hub.signin.company_no_longer_verifies_link'), begin_registration_path))
+    else
+      format(current_transaction.idp_disconnected_hint_html, company: idp_name, begin_registration_path: begin_registration_path)
```
phantas: I would prefer to have a name more consistent with other related variables than `idp_disconnected_hint_html`. I am happy to go ahead with this now and revisit this next week

PR #670, opened 2019-03-01
GfHEPKRG: Handle Single IDP journey crashing when transaction entity …

JohnWattsGDS:
…id missing

Enable validation for missing transaction entity id.

Co-authored-by: Benjamin Gill <benjamin.gill@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/models/session_validator/transaction_entity_id_presence.rb
+++ b/app/models/session_validator/transaction_entity_id_presence.rb
@@ -2,15 +2,11 @@ class SessionValidator
   class TransactionEntityIdPresence
     ERROR_MESSAGE = "Transaction entity ID can not be found in the user's session".freeze
     def validate(_cookies, session)
-      # When confident of no false positives, remove the log and reinstate the validation failure.
-      # if session.include?(:transaction_entity_id)
-      #   SuccessfulValidation
-      # else
-      #   ValidationFailure.something_went_wrong(ERROR_MESSAGE)
-      # end
-
-      Rails.logger.error(ERROR_MESSAGE) unless session.include?(:transaction_entity_id)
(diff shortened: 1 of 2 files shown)


app/models/session_validator/transaction_entity_id_presence.rb
```diff
@@ -3,14 +3,12 @@ class TransactionEntityIdPresence
     ERROR_MESSAGE = "Transaction entity ID can not be found in the user's session".freeze
     def validate(_cookies, session)
       # When confident of no false positives, remove the log and reinstate the validation failure.
```
bjgill:
```suggestion
```

bjgill:
Also, is the commit message/PR title accurate?

JohnWattsGDS:
Yes, I believe the title is fine.  That's what we're doing; we're handling the case where the transaction entity id is missing.  Originally it was left to cause problems later, then we logged that it had happened but did nothing else and now we handle it in the only way we can - by halting processing and letting the user know that something went wrong.

app/models/session_validator/transaction_entity_id_presence.rb
```diff
@@ -3,14 +3,11 @@ class TransactionEntityIdPresence
...
```
bjgill: Are you sure we want to keep this comment?

JohnWattsGDS: Didn't even notice it.

galund:
You should still fix Benjamin's email address in the commit message :-)

JohnWattsGDS:
Indeed I should - I thought I'd got away with editing it above but I've done it for real now.

PR #671, opened 2019-03-01
BAU: Commonise nil handling

bjgill:
Create a new method to abstract away nil checking. Also remove a bunch of duplicate calls to journey_hint_value.

### DIFF
--- a/app/controllers/partials/journey_hinting_partial_controller.rb
+++ b/app/controllers/partials/journey_hinting_partial_controller.rb
@@ -8,47 +8,46 @@ module JourneyHintingPartialController
     nil
   end
 
+  def with(value)
+    unless value.nil?
+      yield value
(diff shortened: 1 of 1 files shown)


bjgill:
Just been looking into `yield` for the firt time in Ruby, which looks rather neat. I _think_ `with(a) { |a| a.b }` is a nicer than `a=b; !a.nil? && a.c`. Of course, it's also possible that I'm bringing too many of my Python sensibilities.

bjgill:
Or maybe `journey_hint_value&.fetch('RESUMELINK', nil)` is the better solution. I'm not sure...

bjgill:
@phantas @galund - this is what I was talking about earlier. Any thoughts as to what the most idiomatic solution here would be?

PR #674, opened 2019-03-06
[DON'T MERGE YET] Add prometheus metrics endpoint

philandstuff:
This commit adds a prometheus metrics endpoint to frontend.  It has
the same metrics that we expose to statsd.

I copied Metrics::ApiRequestReporter and
Metrics::ControllerActionReporter to Prometheus:: versions, but I
skipped Metrics::ResponseStatusReporter in favour of using the
Prometheus::Client::Collector middleware that instruments all http
requests.

It's expected that we don't expose the /metrics endpoint to the
outside world.  I have raised https://github.com/alphagov/verify-infrastructure/pull/161 to do this.


philandstuff:
This is nearly ready to go. The only remaining issue is that multi-process mode doesn't support summaries with quantiles so we need to change them to either quantileless summaries or histograms.

philandstuff:
Huh, turns out the prometheus client has Done The Right Thing and removed the quantiles for us automatically \o/

PR #679, opened 2019-03-14
BAU - bump Rails version due to vulnerability

phantas:
See https://weblog.rubyonrails.org/2019/3/13/Rails-4-2-5-1-5-1-6-2-have-been-released/


Gemfile.lock
```diff
@@ -409,4 +409,4 @@ RUBY VERSION
    ruby 2.4.2p198
 
 BUNDLED WITH
-   1.17.1
+   2.0.1
```
galund: Did we expect a major version to get bumped? I don't know what this means :-)

jakubmiarka: Paula had a newer version of a bundler, and it shouldn't affect anything, but the rest of us should upgrade too :) `gem update bundler`

bjgill: It looks as if bundler v2 was released in early January (see https://bundler.io/blog/2019/01/03/announcing-bundler-2.html). From the release notes, it looks as if we could continue using v1.17.1, if we wanted to.

PR #681, opened 2019-03-18
tocxYVRS: check for empty IDP response

bjgill:
Give more information on the IDP response received if the relay state is missing. Also update test to actually test an invalid relay state.

See https://trello.com/c/tocxYVRS/150-understand-and-fix-relay-state-was-nil-error


app/controllers/authn_response_controller.rb
```diff
@@ -52,7 +52,10 @@ def country_response
 private
 
   def raise_error_if_session_mismatch(relay_state, session_id)
```
galund: Can we change this function to take `params` as its argument (i.e. we would do the `params['RelayState']`extraction here).

This would make it slightly more logical that we are testing `params` here as part of the validation process?

bjgill: Seems like a good idea!

PR #682, opened 2019-03-20
hub content quick fixes

JohnWattsGDS:
Changes to hub content.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -121,8 +121,8 @@ en:
     start:
       title: Start
       heading: Sign in with GOV.UK Verify
-      answer_yes: This is my first time using Verify
-      answer_no: I’ve used Verify before
+      answer_yes: This is my first time using GOV.UK Verify
+      answer_no: I’ve used GOV.UK Verify before
       continue: Continue
       error_message: Please select an option
     signin:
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -146,38 +146,46 @@ en:
...
+      no_charge: You don't need to pay to use GOV.UK Verify.
       summary: How companies can verify identities
       details_html: |
-        <p>These companies can use their own data, and they can access data like credit records. They’re able to check that information from passports and driving licences is correct.</p>
-        <p>Certified companies meet government privacy standards, so you can trust them with your personal information. They won’t share or sell your data for any other purposes without your consent.</p>
-        <p>There won’t be any effect on your credit score.</p>
-        <p>More than one certified company means you choose who verifies your identity, and it means the system overall is stronger.</p>
+        <p>These companies can check it's really you by using their own records, or by accessing information like credit records. If you have a passport or driving licence, they can check that information too.</p>
+        <p>These 5 companies meet government privacy standards, so you can trust them with your personal information. They won’t share or sell your information for any other purpose without your consent.</p>
```
bjgill: > `5`

Have we documented this somewhere so that we know to change this number when the number of IDPs increases/decreases?

JohnWattsGDS: It's a content thing, the content people take care of content when the content changes (for example, if the number of IDPs changes).

bjgill: OK, so we don't need to record it in the IDP disconnection section of the tech manual?

JohnWattsGDS: It may be worth adding for the sake of belts and braces.

PR #683, opened 2019-03-22
EID-1412: Add a static error page for Proxy Node

CharlesIC:
A static error page to which the eIDAS proxy node can redirect users in case of an unrecoverable error.
This is basically the same as the _Something went wrong_ page but with the man text specific to eIDAS. The text is as designed by James Butler in EID-1229

Screen Shot 2019-03-21 at 17 18 24


app/models/feedback_source_mapper.rb
```diff
@@ -18,6 +18,7 @@ def initialize(product_page_url)
       'FAILED_SIGN_IN_PAGE' => 'failed_sign_in',
       'FAILED_COUNTRY_SIGN_IN_PAGE' => 'failed_country_sign_in',
       'EIDAS_SCHEME_UNAVAILABLE' => 'redirect_to_country',
+      'PROXY_NODE_ERROR' => 'proxy_node_error',
```
jakubmiarka: The convention here is to append `_PAGE`

CharlesIC: Fine... Done

config/locales/cy.yml
```diff
@@ -640,6 +641,8 @@ cy:
       heading: Mae’n ddrwg gennym, mae rhywbeth wedi mynd o’i le
       error_message: Gall hyn fod oherwydd bod eich sesiwn wedi’i amseru allan neu roedd gwall yn y system.
       start_again: Oherwydd bod hwn yn wasanaeth diogel, bydd angen i chi ddechrau eto.
+    proxy_node_error:
+      start_again: Because you were using a secure service from another European country, you’ll need to go back to the page where you started using that service.
```
tomrosier: Has the Welsh been requested?

CharlesIC: Yep, doing the spreadsheet now 👍

PR #689, opened 2019-03-27
New heading for when user cannot use Verify.

galund:
New heading "You won’t be able to use GOV.UK Verify" for when user cannot use Verify replaces "GOV.UK Verify will not work for you".

 - Welsh translations pending (from Land Registry perhaps?)

 - unclear if we should change the title of the page too

https://govuk.zendesk.com/agent/tickets/3622311
https://trello.com/c/GETxCo0p/142-sign-your-mortgage-deed-revision-to-page-content

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -272,7 +272,7 @@ cy:
       choose_another_company: Mae'n rhaid i chi ddewis cwmni ardystiedig wahanol i wirio eich hunaniaeth.
       choose_another_certified_company_link: Dewiswch gwmni arall
     verify_will_not_work_for_you:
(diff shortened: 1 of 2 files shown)


jakubmiarka:
@galund is this still a thing which should be merged or closed?

galund:
> @galund is this still a thing which should be merged or closed?

I believe we agreed to make this change (perhaps check with Claire Hansford). But we were waiting for the translation into Welsh, which was going to be supplied by HMLR I think. The ticket is still open, but we didn't hear from them.

If we prefer this version, we should just get it translated for ourselves. Up to Claire,  I think.

PR #690, opened 2019-03-29
ZD-3643231: Do not accept requests without a SAML message

jakubmiarka:
Currently, when the authn request comes in we pass it on to saml-proxy to validate.
However, if the request doesn't have a SAML, the saml-proxy service raises an exception.
This commit adds the validation to frontend so we fail earlier
and don't pass the request on to saml-proxy.

### DIFF
--- a/app/controllers/authn_request_controller.rb
+++ b/app/controllers/authn_request_controller.rb
@@ -1,5 +1,6 @@
 require 'ab_test/ab_test'
 require 'partials/journey_hinting_partial_controller'
+require 'partials/user_errors_partial_controller'
 
 class AuthnRequestController < SamlController
   include JourneyHintingPartialController
@@ -8,6 +9,8 @@ class AuthnRequestController < SamlController
   skip_before_action :set_piwik_custom_variables
 
   def rp_request
+    return if raise_error_if_params_invalid
+
     session_journey_hint_value = session.fetch(:journey_hint, nil)
     session_journey_hint_rp = session.fetch(:journey_hint_rp, nil)
     create_session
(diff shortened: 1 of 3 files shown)


app/controllers/authn_request_controller.rb
```diff
@@ -8,6 +9,8 @@ class AuthnRequestController < SamlController
   skip_before_action :set_piwik_custom_variables
 
   def rp_request
+    return if raise_error_if_params_invalid
```
bjgill: I assume this means that session_error returns a truthy value if called and so halts processing of the request?

bjgill:
Another question - which error page would users previously have been sent to?

app/controllers/authn_request_controller.rb
jakubmiarka: exactly

jakubmiarka:
> Another question - which error page would users previously have been sent to?

They will stay at the same URL (`/SAML2/SSO`) but will see the [session_error page ](https://github.com/alphagov/verify-frontend/blob/master/app/views/errors/session_error.html.erb)

jakubmiarka:
Thinking about it, it's not a good error page we'll be sending them to.. let me re-think

app/controllers/authn_request_controller.rb
```diff
@@ -117,4 +120,8 @@ def store_rp_request_data
     RequestStore.store[:rp_saml_request] = params.fetch('SAMLRequest', nil)
     RequestStore.store[:rp_relay_state] = params.fetch('RelayState', nil)
   end
+
+  def raise_error_if_params_invalid
+    something_went_wrong_warn("Missing/empty SAML message from #{request.referer}", :bad_request) if params['SAMLRequest'].blank?
```
galund: It's not obvious to me that render_error will definitely throw an exception. Is that the case?

If on the other hand this is about return values, then it shouldn't use the word 'raise'?

app/controllers/authn_request_controller.rb
galund: Hmm yeah definitely don't like the word 'raise' here if we're not throwing exceptions...

jakubmiarka: it raises an error for user, but sure I can rename it...

app/controllers/authn_request_controller.rb
jakubmiarka: It logs warning and raises exception for user... it's not about return values, we just use return to return from the method if we encounter the error

PR #692, opened 2019-04-04
FIREBREAK: Deploy PRs as temporary apps

jakubmiarka:
With the removal of the joint environment, we no longer have an environment where people (e.g. content designers) could go and review/try content changes.

This PR adds some Heroku config in order to allow us to automatically deploy PRs to a temporary environment. It makes use of [Heroku review apps](https://devcenter.heroku.com/articles/github-integration-review-apps). It also introduces a new config variable called `STUB_MODE` to enable the test/stub flows and routes while running in a production mode (in Heroku).

Next steps:
- [ ] configure a bot to integrate with Heroku to trigger these PR builds
- [ ] some work on the stubs to enable as many features as possible (e.g. eIDAS doesn't work)

### DIFF
--- a/Dockerfile.heroku
+++ b/Dockerfile.heroku
@@ -1,0 +1,31 @@
+FROM ruby:2.5.5
+
+ENV RAILS_SERVE_STATIC_FILES true
+ENV RAILS_LOG_TO_STDOUT true
+ENV RAILS_ENV production
+ENV SECRET_KEY_BASE notused
+ENV STUB_MODE true
+
+ADD Gemfile Gemfile
+
+RUN gem install --update bundler
+
+RUN bundle install --binstubs
+
+ADD . /verify-frontend/
+
+WORKDIR /verify-frontend
+
+RUN bundle config --global frozen 1 && \
+    bundle check || bundle install --without development test
+
+RUN echo 'heroku' > /verify-frontend/.build-number
+
+RUN bash -c 'source .env \
+             && export $(cut -d= -f1 .env) \
+             && bundle exec rake assets:precompile' \
+    && ln -s /verify-frontend/public/assets /assets
+
+# Puma needs these dockerignored dirs to write to
+RUN mkdir -p log tmp
+
--- a/Dockerfile.heroku
+++ b/Dockerfile.heroku
@@ -1,3 +1,5 @@
+# Dockerfile for Heroku Review App (deployed as part of a PR)
+
 FROM ruby:2.5.5
 
 ENV RAILS_SERVE_STATIC_FILES true
--- a/README.md
+++ b/README.md
@@ -39,3 +39,18 @@ On an OSX system this amounts to:
 brew install pre-commit
 pre-commit install
 ```
+
+## Deploying the application
+The application is deployed using our [CI/CD pipeline](https://deployer.tools.signin.service.gov.uk/teams/main/pipelines/deploy-verify-hub?groups=build-apps&groups=default). 
+Any changes merged to master are automatically deployed. This repo has an active branch protection for `master`. Any changes need to be raised via PR and approved by two other developers.
+
+## PR reviews
+When a PR is raised, it's automatically tested using Travis (runs the ./pre-commit.sh script on the branch and against master) which is configured in the [.travis file](/.travis). The test results are shown directly on the PR. 
+
+In addition to the Travis tests we have also enabled Codacy to check coding style. Again, the results are shown within the PR. Codacy is configured using the [.rubocop.yml file](/.rubocop.yml).
+
+The PR is also deployed to Heroku as [a review app](https://devcenter.heroku.com/articles/github-integration-review-apps). The app is destroyed when the PR is closed/merged or after 5 days of inactivity. It uses docker to run both the Rails app and the stub API server. The Heroku deployment is configured using the 4 files:
+* `Dockerfile.heroku` - to configure the docker image of frontend
+* `heroku.yml` - Heroku [deployment manifest](https://devcenter.heroku.com/articles/build-docker-images-heroku-yml) 
+* `app.json` - Heroku [application manifest](https://devcenter.heroku.com/articles/app-json-schema)
+* `heroku-startup.sh` - startup script used to start the app and api, with supplied port by Heroku
--- a/app/views/test_saml/idp_request.html.erb
+++ b/app/views/test_saml/idp_request.html.erb
@@ -1,8 +1,11 @@
-SAML Request is '<%= @saml_request %>'
-relay state is '<%= @relay_state %>'
-registration is '<%= @registration %>'
-hints are '<%= @hints %>'
-language hint was '<%= @language_hint %>'
+BANG! You're at an IDP!
+<br><br>
+The IDP received this: <br>
+SAML Request is '<%= @saml_request %>' <br>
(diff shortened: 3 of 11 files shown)


bjgill:
Looks useful for changes to verify-frontend. Am I right in thinking, though, that it won't help us review changes to `verify-hub-federation-config` or `verify-frontend-federation-config`? If so, can this approach be extended to cover them in future?

jakubmiarka:
> Am I right in thinking, though, that it won't help us review changes to `verify-hub-federation-config` or `verify-frontend-federation-config`? If so, can this approach be extended to cover them in future?

That is correct, but we've never been able to review these changes in joint. Particularly when real IDPs or RPs are involved. Their config is private therefore is not used anywhere else. We could extend this to pull the private repos, but given this deployment is public we don't wont to use real IDPs/RPs.

galund:
It might be an interesting exercise to consider what parts of the IDP and (more especially) RP configuration actually in fact *could* be made public.

All of the content that appears for users, for example, is clearly public anyway... and it would be very useful to be able to test changes to that content. In which case, theoretically, we really might be able to stub those out...

... completely accept it's probably quite small beer, but e.g. testing "other ways" content has been something that come up recently.

jakubmiarka:
I do agree. However, we have always been discouraged to mention real IDPs or RPs. Given many are in private betas, in early/sensitive stages of onboarding, not official etc... If that steer has changed that would make testing much easier.

Dockerfile.heroku
```diff
@@ -0,0 +1,31 @@
+FROM ruby:2.5.5
```
galund: This should perhaps have a comment at the top explaining why it exists? e.g. perhaps  `# Heroku deployment for creating code/content review apps`

jakubmiarka: yeah, I was going to update README and reference all the stuff there (and forgot)

app/views/test_saml/idp_request.html.erb
```diff
@@ -1,8 +1,11 @@
-SAML Request is '<%= @saml_request %>'
-relay state is '<%= @relay_state %>'
-registration is '<%= @registration %>'
-hints are '<%= @hints %>'
-language hint was '<%= @language_hint %>'
+BANG! You're at an IDP!
```
galund: Is there are change where previously this was `text/plain` and now it's HTML, or was it just an unreadable mess previously? (I'm just checking if adding markup here will break something else perhaps?)

jakubmiarka: it was just unreadable, because the new lines were not new lines when rendered (because it was always HTML)

config/initializers/00_configuration.rb
```diff
@@ -40,4 +40,5 @@
   option_bool 'feedback_disabled', 'FEEDBACK_DISABLED', default: false
   # Feature flags
   option_bool 'single_idp_feature', 'SINGLE_IDP_FEATURE', default: true
+  option_bool 'stub_mode', 'STUB_MODE', default: false
```
galund: Can we explain somewhere what "stub mode" actually means, i.e. is it only intended for use in a review app, or could it be more generally applicable: if I'm writing a new feature on frontend, I need to understand in what circumstances I should be checking this config option.

Dockerfile.heroku
galund: Sure, I think some very brief note in the file would be good too tho so newcomers don't think it has anything to do with production.

(The `app.json` is clear enough because it literally starts with a description already.)

galund:
(Just to add: I think this is brilliant and it's great to see how few changes we seem to need in order to have a usable standalone app.)

Dockerfile.heroku
jakubmiarka: done here and also in the README

config/initializers/00_configuration.rb
jakubmiarka: better? :)

jakubmiarka:
@galund Thanks for the comments, I think I now addressed them all.

README.md
```diff
@@ -39,3 +39,18 @@ On an OSX system this amounts to:
 brew install pre-commit
 pre-commit install
 ```
+
+## Deploying the application
+The application is deployed using our [CI/CD pipeline](https://deployer.tools.signin.service.gov.uk/teams/main/pipelines/deploy-verify-hub?groups=build-apps&groups=default). 
+Any changes merged to master are automatically deployed. This repo has an active branch protection for `master`. Any changes need to be raised via PR and approved by two other developers.
```
galund: It might be worth just referring to the [release policy](https://verify-team-manual.cloudapps.digital/documentation/welcome/release-process.html) (which is admittedly completely out-of-date) – this info probably doesn't need to redundantly appear in every README we have :-)

README.md
```diff
@@ -39,3 +39,18 @@ On an OSX system this amounts to:
...
+## PR reviews
+When a PR is raised, it's automatically tested using Travis (runs the ./pre-commit.sh script on the branch and against master) which is configured in the [.travis file](/.travis). The test results are shown directly on the PR. 
+
+In addition to the Travis tests we have also enabled Codacy to check coding style. Again, the results are shown within the PR. Codacy is configured using the [.rubocop.yml file](/.rubocop.yml).
+
+The PR is also deployed to Heroku as [a review app](https://devcenter.heroku.com/articles/github-integration-review-apps). The app is destroyed when the PR is closed/merged or after 5 days of inactivity. It uses docker to run both the Rails app and the stub API server. The Heroku deployment is configured using the 4 files:
+* `Dockerfile.heroku` - to configure the docker image of frontend
+* `heroku.yml` - Heroku [deployment manifest](https://devcenter.heroku.com/articles/build-docker-images-heroku-yml) 
+* `app.json` - Heroku [application manifest](https://devcenter.heroku.com/articles/app-json-schema)
+* `heroku-startup.sh` - startup script used to start the app and api, with supplied port by Heroku
```
galund: "on the port supplied by Heroku" is presumably what's meant here

config/initializers/00_configuration.rb
galund: I think so, though I'm not entirely clear what "when compiled for a production env" means. Does that mean that the Heroku environment does count as prod, and furthermore this flag has no effect in a non-prod environment? Both those things might be a bit confusing :-)

bjgill:
Another thought - has Heroku been enabled yet? It might be useful to be able to see a real example of the system working in this PR?

jakubmiarka:
> My only question would be whether we should be putting the new files into the root directory. It might reduce the clutter if we could put them in a `heroku` directory, perhaps? I'm not sure I feel too strongly, however.

Thanks @bjgill, I was considering the subfolder, but certain files (like heroku.yml and app.json) must be in the root for Heroku to pick them up. And adding the other two to a subdirectory might only create confusion. We have other stuff there, like travis config etc...

> Another thought - has Heroku been enabled yet? It might be useful to be able to see a real example of the system working in this PR?

It's enabled, but until the app.json is in `master` it's not going to do anything :(

config/initializers/00_configuration.rb
jakubmiarka: `Rails.env` or `ENV['RAILS_ENV']` ... There are three options (development, test, production). When compiling for production, the stub routes (i.e. /test-saml) are not enabled. The `STUB_MODE` config makes it possible by opening the routes.

PR #695, opened 2019-04-16
BAU: Stop relying on a user's browser system clock

adityapahuja:
Currently, Verify Frontend uses JavaScript to compare session's expiry time in UTC taken from the server against the user's browser system clock. The issue with this approach is that when the user's browser system clock is set to a wrong time unintentionally (e.g. adding 2 hours forward from the true time), the user will always be redirected to timeout page since the user's browser clock is always after the session's expiry time taken from the server. Verify Frontend should do session expiry validation on the server side instead of the client side.

This commit changes Verify Frontend to calculate the number of seconds it takes for the session to timeout or expire and provide it instead of the session's expiry time to the JavaScript. The JavaScript then uses it to add the timeout to the user's browser clock to obtain the session's expiry time. This changes will ensure that the user will not be redirected to the timeout page if the user's browser clock is set to wrong time.

Author: @adityapahuja


app/views/further_information/_modal_dialog.html.erb
```diff
@@ -1,4 +1,4 @@
-<dialog id="js-modal-dialog" data-utc-expiry="<%= session[:assertion_expiry] %>" data-seconds-modal-visible="300"
+<dialog id="js-modal-dialog" data-seconds-to-timeout="<%= (Time.parse(session[:assertion_expiry]) - Time.now.utc).to_i %>" data-seconds-modal-visible="300"
```
jakubmiarka: I would prefer to have the logic in the controller instead of the view

adityapahuja: This has been done. Thanks Jakub!

PR #696, opened 2019-04-29
Fbq1 19 update support form page

JohnWattsGDS:


spec/feature_helper.rb
```diff
@@ -193,6 +193,11 @@ def variant_session
         selected_answers: { documents: { driving_licence: false }, device_type: { device_type_other: true } },
     }
   end
+
+  def navigate_to_feedback_form(locale = 'en')
+    click_link id: t('feedback_link.id')
```
jakubmiarka: just curious, any reason why we added the `id` instead of using the text as previously?

config/locales/en.yml
```diff
@@ -48,6 +48,7 @@ en:
     redirect_to_service_signing_in: redirect-to-service/signing-in
     redirect_to_service_start_again: redirect-to-service/start-again
     redirect_to_service_error: redirect-to-service/error
+    feedback_landing: feedback_landing
```
jakubmiarka: we use hyphens in the urls, rather than underscores ;)

spec/feature_helper.rb
JohnWattsGDS: During one iteration, the text of the link changed because it included the service used and I was getting brain-ache trying to work out how to tie it in with the test.  Then it went back to the original link text and I didn't change the test back (I guess because it didn't fail, so didn't think about it).

config/locales/cy.yml
```diff
@@ -461,6 +463,20 @@ cy:
       title: Llwytho
       heading: Mynd ymlaen i’r cam nesaf
       description: Oherwydd nad yw Javascript wedi’i alluogi ar eich porwr, mae’n rhaid i chi bwyso’r botwm parhau
+    feedback_landing:
+      title: Help
+      basic_heading: Cael help i ddefnyddio GOV.UK Verify
+      heading: Get help accessing %{service_name}
```
jakubmiarka: how come the other text is translated and not this sentence? missed or no translation available? (and `title`)

JohnWattsGDS: That was an oversight on my part (I've requested the translation).  As for title - it turns out the Welsh for help is help!

JohnWattsGDS: And Claire got us the other translation back in record time too!

PR #701, opened 2019-05-22
EID-1463: check the entity_id parameter

tomrosier:
validate the entity_id parameter, if its empty then show the user the something wrong page.

### DIFF
--- a/app/controllers/single_idp_journey_controller.rb
+++ b/app/controllers/single_idp_journey_controller.rb
@@ -37,9 +37,13 @@ class SingleIdpJourneyController < ApplicationController
   end
 
   def continue
-    select_viewable_idp_for_single_idp_journey(params.fetch('entity_id')) do |decorated_idp|
-      select_idp(decorated_idp.entity_id, decorated_idp.display_name)
-      redirect_to redirect_to_single_idp_path
+    if params_are_missing(%w(entity_id))
+      redirect_to start_path
+    else
+      select_viewable_idp_for_single_idp_journey(params.fetch('entity_id')) do |decorated_idp|
+        select_idp(decorated_idp.entity_id, decorated_idp.display_name)
+        redirect_to redirect_to_single_idp_path
+      end
     end
   end
 
--- a/spec/controllers/single_idp_journey_controller_spec.rb
+++ b/spec/controllers/single_idp_journey_controller_spec.rb
@@ -289,6 +289,25 @@ describe SingleIdpJourneyController do
 
       expect(subject).to redirect_to redirect_to_single_idp_path
     end
+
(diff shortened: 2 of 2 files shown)


app/controllers/single_idp_journey_controller.rb
```diff
@@ -37,16 +37,24 @@ def continue_to_your_idp
...
+      end
     end
   end
 
   def continue_ajax
-    select_viewable_idp_for_single_idp_journey(params.fetch('entityId')) do |decorated_idp|
-      select_idp(decorated_idp.entity_id, decorated_idp.display_name)
-      ajax_idp_redirection_single_idp_journey_request(single_idp_cookie&.fetch('uuid', nil))
+    if params_are_missing(%w(entityId))
+      redirect_to start_path
```
jakubmiarka: this is an ajax call (`PUT`), so the `redirect_to` will not do anything, unless it returns an error code. We should return an error HTTP response so it falls back to the standard `POST` request and can do the redirection. Alternatively we don't need the check the params here is it will error out on the `params.fetch` and fallback anyway

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -152,6 +152,23 @@
         expect(cookies.encrypted[CookieNames::VERIFY_SINGLE_IDP_JOURNEY]).to be(nil)
       end
     end
+
+    describe 'entityid is invalid' do
```
jakubmiarka: isn't this test same as https://github.com/alphagov/verify-frontend/blob/8b3c848caba67105753ca8afd326019e6dbd16e7/spec/controllers/single_idp_journey_controller_spec.rb#L108

It would be better to add the test in the continue context https://github.com/alphagov/verify-frontend/blob/8b3c848caba67105753ca8afd326019e6dbd16e7/spec/controllers/single_idp_journey_controller_spec.rb#L287 we currently only have the happy condition, not the failure one

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -289,6 +289,25 @@
...
+
+    describe 'invalid idp (entity idp id missing)' do
+      fake_stub_idp = 'fake-stub-idp'
+
+      valid_idp = {
+          'simple_id' => 'stub-idp-one',
+          'entity_id' => fake_stub_idp,
+          'levels_of_assurance' => %w(LEVEL_1 LEVEL_2)
+      }
+      it 'should redirect to the start page and not set a cookie when an incorrect idp is supplied' do
```
CharlesIC: Suggest renaming this to `...when an IDP entity ID is missing` - since it's not been supplied at all in this test

PR #708, opened 2019-05-31
EID-1463: Missing Entity ID for single IDP journey test

tomrosier:


spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -302,6 +302,22 @@
...
 
+        set_selected_idp(valid_idp)
+        post :continue, params: { locale: 'en', entity_id: '' }
+
+        expect(subject).to redirect_to start_path
+      end
+    end
+
+    describe 'invalid idp (entity idp id missing)' do
+      valid_idp = {
```
jakubmiarka: you don't need any of this ;)

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -302,6 +302,22 @@
...
+          'simple_id' => 'stub-idp-one',
+          'levels_of_assurance' => %w(LEVEL_1 LEVEL_2)
+      }
+      it 'should redirect to the start page and not set a cookie when an incorrect idp is supplied' do
+        stub_api_select_idp
```
jakubmiarka: lines 318 to 321 can go too. with the new check there, it will never get to that calls.. You only need the `post :continue...` line and expectations

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -290,7 +290,7 @@
       expect(subject).to redirect_to redirect_to_single_idp_path
     end
 
-    describe 'invalid idp (entity idp id missing)' do
+    describe 'invalid idp (entity idp id is empty)' do
```
CharlesIC: Can remove this and rename the test to `should redirect to the start page when supplied IDP entity ID is empty`)

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -302,6 +302,22 @@
...
         stub_request(:get, INTERNAL_PIWIK.url).with(query: hash_including({}))
 
+        set_selected_idp(valid_idp)
+        post :continue, params: { locale: 'en', entity_id: '' }
+
+        expect(subject).to redirect_to start_path
+      end
+    end
+
+    describe 'invalid idp (entity idp id missing)' do
```
CharlesIC: Can remove this and rename the test to `should redirect to the start page when IDP entity ID is missing`)

spec/controllers/single_idp_journey_controller_spec.rb
jakubmiarka: your two tests can look like this, don't need anything else
```
   describe 'invalid idp (entity idp id is empty)' do
      it 'should redirect to the start page and not set a cookie when an incorrect idp is supplied' do
        post :continue, params: { locale: 'en', entity_id: '' }

        expect(subject).to redirect_to start_path
      end
    end

    describe 'invalid idp (entity idp id missing)' do
      it 'should redirect to the start page and not set a cookie when an incorrect idp is supplied' do

        post :continue, params: { locale: 'en' }

        expect(subject).to redirect_to start_path
      end
    end
```

jakubmiarka: also it shouldn't mention anything about cookies

tomrosier: cheers :) will apply that now

PR #709, opened 2019-06-03
EID-1462: Add check to see if entity_id is set

tomrosier:


CharlesIC:
Can we have some unit tests for this as well? :)

app/controllers/single_idp_journey_controller.rb
```diff
@@ -37,7 +37,7 @@ def continue_to_your_idp
   end
 
   def continue
-    if params_are_missing(%w(entity_id))
+    if params_are_missing(%w(entity_id)) || session[:transaction_entity_id].nil?
```
jakubmiarka: it would probably be better to extend our session_validator to validate the transaction_entity_id too, instead of adding it additionally here https://github.com/alphagov/verify-frontend/blob/b785258f8289a39cc70422d53748f7d3f112ff69/app/models/session_validator.rb#L3

tomrosier: looks like there is already a session validator for the transaction entity id `TransactionEntityIdPresence.new`

jakubmiarka: ah yeah! even better (I was looking locally on an old branch).. That means, the condition here is redundant, because if the entity id was missing it should not get to this line at all.

CharlesIC: https://sentry.tools.signin.service.gov.uk/verify/prod-frontend/issues/1148/
```
PATH_PREFIX.join(IDP_LIST_SINGLE_IDP_SUFFIX % { transaction_name: CGI.escape(transaction_id) }).to_s
```
```
no implicit conversion of nil into String
```

@jakubmiarka So why are we seeing these errors in Sentry?

jakubmiarka: That I'd love to know too! 😄

jakubmiarka: maybe presence is not enough and should be amended to check for the value not presence?

CharlesIC: `TransactionEntityIdPresence` checks for `session.include?(:transaction_entity_id)` but doesn't check for what Tom added: `session[:transaction_entity_id].nil?`
So the value may be present but nil, in which case we get the errors. Should we change `TransactionEntityIdPresence` to check for `include?` and `blank?`?

jakubmiarka: I'm puzzled how it ever could be nil. The value is provided by the config service (so unless there's a bug I have no idea 🤷‍♂ ) ... But yeah, I would probably prefer updating the validator to check for `session[:transaction_entity_id].nil?` instead of include or blank.

CharlesIC: Well we need both `include` and and a check for `nil?` (nil will throw if value is not in the hash). `blank?` checks the value is not nil or all whitespace - we use `blank?` for most checks throughout the codebase. I'd suggest just doing `if session.include?(:transaction_entity_id) || session[:transaction_entity_id].blank?` like we do for other checks

jakubmiarka: `nil?` will not throw error if it's not in the hash. As long as the hash (`session`) is there. 
No reason for `blank?` as it's not a user-provided value.

app/models/session_validator/transaction_entity_id_presence.rb
```diff
@@ -2,7 +2,7 @@ class SessionValidator
   class TransactionEntityIdPresence
     ERROR_MESSAGE = "Transaction entity ID can not be found in the user's session".freeze
     def validate(_cookies, session)
-      if session.include?(:transaction_entity_id)
+      if session.include?(:transaction_entity_id) && !session[:transaction_entity_id].nil? && !session[:transaction_entity_id].blank?
```
jakubmiarka: We could make it simpler and more readable by turning the logic around and getting rid of the unnecessary checks
```
if session[:transaction_entity_id].nil? 
     ValidationFailure.something_went_wrong(ERROR_MESSAGE)
else
     SuccessfulValidation
end
```

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -290,6 +290,17 @@
       expect(subject).to redirect_to redirect_to_single_idp_path
     end
 
+    describe 'invalid session entity id' do
+      it 'should redirect to the start page' do
+        stub_api_select_idp
```
jakubmiarka: this is not needed

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -290,6 +290,17 @@
       expect(subject).to redirect_to redirect_to_single_idp_path
     end
 
+    describe 'invalid session entity id' do
```
jakubmiarka:
```suggestion
    describe 'missing transaction entity id in session' do
```

spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -290,6 +290,17 @@
...
+      it 'should redirect to the start page' do
```
jakubmiarka:
```suggestion
      it 'should show an error page' do
```

app/controllers/single_idp_journey_controller.rb
CharlesIC: If `nil?` won't throw then fine 👍

CharlesIC: Will `nil?` check for empty strings too, though?

spec/controllers/single_idp_journey_controller_spec.rb
jakubmiarka: Actually, this whole test probably shouldn't be here at all. Since we used the SessionValidator we should test it there, which it is actually being already tested: https://github.com/alphagov/verify-frontend/blob/b785258f8289a39cc70422d53748f7d3f112ff69/spec/models/session_validator_spec.rb#L142

app/controllers/single_idp_journey_controller.rb
jakubmiarka: > Will `nil?` check for empty strings too, though?

It will not, but I don't see a way how it could be an empty string. The `blank?` method is used to check user input, because users could put in only spaces, but this is a config value.

PR #711, opened 2019-06-04
EID-1463: Log missing single IDP params as warn rather than error

CharlesIC:
We handle missing params by redirecting users to another page so we don't need to see these logs in Sentry


alexandergeegds:
Except now Travis is failing... O.o

PR #715, opened 2019-06-11
BAU: Explicitly allow robots to index all pages

CharlesIC:
We have `noindex` HTML tags on some pages to prevent them from being indexed. Robots need to be able to access those pages to see the tag.
Google thinks some of our pages are being blocked by `robots.txt` whilst we're actually allowing all robots on all pages.
Maybe an explicit Allow directive will fix this.

_Tested with Google's robots.txt tester tool_

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -3,3 +3,8 @@
 # To ban all spiders from the entire site uncomment the next two lines:
 # User-agent: *
 # Disallow: /
+
+# Explicitly allow all pages to be accessed by crawlers
+User-agent: *
+Disallow:
+Allow: /


jakubmiarka:
Does it say which pages it can't access? I think there's a difference between not being able to access stuff and not index it. The `/start` page without a session returns 500 and it might be that the bot doesn't even look at the robots.txt because of that? 🤷‍♂  it would also explain why it doesn't index any other other text and the page title is incorrect in the google result.

CharlesIC:
@jakubmiarka 
It can't access about 14 different pages but the other ones don't show in search results. If you inspect `https://www.signin.service.gov.uk/start` however it explicitly says it's blocked by `robots.txt` even though it shouldn't. It could also be that we return a 401 but some other docs I've found on Google said that it does index 401 pages.
Screen Shot 2019-06-11 at 17 48 27

PR #716, opened 2019-06-12
BAU: Ban robots from indexing

jakubmiarka:
Setting the meta tag is not enough on error pages as Googlebot doesn't read it.
This uses the existing flag to append the the response headers.

I also considered using the nginx rules, but it would be a bit more complex with the
conditions as we actually control which pages to index from frontend.

Also removed some of the flags on pages which I don't believe should be indexed. Like the feedback form?

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -17,6 +17,8 @@ module ApplicationHelper
   end
 
   def hide_from_search_engine?
+    response.set_header("X-Robots-Tag", "noindex") if !content_for(:show_to_search_engine)
+
     !content_for(:show_to_search_engine)
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -17,9 +17,10 @@ module ApplicationHelper
   end
 
   def hide_from_search_engine?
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -17,6 +17,8 @@ def analytics_title(english_title)
   end
 
   def hide_from_search_engine?
+    response.set_header("X-Robots-Tag", "noindex") if !content_for(:show_to_search_engine)
```
CharlesIC: Wouldn't it be better to refactor this method to

```
if !content_for(:show_to_search_engine)
    response.set_header("X-Robots-Tag", "noindex")
    return true
end

return false
```

jakubmiarka: I did it slightly differently:
```
def hide_from_search_engine?
    return false if content_for(:show_to_search_engine)

    response.set_header("X-Robots-Tag", "noindex")
    true
end
```

Can you re-approve?

CharlesIC: Nicely done!

CharlesIC:
Should we also revert `robots.txt` to be empty?

jakubmiarka:
> Should we also revert `robots.txt` to be empty?

I would leave it for now and see what difference this makes and then do the cleanup?

PR #717, opened 2019-06-13
EID-1519: Update the cookies page for AB test

CharlesIC:
The AB test cookie is valid for 2 weeks, not 1

### DIFF
--- a/app/models/ab_test/ab_test.rb
+++ b/app/models/ab_test/ab_test.rb
@@ -25,7 +25,7 @@ module AbTest
   end
 
   def self.set_ab_test_cookie(value, cookies)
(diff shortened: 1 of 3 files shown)


jakubmiarka:
It's 1 week

https://github.com/alphagov/verify-frontend/blob/019a68666f93b1aadeb9cceedbe01ba11947feb9/app/models/ab_test/ab_test.rb#L28

CharlesIC:
It should be two weeks, though....

CharlesIC:
@jakubmiarka Does that look good now?

jakubmiarka:
> It should be two weeks, though....

Should it? We changed it last year in https://github.com/alphagov/verify-frontend/pull/519

CharlesIC:
Yes but the UC test will run for two weeks

PR #720, opened 2019-07-01
BAU: find out whether errors are recoverable

bjgill:
Add logging to test whether we have enough information to perform partial reecovery when users are shown the 'something has gone wrong' page. If this shows a good number of users could be helped, we'll redirect to a more helpful error page. Otherwise, this experiment should be reverted.

See https://trello.com/c/kl73OxUJ/241-improve-the-something-went-wrong-experience for more information.

### DIFF
--- a/app/controllers/partials/user_errors_partial_controller.rb
+++ b/app/controllers/partials/user_errors_partial_controller.rb
@@ -35,7 +35,6 @@ module UserErrorsPartialController
 
   # How often do we have the information needed to redirect the user back to the
   # service start page?
-  # TODO: if this works, start using RedirectToServiceController (or similar).
   def check_whether_recoverable
     begin
       if session && current_transaction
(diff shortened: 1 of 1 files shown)


app/controllers/partials/user_errors_partial_controller.rb
```diff
@@ -33,13 +33,31 @@ def upstream_error(exception)
     end
   end
 
+  # How often do we have the information needed to redirect the user back to the
+  # service start page?
+  # TODO: if this works, start using RedirectToServiceController (or similar).
```
JHjava: We generally try to avoid adding TODO's to the codebase.

bjgill: OK - I was intending this as a marker that this bit of code shouldn't stay in its current form for more than a week.

```suggestion
```

PR #721, opened 2019-07-01
refactor: use concerns to replace partial controllers

chrisholmes:
We previously introduced some mixins that were named `PartialContoller`s to
decompose some of the behaviour defined in `ApplicationController`.
Not surprisingly, Rails has its own way do this kind of decomposition
with `ActiveSupport::Concern`. This change moves these mixins to use
`ActiveSupport::Concern`.

### DIFF
--- a/spec/controllers/single_idp_journey_controller_spec.rb
+++ b/spec/controllers/single_idp_journey_controller_spec.rb
@@ -2,7 +2,7 @@ require 'rails_helper'
 require 'controller_helper'
 require 'api_test_helper'
 require 'piwik_test_helper'
-require 'partials/user_cookies_partial_controller'
+# require 'concerns/user_cookies'
 
 describe SingleIdpJourneyController do
   VALID_TEST_RP = 'http://www.test-rp.gov.uk/SAML2/MD'.freeze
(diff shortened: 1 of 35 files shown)


spec/controllers/single_idp_journey_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'controller_helper'
 require 'api_test_helper'
 require 'piwik_test_helper'
-require 'partials/user_cookies_partial_controller'
+# require 'concerns/user_cookies'
```
jakubmiarka: is this meant to be here?

app/controllers/certified_company_unavailable_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+        IdentityProvider.new('simpleId' => simple_id, 'entityId' => simple_id, 'levelsOfAssurance' => [])
+      )
+      @other_ways_description = current_transaction.other_ways_description
+      @other_ways_text = current_transaction.other_ways_text
+      render 'index'
+    else
+      render 'errors/404', status: 404
+    end
+  end
+end
```
bjgill: What's this file for?

galund:
I'm closing this: it's linked from the [Trello card](https://trello.com/c/8kykoFem/96-replace-partials-with-ruby-concerns-in-frontend), and clearly some substantial further work would be required to sort out the conflicts.

PR #723, opened 2019-07-02
BAU: Try not to present users with a list of all services

bjgill:
When a user hits the 'something has gone wrong' page, we used to expect them to choose their service from a list of all the services. This change sees whether we know the service the user came from. If so, we give the user a simple link to click to go back to the start. Additionally, present the user with information on other ways to acieve their goal, in case Verify is repeatedly failing for them.

I intend to merge this tomorrow, after #722 has had some time in production, so I can be sure that this will help some meaningful number of users.

### DIFF
--- a/app/views/errors/something_went_wrong.html.erb
+++ b/app/views/errors/something_went_wrong.html.erb
@@ -6,10 +6,10 @@
   <div class="column-two-thirds">
     <h1 class="heading-large"><%= t 'errors.something_went_wrong.heading' %></h1>
     <p><%= t 'errors.something_went_wrong.error_message' %></p>
-    <p><%= t 'errors.something_went_wrong.start_again' %></p>
+    <p><%= t 'errors.something_went_wrong.start_again_text' %></p>
 
     <% if local_assigns[:transaction_homepage] %>
-      <p><%= link_to t('errors.session_timeout.start_again'), transaction_homepage, class: 'button' %></p>
+      <p><%= link_to t('errors.something_went_wrong.start_again_link'), transaction_homepage, class: 'button' %></p>
 
       <% if local_assigns[:other_ways_text] && local_assigns[:other_ways_description] %>
         <%= render partial: 'shared/other_ways', locals: {other_ways_text: other_ways_text, other_ways_description: other_ways_description} %>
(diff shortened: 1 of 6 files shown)


bjgill:
For reference, the old page (which we still display if a user's session doesn't have an RP):

Screen Shot 2019-07-02 at 16 51 08


and the new page:

Screen Shot 2019-07-02 at 16 49 25

app/views/errors/something_went_wrong.html.erb
```diff
@@ -7,7 +7,14 @@
     <h1 class="heading-large"><%= t 'errors.something_went_wrong.heading' %></h1>
     <p><%= t 'errors.something_went_wrong.error_message' %></p>
     <p><%= t 'errors.something_went_wrong.start_again' %></p>
-    <% if transaction_taxon_list.any? %>
+
+    <% if local_assigns[:transaction_homepage] %>
+      <p><%= link_to t('errors.session_timeout.start_again'), transaction_homepage, class: 'button' %></p>
```
bjgill: > `errors.session_timeout.start_again`

This does work, but I need to do something better here, I suspect...

bjgill:
@jakubmiarka and @JHjava - thanks for your reviews of my earlier PRs. This is the final one in the chain, and actually starts helping users. If you could review this tomorrow morning, I'd be much obliged - I'm off Thursday and Friday, so don't want to leave it too late.

jakubmiarka:
Looking good! I have a few questions:
- Do we know the source of the errors or what kind of users is this affecting? 
- I would also add it to the timeout page (session expired) where it can be even more helpful.
- Has someone from the content team looked at it? Just being a but wary of hiding all the other services, if it's a stuck session from a few days before and we only give them the single service option

bjgill:
Some good points.

* No. It's currently very hard to look up what's causing the errors in Kibana. I've raised a small PR to fix this - https://github.com/alphagov/verify-frontend/pull/724
* Agree that this would be useful on other pages as well. Once this PR has been merged, I'll look at doing the same for the other pages.
* No, I've not been able to get time from a content person. That is a good point. I'd hope that https://github.com/alphagov/verify-frontend/pull/724 would show whether this was something we needed to worry about. From code reading, I think users with expired sessions get shown either the `cookie_expired` or `session_timeout` pages instead.

So, I think the right solution would probably be to put this on hold, and get https://github.com/alphagov/verify-frontend/pull/724 merged today. When I get back on Monday, I can then take a look at the logs. If I am seeing timeout errors, I'll probably withdraw this. Otherwise, I'll run it by a content person and then ask for code reviews.

PR #724, opened 2019-07-03
BAU: log error causes

bjgill:
Currently, we just log the raw exception. This makes it hard to work out what is causing the error pages to be shown. By introducing a log that includes both the name of the partial and the error message, we should then be able to do a search in the logs that will show us what is causing the errors.

This was prompted by https://github.com/alphagov/verify-frontend/pull/723#issuecomment-507752314

### DIFF
--- a/app/controllers/partials/user_errors_partial_controller.rb
+++ b/app/controllers/partials/user_errors_partial_controller.rb
@@ -57,12 +57,14 @@ module UserErrorsPartialController
 
   def something_went_wrong(exception, status = :internal_server_error)
     logger.error(exception)
+    logger.info("Something went wrong: #{exception.try(:message) || exception}")
     check_whether_recoverable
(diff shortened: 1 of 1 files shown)


bjgill:
OK - just realised that there was a simpler way to achieve my goal here. Should now be ready to review. Ideally, I'd like to get this in today, since I'm off for the rest of the week.

bjgill:
The Kibana search for this log would be `"Something went wrong" AND log.service-name:front`. I've checked, and there is nothing in the past 30 days in Kibana that matches that pattern.

jakubmiarka:
> The Kibana search for this log would be `"Something went wrong" AND log.service-name:front`. I've checked, and there is nothing in the past 30 days in Kibana that matches that pattern.

Cool then! You have my tick :) .... (I was thinking the page title could be caught by that)

PR #733, opened 2019-09-05
TfEzFC4W: implement cross domain google analytics tracking for verify

chrisclayson:
## What

We need to implement some Google Analytics (GA) tracking code on the Hub.

This isn't because we're going to use GA for Verify analytics.

It's because GOV.UK is now implementing GA tracking across various x.gov.uk subdomains.

GDS needs to be able to track user journeys across campaign pages, GOV.UK content and transactions on the service domain.

Verify is part of these journeys, hence the need to get the GA code up and running on the Hub. 

## How

Add GoogleAnalytics class for configuration
Add configuration class instance to AnalyticsPartialController
Initialise GA config from environment
Add GA shared view
Render GA shared view in application layout
Update CSP to allow GA source domain
Update analytics.js to execute GA code if config values present
Add `_ga` parameter handling to `authn_request_controller`
Adjust CSP to allow Google Analytics to make calls to www.google-analytics.com
Correct links to `gov.uk` to `www.gov.uk` to ensure link decoration works
Add form decoration and ensure works on SAML response form


.env
```diff
@@ -31,3 +31,6 @@ ZENDESK_TOKEN=123
 SAML_PROXY_HOST=http://localhost:50220
 
 VERIFY_PRODUCT_PAGE=https://govuk-verify.cloudapps.digital/
+
+CROSS_GOV_GOOGLE_ANALYTICS_TRACKER_ID=UA-145652997-1
+CROSS_GOV_GOOGLE_ANALYTICS_DOMAIN_LIST=www.gov.uk localhost
```
alexandergeegds: Why is `localhost` on this list??

app/assets/javascripts/auto_submit_form.js
```diff
@@ -5,7 +5,9 @@
 
   GOVUK.autoSubmitForm = {
     attach: function () {
-      $('form.js-auto-submit').submit();
+        var destinationLink = document.getElementById('saml-response-form');
```
alexandergeegds: We now have a combination of jQuery and native API lookups here. Much though it pains me to suggest using jQuery over the native API, the following might be more consistent:

```suggestion
        var destinationLink = $('#saml-response-form')[0];
```

Also, this file seems to be using 2 space indentation, whereas this block is now indented by 4. Not sure if this is a problem or not. (I would personally applaud the switch to 4 space indenting... 😅 )

.env
chrisclayson: The `.env` file is only used when running locally and as everything is run as `localhost` if you want to see form decoration in action `localhost` needs to be in this list.

In production and other 'real' environments `localhost` will not be present.

alexandergeegds: Ahh 😌 👌

app/views/shared/_ga.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if cross_gov_ga.enabled? %>
+  <span class="hidden" id="cross-gov-ga-tracker-id"><%= cross_gov_ga.tracker_id %></span>
+  <span class="hidden" id="cross-gov-ga-domain-list"><%= raw(cross_gov_ga.cross_domain_list.to_json) %></span>
+  <noscript>
```
galund: We can remove 👍

app/assets/javascripts/analytics.js
```diff
@@ -23,54 +23,51 @@
...
-        ga('create', cross_gov_ga_tracker_id, 'auto', 'govuk_shared', {'allowLinker': true});
-        ga('govuk_shared.require', 'linker');
-        ga('govuk_shared.linker.set', 'anonymizeIp', true);
-        ga('govuk_shared.linker:autoLink', domain_list, false, true);
-        ga('govuk_shared.send', 'pageview');
+        var domain_list = JSON.parse($("#cross-gov-ga-domain-list").text());
+        ga("create", cross_gov_ga_tracker_id, "auto", "govuk_shared", {"allowLinker": true});
+        ga("govuk_shared.require", "linker");
+        ga("govuk_shared.linker.set", "anonymizeIp", true);
+        ga("govuk_shared.linker:autoLink", domain_list, false, false);
```
alexandergeegds: If this is now `false` again, do we need the two optional parameters at all?

Why is it false again?

app/assets/javascripts/analytics.js
```diff
@@ -1,60 +1,68 @@
...
+      piwikAnalyticsQueue.unshift(['setCustomUrl', customUrl]);
     }
 
     global._paq = piwikAnalyticsQueue;
+  }
+
+  var cross_gov_ga_tracker_id = $("#cross-gov-ga-tracker-id").text();
+  if (cross_gov_ga_tracker_id) {
+    var domain_list = JSON.parse($("#cross-gov-ga-domain-list").text());
+    ga("create", cross_gov_ga_tracker_id, "auto", "govuk_shared", {"allowLinker": true});
```
alexandergeegds: I'm not sure if switching from `ga` to `window.ga` would be enough to satisfy Codacy? Otherwise, we could try to mimic the snippet - it should never get called since the snippet will have been run by now.

```suggestion
    if (!window['ga']) {
        window.ga = function() {
            if (!window.ga.q) {
                window.ga.q = [];
            }
            window.ga.q.push(arguments)
        }
    }
    ga("create", cross_gov_ga_tracker_id, "auto", "govuk_shared", {"allowLinker": true});
```

🤷‍♂

.env
```diff
@@ -31,3 +31,6 @@ ZENDESK_TOKEN=123
 SAML_PROXY_HOST=http://localhost:50220
 
 VERIFY_PRODUCT_PAGE=https://govuk-verify.cloudapps.digital/
+
+CROSS_GOV_GOOGLE_ANALYTICS_TRACKER_ID=UA-145652997-1
```
galund: It's not a secret, but probably this shouldn't be repeated multiple in times in a public repo. As we don't actually have any access to the GA instance, so can't do any testing locally, there doesn't seem much point in having the real one.

app/assets/javascripts/analytics.js
chrisclayson: An error ... didn't mean to commit it with `false`

.env
chrisclayson: Done

app/assets/javascripts/analytics.js
chrisclayson: Codacy is happy, will leave as it is for now :-)

galund:
Get an OK from Tom or Michaela before putting live please – not until Monday now I think.

app/assets/javascripts/auto_submit_form.js
```diff
@@ -5,9 +5,7 @@
 
   GOVUK.autoSubmitForm = {
     attach: function () {
-        var destinationLink = $("#saml-response-form")[0];
-        window.ga("govuk_shared.linker:decorate", destinationLink);
-        $("form.js-auto-submit").submit();
+      $("#continue-button").click();
```
alexandergeegds: Is it worth adding a comment (or something equivalent) to say that the form cannot auto-submit directly because of GA tracking?

brenetic: 👍

app/controllers/test_saml_controller.rb
```diff
@@ -6,6 +6,10 @@ class TestSamlController < ApplicationController
   layout 'test'
 
   def index
+    @form_action = '/SAML2/SSO'
```
brenetic: worth making this a constant?

app/controllers/partials/analytics_partial_controller.rb
```diff
@@ -7,6 +7,10 @@ def cross_gov_ga
     CROSS_GOV_GA
   end
 
+  def set_render_cross_gov_uk
```
galund: For consistency shouldn't this be called `set_render_cross_gov_ga`?

app/assets/javascripts/auto_submit_form.js
chrisclayson: Commit pushed

PR #734, opened 2019-09-05
BAU: Enable prove-identity for loa1 journeys

tomrosier:
To enable Italy to use our services at LOA1

### DIFF
--- a/config/main_routes.rb
+++ b/config/main_routes.rb
@@ -7,6 +7,8 @@ post 'sign_in', to: 'sign_in#select_idp', as: :sign_in_submit
 get 'begin_sign_in', to: 'start#sign_in', as: :begin_sign_in
 
 constraints IsLoa1 do
+  get 'prove_identity', to: 'prove_identity#index', as: :prove_identity
+  get 'prove_identity_retry', to: 'prove_identity#retry_eidas_journey', as: :prove_identity_retry
   get 'start', to: 'start#index', as: :start
   post 'start', to: 'start#request_post', as: :start
   get 'begin_registration', to: 'start#register', as: :begin_registration


CharlesIC:
@jakubmiarka We might need your input on this

This change will enable eIDAS for all LOA1 services that also have eIDAS enabled, however, we don’t want services to be able to access eIDAS with LOA1. We only want Test RP in integration to initiate an eIDAS journey at LOA1 (because Italy test credentials are LOA1).
Now - it may be that all services that have eIDAS enabled flag are also at least LOA2 - in that case that’s fine (but flimsy)
Hub-fed-config only says whether `eidasEnabled` is true or false for each service and the LOA the service supports. We have this this blanket block for LOA1 + eIDAS that that this change would remove, so we have to make sure this is the right thing to do or find a workaround.

jakubmiarka:
The assumption is correct, the users will not be redirected to `prove-identity` if the transaction doesn't support eidas. Which is only set on LOA2 services, otherwise people would be already being redirected to that page with 404 (as it's constrained). So its' already flimsy.
You can test this without making a frontend change - enable the eidas on the test-rp-loa1 (or whichever you're using) and go manually to `/choose-a-country` and there you are. Assuming there are no further changes required in the hub/policy... 
If you're worried about the LOA1+eIDAS block I'd suggest adding additional check for LOA on this line: https://github.com/alphagov/verify-frontend/blob/b21c86b3df9f99c8cec00c150c064fa630d39fd5/app/controllers/authn_request_controller.rb#L107 and this validator: https://github.com/alphagov/verify-frontend/blob/master/app/controllers/partials/eidas_validation_partial_controller.rb

CharlesIC:
Thanks @jakubmiarka , that's super helpful.

@tomrosier fine to apply this change given what Jakub has said, just add his suggested change to do more checks.

Wynndow:
@CharlesIC STOP DOING WORK

PR #741, opened 2019-10-09
Fix for incorrect form auto-submission

galund:
Only auto-submit forms that are marked with the js-auto-submit class
 - the form on the 'further information' (cycle 3 data request) screen was automatically being submitted, resulting in an error message appearing for users;
 - add a test to catch that specific symptom on the 'further information' page.

https://govuk.zendesk.com/agent/tickets/3819384


spec/features/user_visits_further_information_page_spec.rb
```diff
@@ -146,6 +146,16 @@
...
     end
+
+    it 'will avoid displaying an error message until the form is submitted by the user' do
+      stub_cycle_three_attribute_request('NationalInsuranceNumber')
+
+      stub_matching_outcome
+
+      visit further_information_path
+
+      expect(page).not_to have_css '.error-message', text: t('hub.further_information.attribute_validation_message', cycle_three_name: attribute_name)
```
jakubmiarka: you could just move this expectations to the beginning of the previous test, instead of creating a new test...

galund: Hoho yeah I've had people tell me off before, for not keeping tests separate. I'll change it, less code is better.

galund: Done

PR #742, opened 2019-10-18
BAU: Upgrade to latest Ruby version (2.6.5)

Rosa-Fox:


Rosa-Fox:
> The Heroku fails to deploy as it also has a ruby version specified:
> 
> https://github.com/alphagov/verify-frontend/blob/73a8356e356ed7b8708956d3b658fbc851576b3b/Dockerfile.heroku#L3
> 
> There will also be changes in the pipeline.

Thanks @jakubmiarka... weirdly this file doesn't open up with the rest of the project files in my text editor, but does open individually. I have updated it and amended the commit.

PR #743, opened 2019-10-18
Updated 'Cookies and analytics' section

thomasjhughes1:
Explained what information is shared with the rest of GDS using Google Analytics. I've also updated the 'Last updated' date.

We'll arrange for the new content to be translated into Welsh next week.


thomasjhughes1:
@jakubmiarka I've requested a translation of these changes and will update the Welsh version later in the week.

PR #744, opened 2019-10-18
EID-1733: warn instead of error on choose a country (non-eIDAS)

galund:
We expect this error only to occur if the user goes to the country controller by mistake, by manually entering the URI, or by having multiple Verify sessions running at the same time (ie from different transaction one of which does not support eIDAS).

> Not sure if we want to land this – see discussion on ticket.

https://govukverify.atlassian.net/browse/EID-1733

### DIFF
--- a/app/controllers/partials/eidas_validation_partial_controller.rb
+++ b/app/controllers/partials/eidas_validation_partial_controller.rb
@@ -2,7 +2,7 @@ module EidasValidationPartialController
   def ensure_session_eidas_supported
     txn_supports_eidas = session[:transaction_supports_eidas]
     unless txn_supports_eidas
-      something_went_wrong('Transaction does not support Eidas', :forbidden)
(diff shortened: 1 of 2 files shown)


CharlesIC:
Thanks for the detailed info in the story. I think that, as you said, this is due to a user action that we cannot recover from or prevent, it's okay to warn instead of error. We _could_ consider redirecting to the IDP picker page, however, if they' get to the `choose_a_country_page` for a transaction that doesn't support eIDAS. That may be a better way of recovering from this than an error page. @jakubmiarka what do you think?

jakubmiarka:
Probably question for UX/service design. Where to redirect them. The user would have a session, so we should be able to redirect them to the /start page or IDP sign-in ... personal preference would be /start since we don't know whether the user has an account or not, but 🤷‍♂

galund:
Let's do https://github.com/alphagov/verify-frontend/pull/745 instead.

PR #745, opened 2019-10-21
ZD-3731900 When your service doesn't support eIDAS but the user think…

JohnWattsGDS:
…s it should

This addresses two problems.

The first is that if a user replaces /start with /redirect-to-country on a service that doesn't support eIDAS the front end logs an error and Sentry starts showing red things.  This isn't really an error in the sense of something that is going wrong because our code isn't handling a particular set of circumstances well so should only be logged as a warning.  In the course of investigating
this, @bjgill discovered that the test that tests this was passing even though it was failing (the second problem being addressed).  The reason for this was that a specific session id was being set in the test but because the session id didn't match what had been set-up, the something went wrong page was being displayed - this was the page that was expected so the test passed anyway.  Deleting the line that sets the session id leaves the test to fail for the right reason - which was verified with the use of a temporary custom error page that could only be rendered from the place it was expected.  The temporary error page doesn't feature in this PR.

This incorporates the changes from EID-1733 which tests the controller directly whereas the feature test tests a user typing a URL in.

### DIFF
--- a/app/controllers/partials/eidas_validation_partial_controller.rb
+++ b/app/controllers/partials/eidas_validation_partial_controller.rb
@@ -2,7 +2,7 @@ module EidasValidationPartialController
   def ensure_session_eidas_supported
     txn_supports_eidas = session[:transaction_supports_eidas]
     unless txn_supports_eidas
-      something_went_wrong('Transaction does not support Eidas', :forbidden)
+      something_went_wrong_warn('Transaction does not support Eidas', :forbidden)
     end
   end
 end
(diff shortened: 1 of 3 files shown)


jakubmiarka:
is it a duplicate of https://github.com/alphagov/verify-frontend/pull/744?

JohnWattsGDS:
Pretty much @jakubmiarka - but if nobody updates their JIRA tickets, how am I supposed to know it's already been done? £$%&* ...

CharlesIC:
EID-1733 is updated with good detail but I assumed this story was a different (but related bug). Fine, we'll decide tomorrow how to handle this.

PR #761, opened 2019-11-29
HUH-234 and HUH-233: Short Hub ABC test

instantiator:
Variant C is mostly implemented, and ready for critique...

Please bear in mind the following:
- The new flow is now in place, with checkboxes.
- The other ways text is drawn from `stub-rp` and it's rubbish.
- The logic maps the 2 test IDPs to `stub-idp-one` and `stub-idp-three` in alphabetical order.

There's a [separate PR](https://github.com/alphagov/verify-frontend-federation-config/pull/142) for the following, which should land first:
- duplicate the `stub` configuration into `verify-frontend-federation-config`.

To follow in future tickets:
- tracking (and any virtual pages), Welsh (or stubbed Welsh).


app/views/select_documents_variant_c/index.html.erb
```diff
@@ -0,0 +1,44 @@
+<%= page_title 'hub_variant_c.select_documents.title' %>
+<% content_for :feedback_source, 'SELECT_DOCUMENTS_PAGE' %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= form_for @form, url: select_documents_submit_path, html: {id: 'validate-select-documents', class: 'select-documents-form', novalidate: 'novalidate', 'data-msg': t('hub_variant_c.select_documents.errors.no_selection')} do |f| %>
+       <%= content_tag :div, class: form_question_class do %>
+        <fieldset>
+          <legend>
+            <h1 class="govuk-fieldset__heading heading-large" style="margin-top: 0;"><%= t 'hub_variant_c.select_documents.heading' %></h1>
```
jakubmiarka: `govuk-fieldset__heading` is again from the design system... the inline styles 😱

app/views/select_documents_variant_c/index.html.erb
```diff
@@ -0,0 +1,44 @@
+<%= page_title 'hub_variant_c.select_documents.title' %>
+<% content_for :feedback_source, 'SELECT_DOCUMENTS_PAGE' %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= form_for @form, url: select_documents_submit_path, html: {id: 'validate-select-documents', class: 'select-documents-form', novalidate: 'novalidate', 'data-msg': t('hub_variant_c.select_documents.errors.no_selection')} do |f| %>
+       <%= content_tag :div, class: form_question_class do %>
+        <fieldset>
```
jakubmiarka: not sure the heading should be part of the fieldset, it diverges from the rest of the views

config/main_routes.rb
```diff
@@ -110,3 +123,82 @@ def add_routes(routes_name)
...
+  get 'select_documents', to: 'select_documents_variant_c#index', as: :select_documents
+
+  #TODO: fix these to go with the vC controller
+  get 'select_documents_none', to: 'select_documents_variant_c#no_documents', as: :select_documents_no_documents
+  post 'select_documents', to: 'select_documents_variant_c#select_documents', as: :select_documents_submit
+  get 'select_documents_advice', to: 'select_documents_variant_c#advice', as: :select_documents_advice
+
+  get 'choose_a_certified_company', to: 'choose_a_certified_company_loa2_variant_c#index', as: :choose_a_certified_company
+  get 'choose_a_certified_company/:company', to: 'choose_a_certified_company_loa2_variant_c#about', as: :choose_a_certified_company_about
+  post 'choose_a_certified_company', to: 'choose_a_certified_company_loa_variant_c2#select_idp', as: :choose_a_certified_company_submit
```
jakubmiarka:
```suggestion
  post 'choose_a_certified_company', to: 'choose_a_certified_company_loa2_variant_c#select_idp', as: :choose_a_certified_company_submit
```
the `2` is in a wrong place :D

app/models/select_documents_variant_c_form.rb
```diff
@@ -0,0 +1,52 @@
...
+  attr_accessor :has_valid_passport, :has_driving_license, :has_phone_can_app, :has_credit_card, :has_nothing
+  validate :must_choose_something
+
+  def self.from_post(hash)
+    instance = self.new
+    instance.has_nothing = hash[:has_nothing] == 't'
+    instance.has_valid_passport = @has_nothing ? false : hash[:has_valid_passport] == 't'
+    instance.has_driving_license = @has_nothing ? false :  hash[:has_driving_license] == 't'
+    instance.has_phone_can_app = @has_nothing ? false : hash[:has_phone_can_app] == 't'
+    instance.has_credit_card = @has_nothing ? false : hash[:has_credit_card] == 't'
```
jakubmiarka: if you use standard `check_box` with a `multiple` option set to true or use `check_box_tag` you could simplify this significantly, as you would then receive only the values which are checked... currently you get all the fields passed in

app/views/select_documents_variant_c/advice.html.erb
```diff
@@ -0,0 +1,30 @@
...
+      <%= raw t 'hub_variant_c.select_documents_advice.advice_html.3-more-things-special-heading' %>
+      <div class="other-ways-to-complete-transaction">
+        <%= raw @other_ways_text %>
+      </div>
+    <% end %>
+
+    <% @advice_codes.each do |code| %>
+      <%= raw t('hub_variant_c.select_documents_advice.advice_html.' + code,
+                try_again_link_start: "<a href='"+ html_escape(url_for(select_documents_path)) + "'>",
+                try_again_link_end: "</a>") %>
```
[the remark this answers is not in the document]
galund: Nope, that won't actually translate properly because the word order may change in some languages. This is the correct way to do translations ;-)

app/views/select_documents_variant_c/index.html.erb
galund: It's weird I agree but this is legit what we are supposed to do in the design system apparently https://design-system.service.gov.uk/components/fieldset/

app/views/select_documents_variant_c/index.html.erb
galund: We did it like this on the assumption that @j-maynard would have to touch it anyway in order to fix all the things ;-(

app/views/select_documents_variant_c/index.html.erb
```diff
@@ -0,0 +1,44 @@
...
+
+          <%= f.custom_check_box :has_valid_passport, { }, "t", "f", t('hub_variant_c.select_documents.documents_option_valid_passport') %>
+          <%= f.custom_check_box :has_driving_license, { }, "t", "f", t('hub_variant_c.select_documents.documents_option_driving_license') %>
+          <%= f.custom_check_box :has_phone_can_app, { }, "t", "f", t('hub_variant_c.select_documents.documents_option_phone_can_app') %>
+          <%= f.custom_check_box :has_credit_card, { }, "t", "f", t('hub_variant_c.select_documents.documents_option_credit_card') %>
+          <%= f.custom_check_box :has_nothing, { }, "t", "f", t('hub_variant_c.select_documents.documents_option_nothing') %>
+        </fieldset>
+      <% end %>
+
+      <details class="govuk-details" style="margin-bottom: 20px;">
```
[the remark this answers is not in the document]
galund: Same answer. We know the design system is coming and wanted to save @j-maynard from having to put it back in again. The inline styles are essential at least until the design system comes in, and then maybe we can get rid again.

app/views/select_documents_variant_c/advice.html.erb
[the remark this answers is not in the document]
jakubmiarka: how come? you have a text where you're just inserting a link.. it's up to you where you place it in the corresponding language... plus this has no translation anyway...

config/main_routes.rb
galund: :-o I'd sweat this worked

app/views/select_documents_variant_c/index.html.erb
jakubmiarka: yeah, but we're not using design system, so it doesn't apply any design system logic... the fieldset guidance doesn't take the full paragraph of text into consideration though... but 🤷‍♂

app/views/select_documents_variant_c/advice.html.erb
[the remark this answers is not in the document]
galund: Well so if you have two separate translation strings, then the translator doesn't get given any context for the link text, and they will probably translate it wrong. (Once had to get a site translated into Chinese; learnt the hard way.)

app/views/select_documents_variant_c/index.html.erb
galund: I thought Jamie was changing it so that we were using the design system, maybe I misunderstood what he was up to :-)

app/views/select_documents_variant_c/index.html.erb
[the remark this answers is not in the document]
jakubmiarka: the inline styles will make it more difficult for Jamie I'm afraid, as it won't be just replacing the classes but will have to check the styles attributes in addition to classes... inline styles are evil in general

app/views/select_documents_variant_c/advice.html.erb
```diff
@@ -0,0 +1,30 @@
+<%= page_title 'hub_variant_c.select_documents.title' %>
+<% content_for :feedback_source, 'SELECT_DOCUMENTS_PAGE' %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <!-- short codes: <%= @advice_codes %> -->
+
+    <!-- special case -->
+    <% if @advice_codes.include?('3-more-things-and-req-photo-id') %>
+      <%= raw t 'hub_variant_c.select_documents_advice.advice_html.3-more-things-special-heading' %>
```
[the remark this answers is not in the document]
jakubmiarka: if you append `_html` to the last node name, you don't need to raw it and it will render properly as html

app/views/select_documents_variant_c/index.html.erb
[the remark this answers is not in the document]
galund: He's gonna have to check the layout anyway. I think this is excusable in the context, and I'm surprised if it'd take longer this way, but perhaps @instantiator and @j-maynard should talk and come up with a plan.

app/views/select_documents_variant_c/advice.html.erb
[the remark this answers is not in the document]
jakubmiarka: but you don't have two separate strings, you have one with a placeholder for the the link...
```
something:
    paragraph: This is my paragraph and you can %{href} for more.
    link: click here
```
then you order doesn't matter at all, as each language can have it whatever way it needs

app/views/select_documents_variant_c/index.html.erb
jakubmiarka: yep, we're, but we're not using the heading in the fieldset for other question... just something to remember when going through the updates...

app/assets/javascripts/application.js.erb
```diff
@@ -19,6 +19,7 @@ window.GOVUK.shimLinksWithButtonRole.init();
 
 window.GOVUK.validation.init();
 window.GOVUK.selectDocuments.init();
+window.GOVUK.selectDocumentsVariantC.init();
```
jakubmiarka: this should have been added as per the commented example below using the `abTest()` function... from a quick look at the new JS, I think it shouldn't cause any major issues being used like this...

app/controllers/choose_a_certified_company_loa2_variant_c_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    else
+      redirect_to select_documents_advice_path
+    end
+  end
+
+  def select_idp
+    if params[:entity_id].present?
+      select_viewable_idp_for_loa(params.fetch('entity_id')) do |decorated_idp|
+        session[:selected_idp_was_recommended] = recommendation_engine.recommended?(decorated_idp.identity_provider, current_identity_providers_for_loa_by_variant('c'), selected_evidence, current_transaction_simple_id)
+        # TODO - do the spinny thing page
```
jakubmiarka: (the spinny thing page is on the matching page, not on select idp)

PR #767, opened 2019-12-16
HUH-74: Interstitial page sign-in hint: Set session and redirect to IDP

Wynndow:
We're testing a new flow where a journey hint is set on the interstitial
page if it is determined the user has previously signed in successfully.
The hint is the simple_id of the IDP, prefixed with idp_.

These changes allow that hint to be picked up, and the idp selected,
before redirecting the user to the idp.

### DIFF
--- a/app/controllers/authn_request_controller.rb
+++ b/app/controllers/authn_request_controller.rb
@@ -100,6 +100,9 @@ private
         do_eidas_sign_in_redirect(redirect_params)
       when 'submission_confirmation'
         redirect_to confirm_your_identity_path(redirect_params)
+      when successful_idp_hint?
+        flash[:journey_hint] = hint
+        redirect_to redirect_to_idp_sign_in_with_last_successful_idp_path(redirect_params)
       else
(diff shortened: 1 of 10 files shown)


app/controllers/authn_request_controller.rb
```diff
@@ -128,4 +131,8 @@ def store_rp_request_data
   def raise_error_if_params_invalid
     something_went_wrong_warn("Missing/empty SAML message from #{request.referer}", :bad_request) if params['SAMLRequest'].blank?
   end
+
+  def successful_idp_hint?
+    ->(hint) { hint && hint.start_with?('idp_') }
```
CharlesIC: `->(hint)` is this correct Ruby syntax?

Wynndow: It is.

CharlesIC: Hmm that's right... https://www.rubyguides.com/2016/02/ruby-procs-and-lambdas/
I'm so used to lambdas being `x ​-> { f(x) }` that this looks weird to me

Wynndow: Ruby's a weird language dude.

PR #771, opened 2019-12-16
HUH-74 Fix bug if last successful IDP not available

Wynndow:
We've [seen a few errors in Sentry](https://sentry.tools.signin.service.gov.uk/verify/prod-frontend/issues/3791/events/277199/) where the last successful IDP wasn't found in
the list of available IDPs. This will prevent an error being thrown and
log so we can investigate, whilst being invisible to the user.

### DIFF
--- a/app/controllers/hint_controller.rb
+++ b/app/controllers/hint_controller.rb
@@ -39,11 +39,16 @@ class HintController < ApplicationController
         entity_id
       ).first
 
-      json_object = {
-        'found': 'true',
-        'simpleId': idp.simple_id,
-        'displayName': idp.display_name
-      }
+      if idp.nil?
+        logger.info "No IDP found for entity ID #{entity_id} and identity providers #{identity_providers}"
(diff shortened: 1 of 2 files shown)


app/controllers/hint_controller.rb
```diff
@@ -39,11 +39,16 @@ def last_successful_idp
         entity_id
       ).first
 
-      json_object = {
-        'found': 'true',
-        'simpleId': idp.simple_id,
-        'displayName': idp.display_name
-      }
+      if idp.nil?
+        logger.info "No IDP found for entity ID #{entity_id} and identity providers #{identity_providers}"
```
jakubmiarka: not sure `identity_providers` will be readable in the log, as it might be an array of objects? 🤷‍♂

Wynndow: I tried it locally and it nicely printed the array?

jakubmiarka: ok then!

PR #772, opened 2019-12-18
HUH-253 Cookie Compliance - disable GA [DO NOT MERGE]

JohnWattsGDS:
Disable GA tracking until a consent system is implemented.

Set render GA components to false.

Skip existing tests that will now fail because it has been disabled.  Deliberately have tests as skipped instead of deleting or re-purposing.

Implement tests that ensure GA doesn't get re-enabled by accident.

Co-authored by: Lewis Westbury <lewis.westbury@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/controllers/partials/analytics_partial_controller.rb
+++ b/app/controllers/partials/analytics_partial_controller.rb
@@ -8,7 +8,7 @@ module AnalyticsPartialController
   end
 
   def render_cross_gov_ga
-    @render_cross_gov_ga = true
+    @render_cross_gov_ga = false
   end
 
   def report_to_analytics(action_name)
--- a/spec/features/user_sends_authn_request_spec.rb
+++ b/spec/features/user_sends_authn_request_spec.rb
@@ -173,7 +173,8 @@ describe 'user sends authn requests' do
   end
 
   context 'and request contains cross-domain GA code' do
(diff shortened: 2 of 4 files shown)


spec/features/user_visits_further_information_page_spec.rb
```diff
@@ -197,6 +197,7 @@
     end
 
     it 'will show up automatically if the timeout limit reaches 5 mins' do
+      skip "Google Analytics disabled until consent mechanism in place"
```
jakubmiarka: why does it have to be skipped on this test? it's not GA-related

JohnWattsGDS: So, I ran the tests, found the ones that failed, and pasted in the skip line.  Except I ctrl-c'd the tests to stop them once the ones that had been failing had failed so that the test that was running when I sent the interrupt was also in the list of failures and I wasn't paying enough attention when skipping tests.

jakubmiarka: so in other words... laziness :D

JohnWattsGDS: lol - I'd go for over-enthusiastic ;)

JohnWattsGDS:
At ten days short of a year old I think this can be put to bed.

PR #774, opened 2019-12-18
HUH-253 Disable Piwik (Matomo) cookies [DO NOT MERGE]

JohnWattsGDS:
Temporary measures to suspend Matomo cookie activity whilst a consent mechanism is put into place.

Push value on to piwikAnalyticsQueue to disable cookies.

Remove code that deposits cookies.  Amend tests so that they aren't broken.

### DIFF
--- a/app/assets/javascripts/piwik_event_tracking.js
+++ b/app/assets/javascripts/piwik_event_tracking.js
@@ -26,6 +26,8 @@
 
     global.GOVUK.piwikEventsTracking = {
         init: function() {
+            _paq.push(['disableCookies']);
+
             var piwikEvents = {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/piwik_event_tracking.js
```diff
@@ -26,6 +26,8 @@
 
     global.GOVUK.piwikEventsTracking = {
         init: function() {
+            _paq.push(['disableCookies']);
```
Wynndow: Codacy is upset as `'_paq' is not defined. (no-undef)`

My JS is not good, but it looks like it's correct?

PR #777, opened 2019-12-19
HUH-254 Delete cookies without consent

JohnWattsGDS:
This is branched from HUH-254_cookie_compliance_matomo_edition as there's no point in deleting cookies if we're just going to set them again.  If we delete cookies then only this one needs merging.

This deletes the PIWIK_USER_ID cookie if it's found and also any cookies that start _pk as they're Piwik cookies.  I may be necessary to be more selective about the cookies which are deleted once a consent mechanism is in place as we don't want to be deleting cookies that have been consented to.  If the cookies are named differently that's not a problem but I don't think we get to choose how piwik names its cookies.

### DIFF
--- a/app/controllers/partials/cookie_deletion_partial_controller.rb
+++ b/app/controllers/partials/cookie_deletion_partial_controller.rb
@@ -1,0 +1,6 @@
+module CookieDeletionPartialController
+  def delete_cookies_without_consent
+    cookies.delete CookieNames::PIWIK_USER_ID
+
+  end
+end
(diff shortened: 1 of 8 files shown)


app/controllers/partials/cookie_deletion_partial_controller.rb
```diff
@@ -0,0 +1,6 @@
+module CookieDeletionPartialController
```
JohnWattsGDS: This should be deleted.  It's not used.

PR #788, opened 2020-01-29
EID-1774: Implement new escape route for eIDAS users whose country isn't listed

AmritSidhu:
continuing from @instantiator 's work... 
Users may choose to leave the eIDAS choose-a-country view because their country is not listed.

We have designed a new other ways page for them which provides slightly more relevant information, as some users may still be able to use Verify.

- Adds a new after_eidas action for the OtherWaysToAccessServiceController
- Adds a new route: other_ways_after_eidas
- Allows the user to enter a Verify journey directly from other_ways_after_eidas
- Adds local resource to define the new route, and new text for other_ways_after_eidas
- Pending Welsh wording for the new resources - so currently provides English wording in both en.yml and cy.yml

### DIFF
--- a/app/views/other_ways_to_access_service/after_eidas.html.erb
+++ b/app/views/other_ways_to_access_service/after_eidas.html.erb
@@ -16,7 +16,7 @@
 
     <%= button_link_to t('hub.other_ways_after_eidas.verify.button'), start_path, id: 'next-button' %>
 
-    <h2 class="heading-medium"><%= t 'hub.other_ways_heading', other_ways_description: @other_ways_description %></h2>
+    <h2 class="heading-medium"><%= t 'hub.other_ways_after_eidas.cannot_verify', other_ways_description: @other_ways_description %></h2>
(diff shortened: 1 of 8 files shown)


AmritSidhu:
Heroku link: https://verify-front-eid-1774-i-z4usbx.herokuapp.com/

app/views/other_ways_to_access_service/after_eidas.html.erb
```diff
@@ -0,0 +1,25 @@
+<%= page_title 'hub.other_ways_after_eidas.title' %>
+<% content_for :feedback_source, 'OTHER_WAYS_AFTER_EIDAS_PAGE' %>
```
jakubmiarka: don't forget to add it to the list of feedback sources:
https://github.com/alphagov/verify-frontend/blob/6afedffcb506bb767a70171f30a67882e2485f4c/app/models/feedback_source_mapper.rb#L3

PR #793, opened 2020-02-05
BUGFIX Undo change to digest config

j-maynard:
Now we've confirmed some thoughts on this we need to undo this change.

### DIFF
--- a/Dockerfile.heroku
+++ b/Dockerfile.heroku
@@ -29,5 +29,6 @@ RUN bash -c 'source .env \
     && ln -s /verify-frontend/public/assets /assets
 
 # Puma needs these dockerignored dirs to write to
+RUN bundle exec rake assets:precompile assets:undigest_assets
 RUN mkdir -p log tmp
 
--- a/Dockerfile.heroku
+++ b/Dockerfile.heroku
@@ -29,6 +29,5 @@ RUN bash -c 'source .env \
     && ln -s /verify-frontend/public/assets /assets
 
 # Puma needs these dockerignored dirs to write to
-RUN bundle exec rake assets:precompile assets:undigest_assets
 RUN mkdir -p log tmp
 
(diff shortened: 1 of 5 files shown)


Dockerfile.heroku
```diff
@@ -25,9 +25,12 @@ RUN echo 'heroku' > /verify-frontend/.build-number
 
 RUN bash -c 'source .env \
              && export $(cut -d= -f1 .env) \
-             && bundle exec rake assets:precompile' \
+             && bundle exec rake assets:precompile \
+             && bundle exec rake assets:copy_govuk_dependencies \
```
jakubmiarka: `copy_govuk_dependencies` already happens in the `assets:precompile` so you're doing this twice...

lib/tasks/copy_govuk_dependencies.rake
```diff
@@ -1,11 +1,34 @@
...
-  source = 'lib/node_modules/govuk-frontend/govuk/assets/'
-  destination = 'public/assets/govuk-frontend/govuk/assets/'
-
-  dirname = File.dirname(destination)
-  FileUtils.mkdir_p(dirname)
-  FileUtils.copy_entry source, destination
+desc 'Copy and undigest govuk-frontend assets'
+
+namespace :assets do
+  task 'copy_govuk_dependencies' do
```
jakubmiarka: I would suggest hooking this to the precompile task and you don't have mess around with dockerfiles and stuff...
see https://coderwall.com/p/qhdhgw/adding-a-post-execution-hook-to-the-rails-db-migrate-task

lib/tasks/copy_govuk_dependencies.rake
```diff
@@ -1,11 +1,39 @@
...
-task 'copy_govuk_dependencies' do
-  source = 'lib/node_modules/govuk-frontend/govuk/assets/'
-  destination = 'public/assets/govuk-frontend/govuk/assets/'
-
-  dirname = File.dirname(destination)
-  FileUtils.mkdir_p(dirname)
-  FileUtils.copy_entry source, destination
+desc 'Copy and undigest govuk-frontend assets'
+
+namespace :assets do
```
jakubmiarka: this probably should be indented

PR #797, opened 2020-02-07
BUGFIX Fixes unsigned cookie script in FireFox

j-maynard:
This fixes the blocking of the cookie script in FireFox

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -43,7 +43,7 @@
                   <%= link_to t('cookie_message.link'), cookies_path %>
                 </p>
             </div>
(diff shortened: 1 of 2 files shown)


j-maynard:
@galund I think we should maybe do a script audit do we can match scripts with SHA's at some point in the future.  When I started working on this there was no list of which SHA matches what so I guess we carry on that pattern until we decided to change it.

galund:
> @galund I think we should maybe do a script audit do we can match scripts with SHA's at some point in the future. When I started working on this there was no list of which SHA matches what so I guess we carry on that pattern until we decided to change it.

Can you create a ticket in the backlog please, ta

PR #799, opened 2020-02-11
ZEN-3915938 Fix for broken interstital question

j-maynard:
This a quick bug for for a bug fix noticed by barclays and reported
to us via ZenDesk


Wynndow:
What was the bug?

PR #800, opened 2020-02-13
BUGFIX for missing styles on failed registration

j-maynard:
@jakubmiarka reported the following bug in the verify frontend.  This should hopefully fix the issue:

image

### DIFF
--- a/app/views/failed_registration/_custom_failed_registration.html.erb
+++ b/app/views/failed_registration/_custom_failed_registration.html.erb
@@ -1,6 +1,6 @@
 <h1 class="govuk-heading-l"><%= raw transaction.custom_fail_heading  %></h1>
 
-<p><%= raw(transaction.custom_fail_what_next_content % { idp_name: idp.display_name }) %></p>
+<p class="govuk-body"><%= raw(transaction.custom_fail_what_next_content % { idp_name: idp.display_name }) %></p>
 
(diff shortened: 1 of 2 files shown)


app/views/failed_registration/_custom_failed_registration.html.erb
```diff
@@ -1,6 +1,6 @@
 <h1 class="govuk-heading-l"><%= raw transaction.custom_fail_heading  %></h1>
 
-<p><%= raw(transaction.custom_fail_what_next_content % { idp_name: idp.display_name }) %></p>
+<p class="govuk-body"><%= raw(transaction.custom_fail_what_next_content % { idp_name: idp.display_name }) %></p>
```
jakubmiarka: you should also update it here to fix the whole page: https://github.com/alphagov/verify-frontend-federation-config/blob/2fcf4638d469cfc93a7b50b97854663266ae67b2/configuration/display-locales/idps/digidentity.yml#L18 etc

jakubmiarka: i.e. the "Digidentity helpline" text needs to be fixed there (and other IDPs)

j-maynard: I'll do that now... Thank you for pointing it out.

PR #803, opened 2020-02-26
HUB-422: Fix backend validation models

Wynndow:
This PR removes the use of the base element from the form models used in the Verify frontend, and instead applies validation messages to the individual element themselves.

This has required creating some new content for new validation messages. This will need to be checked by someone with content skills.

It also moves the validation summary above the H1 on the relevant pages as per the design system recommendations.

As well as this, the links in the summary will now take you either to the relevant text input field, or in the case of radio buttons, the first one in the group.

**Note**
These changes are only to the server side rendering of the validation messages. To view then you will need to turn your JS off. When JS is on it controls the validation messages.

Also note, a lot of these forms are going away soon due to the short hub changes due to be implemented by the end of March. As such, we've decided to not get Welsh translations. Except for the "will it work for me" page - we've requested translations for the updated content.

Example screenshots:
| Before  | After |
| ------------- | ------------- |
| www signin service gov uk_start(Laptop with HiDPI screen)  | localhost_50300_start(Laptop with HiDPI screen)  |
| www signin service gov uk_will-it-work-for-me(Laptop with HiDPI screen) (1)  | localhost_50300_will-it-work-for-me(Laptop with HiDPI screen) (1)  |

### DIFF
--- a/app/assets/stylesheets/pages/_slides.scss
+++ b/app/assets/stylesheets/pages/_slides.scss
@@ -129,6 +129,19 @@
   .govuk-details__text {
     border-left: 5px solid govuk-colour("white");
   }
+
+  .govuk-error-summary {
+    border: 5px solid govuk-colour("white");
+    color: govuk-colour("white");
+  }
+
+  .govuk-error-message {
+    color: govuk-colour("white");
+  }
+
+  .govuk-error-summary__list a {
+    color: govuk-colour("white");
+  }
 }
 
 .hide-mobile {
--- a/app/models/cycle_three/cycle_three_attribute.rb
+++ b/app/models/cycle_three/cycle_three_attribute.rb
@@ -48,7 +48,7 @@ module CycleThree
 
     def matches_regex
       unless pattern.match?(@cycle_three_data)
-        errors.add(:cycle_three_data, format(I18n.t('hub.further_information.attribute_validation_message'), cycle_three_name: name))
+        errors.add(:cycle_three_data, I18n.t('hub.further_information.attribute_validation_message', cycle_three_name: name))
       end
     end
   end
--- a/app/models/cycle_three/cycle_three_attribute.rb
+++ b/app/models/cycle_three/cycle_three_attribute.rb
@@ -48,10 +48,7 @@ module CycleThree
 
     def matches_regex
       unless pattern.match?(@cycle_three_data)
-        errors.add(
-          :cycle_three_data,
-          'hub.further_information.attribute_validation_message'
-        )
+        errors.add(:cycle_three_data, format(I18n.t('hub.further_information.attribute_validation_message'), cycle_three_name: name))
       end
     end
   end
--- a/app/models/select_documents_form.rb
+++ b/app/models/select_documents_form.rb
@@ -2,7 +2,7 @@ class SelectDocumentsForm
   include ActiveModel::Model
 
   attr_reader :driving_licence, :ni_driving_licence, :passport, :any_driving_licence
-  validate :any_driving_licence_and_passport_must_be_present
+  validate :any_driving_license_must_be_present, :passport_must_be_present
   validate :driving_licence_details_present
 
   def initialize(hash)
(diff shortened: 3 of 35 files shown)


spec/models/cycle_three/cycle_three_attribute_spec.rb
```diff
@@ -35,7 +39,7 @@ module CycleThree
       form = form_class.new(cycle_three_data: '123')
 
       expect(form).to_not be_valid
-      expect(form.errors.full_messages).to eql ['Cycle three data hub.further_information.attribute_validation_message']
+      expect(form.errors.full_messages).to eql ['Cycle three data Enter a valid Sausage']
```
JohnWattsGDS: 🈂️ 🇺🇸 🅰️ G 🇪🇸

app/models/cycle_three/cycle_three_attribute.rb
```diff
@@ -48,10 +48,7 @@ def simple_id
 
     def matches_regex
       unless pattern.match?(@cycle_three_data)
-        errors.add(
-          :cycle_three_data,
-          'hub.further_information.attribute_validation_message'
-        )
+        errors.add(:cycle_three_data, format(I18n.t('hub.further_information.attribute_validation_message'), cycle_three_name: name))
```
galund: I'm confused whether the second argument to this thing is a translated string, or a key to pass to the translation engine. We presumably haven't actually changed that?

app/views/shared/_form-errors.html.erb
```diff
@@ -1,5 +1,8 @@
 <% if errors %>
   <div  class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      There is a problem
```
galund: Can we stick this in as a translation string, even if we aren't going to get it translated yet?

app/assets/stylesheets/pages/_slides.scss
```diff
@@ -129,6 +129,19 @@
   .govuk-details__text {
     border-left: 5px solid govuk-colour("white");
   }
+
+  .govuk-error-summary {
+    border: 5px solid govuk-colour("white");
+    color: govuk-colour("white");
```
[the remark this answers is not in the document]
Wynndow: My understanding of scss is low/nothing, but will these styles not inherit the `background-color` of the parent style their nested in, `slides`, up on line 6?

app/models/cycle_three/cycle_three_attribute.rb
Wynndow: It's a key/value pair for string formatting, as the translation string is `Enter a valid %{cycle_three_name}`

I think I may have over-egged it though and suspect `t()` can do the formatting for me, without the need for `format()`. I'll see if I can reduce it down.

Previously we were just adding a literal string for the location of the translation string to the errors. This was never actually being rendered in the page so wasn't an issue. Now it's being rendered in the summary, so need to template it before adding it to the errors.

app/views/shared/_form-errors.html.erb
Wynndow: Ah yeah - that's a good point.

app/assets/stylesheets/pages/_slides.scss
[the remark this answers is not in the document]
galund: If it's SCSS that's going to make the `background` get combined in, then sure that's fine. If a CSS style ends up in the browser with a `color` but no `background` then technically that can break accessibility depending on how the cascade with user stylesheets / !important options happens. Can't find a reference for that now, and maybe it's something nobody worries about these days.

app/models/select_documents_form.rb
```diff
@@ -36,23 +36,19 @@ def further_id_information_required?
...
 
   def driving_licence_details_present
-    if any_driving_licence == 'true' && no_driving_licence_details?
-      add_documents_error
-    end
+    errors.add(:driving_licence_great_britain, I18n.t('hub.select_documents.errors.no_driving_licence_issuer_selection')) if any_driving_licence == 'true' && no_driving_licence_details?
   end
 
   def no_driving_licence_details?
     !(driving_licence == 'great_britain' || driving_licence == 'northern_ireland')
```
jkunle: I wish this was 
`
driving_licence != 'great_britain' && driving_licence != 'northern_ireland'
`
instead you don't have to change this

galund: happily this code will go away (assuming the AB test goes well in April)

the new version is also weird in places, but doesn't have this bit of logic anyway :-) https://github.com/alphagov/verify-frontend/blob/master/app/models/select_documents_variant_c_form.rb

app/views/redirect_to_idp_question/redirect_to_idp_question_LOA1.html.erb
```diff
@@ -3,18 +3,24 @@
...
+
     <h1 class="govuk-heading-l"><%= t 'hub.redirect_to_idp_question.loa1_heading', display_name: @idp.display_name %></h1>
     <%= raw @idp.interstitial_question %>
 
-    <%= render 'shared/form-errors', errors: flash[:errors], form_name: 'interstitial-question-form' %>
-    
     <%= form_for @form, url: redirect_to_idp_question_submit_path, html: {id: 'interstitial-question-form', class: 'js-validate', novalidate: 'novalidate'} do |f| %>
         <%= content_tag :div, class: form_question_class do %>
-        <div class="govuk-form-group">
+        <div class="govuk-form-group <% if @form.errors.include?(:interstitial_question_result_true) %>govuk-form-group--error<% end %>">
```
jkunle: This is fine and is ok (do have to change), but I think this is better 
```
 <div class="govuk-form-group<% @form.errors.include?(:interstitial_question_result_true) ? 'govuk-form-group--error' : '' %>">
```

jkunle: There other instances of these but, since this work it is ok

PR #810, opened 2020-03-23
HUB-235: Add additional tracking to short hub variant c abc test

jkunle:
see https://govukverify.atlassian.net/browse/HUB-235
**What**

We need to measure a bunch of things as part of the short hub ABC test. Some of those things will require additional tracking.

The user will be asked to select what evidence they have, and may return to that page for several attempts. For example, if the user does not select enough evidence the first time, they get to an ‘other ways’ page, but may use their back button / click “try again” and have another attempt, in which they confess to having a passport.

We think we want to be able to track (in priority order):

evidence selection on submission of the new evidence checkbox page (/select-documents) - what combinations of evidence do users select on each attempt.  (If each combination of ticked evidence becomes a virtual page title e.g. maybe EVIDENCE_CREDIT_CARD_PASSPORT, and with each virtual page view we record a custom variable indicating the “attempt” e.g. ATTEMPT_1, we think we can get useful figures showing how this new page is used.)

use of the progressive disclosure link 'What GOV.UK Verify uses these for' on the new evidence checkbox page (/select-documents)  – again a virtual page view for this.

**Why**

We need to know if users select more evidence when prompted to do so, on subsequent attempts.

### DIFF
--- a/app/controllers/select_documents_variant_c_controller.rb
+++ b/app/controllers/select_documents_variant_c_controller.rb
@@ -45,7 +45,6 @@ private
   end
 
   def increase_attempt_number
-    session[:evidence_attempt_number] = 0 if session[:evidence_attempt_number].nil?
-    session[:evidence_attempt_number] = session[:evidence_attempt_number] + 1
+    session[:evidence_attempt_number] = (session[:evidence_attempt_number] || 0) + 1
   end
 end
--- a/app/controllers/select_documents_variant_c_controller.rb
+++ b/app/controllers/select_documents_variant_c_controller.rb
@@ -15,6 +15,7 @@ class SelectDocumentsVariantCController < ApplicationController
     if @form.valid?
       selected_answer_store.store_selected_answers("documents", @form.to_session_storage)
       idps_available = IDP_RECOMMENDATION_ENGINE_variant_c.any?(current_identity_providers_for_loa_by_variant("c"), selected_evidence, current_transaction_simple_id)
+      report_user_evidence_to_piwik(selected_evidence)
       redirect_to idps_available ? choose_a_certified_company_path : select_documents_advice_path
     else
       flash.now[:errors] = @form.errors.full_messages.join(", ")
@@ -24,12 +25,27 @@ class SelectDocumentsVariantCController < ApplicationController
 
   def advice
     suggestions = IDP_RECOMMENDATION_ENGINE_variant_c.get_suggested_idps_for_registration(current_identity_providers_for_loa_by_variant("c"), selected_evidence, current_transaction_simple_id)
-
     @evidence = selected_evidence
     @advice_codes = segment_advice(suggestions[:user_segments])
     @other_ways_description = current_transaction.other_ways_description
     @other_ways_text = current_transaction.other_ways_text
 
     render :advice
+  end
+
+private
+
+  def report_user_evidence_to_piwik(selected_evidence)
+    FEDERATION_REPORTER.report_user_evidence_attempt(
+      current_transaction: current_transaction,
+      request: request,
+      attempt_number: increase_attempt_number,
+      evidence_list: selected_evidence,
+    )
+  end
+
+  def increase_attempt_number
+    session[:evidence_attempt_number] = 0 if session[:evidence_attempt_number].nil?
+    session[:evidence_attempt_number] = session[:evidence_attempt_number] + 1
   end
 end
--- a/app/models/analytics/federation_reporter.rb
+++ b/app/models/analytics/federation_reporter.rb
@@ -151,6 +151,17 @@ module Analytics
       )
     end
 
+    def report_user_evidence_attempt(attempt_number:, current_transaction:, request:, evidence_list: [])
+      list_of_evidence = evidence_list.map { |evidence| evidence.to_s.gsub("has_", "") }
+                                      .sort.join("_")
+                                      .upcase
+      report_action(
+        current_transaction,
+        request,
+        "EVIDENCE_ATTEMPT_#{attempt_number} | #{list_of_evidence} |",
+      )
+    end
+
     def report_number_of_idps_recommended(current_transaction, request, number_of_idps_recommended)
       report_event(
         current_transaction,
@@ -201,15 +212,13 @@ module Analytics
     end
 
     def report_action(current_transaction, request, action, extra_custom_vars = {})
-      begin
-        @analytics_reporter.report_action(
-          request,
-          action,
-          universal_custom_variables(current_transaction, request).merge(extra_custom_vars),
-        )
-      rescue Display::FederationTranslator::TranslationError => e
-        Rails.logger.warn e
-      end
+      @analytics_reporter.report_action(
+        request,
+        action,
+        universal_custom_variables(current_transaction, request).merge(extra_custom_vars),
+      )
+    rescue Display::FederationTranslator::TranslationError => e
+      Rails.logger.warn e
     end
 
     def report_action_without_current_transaction(request, action, extra_custom_vars = {})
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/select_documents_variantc.js
```diff
@@ -0,0 +1,14 @@
+(function (global) {
+  "use strict";
+
+  global.GOVUK.selectDocumentsVariantC = {
+      init: function () {
+        $('#progressive_disclosure').on('toggle', function(event) {
+            if (event.target.open) {
+              _paq.push(['trackEvent', 'Engagement', 'yes', 'progressive_disclosure']);
```
galund, line 8: What does the 'yes' relate to? - maybe that deserves a code comment?

app/controllers/select_documents_variant_c_controller.rb
```diff
@@ -15,6 +15,8 @@ def select_documents
     if @form.valid?
       selected_answer_store.store_selected_answers("documents", @form.to_session_storage)
       idps_available = IDP_RECOMMENDATION_ENGINE_variant_c.any?(current_identity_providers_for_loa_by_variant("c"), selected_evidence, current_transaction_simple_id)
+      FEDERATION_REPORTER.report_number_of_evidence_attempts(current_transaction, request, selected_evidence.length)
```
galund, line 18: Your third argument here is supposed to be `number_of_attempts` but you seem to be reporting the number of pieces of evidence that have been selected?

app/models/analytics/federation_reporter.rb
```diff
@@ -151,6 +151,27 @@ def report_cycle_three_cancel(current_transaction, request)
...
+                                      .sort.join("_")
+                                      .upcase
+      report_action(
+        current_transaction,
+        request,
+        "EVIDENCE_ATTEMPT_#{attempt_number} | #{list_of_evidence}",
+      )
+    end
+
+    def report_number_of_evidence_attempts(current_transaction, request, number_of_attempts)
```
galund, line 165: I'm not convinced this needs to be a separate function at all, couldn't it be done by the `report_user_evidence_attempt` function above?

app/assets/javascripts/select_documents_variantc.js
jkunle: I'll put a comment it just means the the disclosure was expanded by the users

app/controllers/select_documents_variant_c_controller.rb
jkunle: That was my intent, I thought it could provide complimentary insight (using events) that's easier to dice when viewed through matomo Behaviour -> Events. We would be able to see how many users are selecting just 1, 2, 3.  What do you think.

jkunle: `report_user_evidence_attempt_to_piwik(selected_evidence)` does combinations of evidence do users select on each attempt and implements as virtual page view

app/models/analytics/federation_reporter.rb
jkunle: I didn't name this properly `number_of_attempts` should be `number_of_evidence_selected`

galund: Aha! I understand now.. the word you're missing is "quantity" instead of "number" :-)

spec/javascripts/select_documents_variant_c_spec.js
```diff
@@ -0,0 +1,48 @@
...
+  function expectLatestEntryInPiwikQueueToMatch(category, action, name) {
+      var latestEntryInPiwikQueue = _paq.push.calls.mostRecent().args[0];
+      expect(latestEntryInPiwikQueue[0]).toBe('trackEvent');
+      expect(latestEntryInPiwikQueue[1]).toBe(category);
+      expect(latestEntryInPiwikQueue[2]).toBe(action);
+      expect(latestEntryInPiwikQueue[3]).toBe(name);
+  };
+
+  describe('page with disclosure', function() {
+    var htmlWithMarkup = '<details id="progressive_disclosure" class="govuk-details govuk-!-padding-top-6" data-module="govuk-details">' +
```
galund, line 16: I don't think we should be reproducing the entire thing into our tests. It seems like we're not really testing the actual code itself if we do that?

spec/javascripts/select_documents_variant_c_spec.js
```diff
@@ -0,0 +1,48 @@
...
+                            'They cannot see your transactions and cannot charge your account.' +
+                            '</p>' +
+                          '</div>' +
+                        '</details>';
+
+      beforeEach(function () {
+          setUp(htmlWithMarkup);
+      });
+
+      it('should report to Piwik when evidence is selected', function () {
```
galund, line 41: You're only testing the progressive disclosure, not the selecting of the evidence. So I think this is mis-named.

spec/javascripts/select_documents_variant_c_spec.js
jkunle: I'm open to thoughts on this, the idea here is to have html fragment that resembles the real thing and when selecting the `#progressive_disclosure` the expected piwik event properties have been populated. It is very similar to what has been done with other tests here anyway.

galund: Which code are we purporting to test here?

jkunle: Want to test that when the Disclosure link is selected the event below is called as expected
```
track_disclosure: function(event) {
        // user selects progressive disclosure link track user opens the link
            return _paq.push(['trackEvent','Engagement','Disclosure link selected', 'Opened']);
}
```

app/controllers/select_documents_variant_c_controller.rb
```diff
@@ -24,12 +25,27 @@ def select_documents
...
+      current_transaction: current_transaction,
+      request: request,
+      attempt_number: increase_attempt_number,
+      evidence_list: selected_evidence,
+    )
+  end
+
+  def increase_attempt_number
+    session[:evidence_attempt_number] = 0 if session[:evidence_attempt_number].nil?
+    session[:evidence_attempt_number] = session[:evidence_attempt_number] + 1
```
galund, line 49: I think this could have been one line i.e.
```
session[:evidence_attempt_number] = (session[:evidence_attempt_number] || 0) + 1
```

spec/controllers/select_documents_variant_c_controller_spec.rb
```diff
@@ -22,6 +23,26 @@
...
+        attempt_number: 1,
+        evidence_list: { device_type_other: true }.merge!(evidence).keys,
+      )
+      post :select_documents, params: { locale: "en", select_documents_variant_c_form: evidence }
+      expect(subject).to redirect_to select_documents_advice_path
+    end
+
+    it "redirects to the advice page when None of the above is selected" do
+      evidence = { has_nothing: "t" }
+      expect(FEDERATION_REPORTER).to receive(:report_user_evidence_attempt)
```
galund, line 39: Seeing as this is basically the same as the above, it'd be nice if it could be a bit more DRY.

spec/features/user_visits_select_documents_variant_c_page_spec.rb
```diff
@@ -24,22 +24,98 @@
...
     click_button t("navigation.continue")
 
+    expect(page).to have_title t("hub_variant_c.choose_a_certified_company.title")
     expect(page).to have_current_path(choose_a_certified_company_path)
   end
 
   it "redirects to the select advice page when selects 2 documents" do
     visit "/select-documents"
+    expect_reporter_to_receive(
+      evidence: %I(has_driving_license has_credit_card device_type_other),
```
galund, line 43: If you aren't using, or expecting to use, interpolated strings, I think you should use `%i` rather than `%I`

spec/javascripts/select_documents_variant_c_spec.js
```diff
@@ -0,0 +1,33 @@
...
+          setUp('<details id="progressive_disclosure" class="govuk-details govuk-!-padding-top-6" data-module="govuk-details">' +
+          '<summary piwik_event_tracking="progressive_disclosure" class="govuk-details__summary">' +
+            '<span class="govuk-details__summary-text">What GOV.UK Verify uses these for</span>' +
+          '</summary>' +
+          '<div class="govuk-details__text">' +
+          '</div>' +
+        '</details>');
+      });
+
+      it('should report to Piwik when disclosure link is selected', function () {
```
galund, line 26: Let's chat about this on a hangout, cos I'm still confused about what it's actually testing. Seems like it might just be testing Matomo code :-)

jkunle: Yes that what it does, the Matomo code wouldn't be called unless the user clicked on the disclosure link.

galund: We chatted, I am mollified (for now anyway :-) )

PR #811, opened 2020-03-23
HUB-545: Rename registration flow methods for more clarity and consistency

CharlesIC:
Rename methods ending in `_for_loa` to `_for_registration`. This is because the `_for_loa` bit doesn't say anything about the purpose of the method and also for sign-in we have `_for_sign_in` everywhere so the naming for the registration should also follow that convention.

Use new IDP list for registration endpoint introduced in alphagov/verify-hub#435

Change the default IDP hiding period to 30 mins

### DIFF
--- a/app/models/config_endpoints.rb
+++ b/app/models/config_endpoints.rb
@@ -1,7 +1,7 @@
 module ConfigEndpoints
   PATH = '/config'.freeze
   PATH_PREFIX = Pathname(PATH)
-  IDP_LIST_SUFFIX = 'idps/idp-list/%<transaction_name>s/%<loa>s'.freeze
+  IDP_LIST_REGISTRATION_SUFFIX = 'idps/idp-list-for-registration/%<transaction_name>s/%<loa>s'.freeze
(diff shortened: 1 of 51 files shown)


app/models/config_endpoints.rb
```diff
@@ -1,16 +1,16 @@
 module ConfigEndpoints
   PATH = '/config'.freeze
   PATH_PREFIX = Pathname(PATH)
-  IDP_LIST_SUFFIX = 'idps/idp-list/%<transaction_name>s/%<loa>s'.freeze
+  IDP_LIST_REGISTRATION_SUFFIX = 'idps/idp-list-for-registration/%<transaction_name>s/%<loa>s'.freeze
```
jakubmiarka: did you make the same change on the config service?

CharlesIC: Yes
https://github.com/alphagov/verify-hub/pull/435/files#diff-9f1b308a39c55558dd4f6c3052a003a0R61

PR #817, opened 2020-03-25
HUB-555: Added a /completed page for IDPs to return the users

jakubmiarka:
If the IDP no longer has the original SAML request for the user, they're not able
to issue a SAML response back to the hub. This new page will allow them to send the
user back. Also the page updates the journey hint to hint the user next time round.

### DIFF
--- a/app/controllers/completed_registration_controller.rb
+++ b/app/controllers/completed_registration_controller.rb
@@ -1,0 +1,30 @@
+require 'partials/user_cookies_partial_controller'
+require 'partials/viewable_idp_partial_controller'
+require 'partials/journey_hinting_partial_controller'
+
+class CompletedRegistrationController < ApplicationController
+  include UserCookiesPartialController
+  include ViewableIdpPartialController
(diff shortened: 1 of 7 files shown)


app/controllers/completed_registration_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  skip_before_action :validate_session
+  skip_before_action :set_piwik_custom_variables
+
+  def index
+    idp_simple_id = params[:idp]
+    @idp = IDP_DISPLAY_REPOSITORY.fetch(idp_simple_id, nil)
+    if @idp.nil?
+      redirect_to verify_services_path
+    else
+      # Hack - we need and IDP list, but the current API needs a transaction
```
CharlesIC: 🤦‍♂

CharlesIC: We should change the API 😬

jakubmiarka: oh definitely :D

PR #819, opened 2020-03-26
HUB-563: Show journey hint for users at the beginning of the journey

jakubmiarka:
Users who have successfully verified with an IDP before, will now be presented with a screen to continue
with that IDP. They can confirm and go straight to the IDP, if they choose not to, they will be presented with the page they were originally visiting (i.e. `/prove-identity` or `/start`). Other hints, such as Pause & Resume still take precedence.

image


app/controllers/prove_identity_controller.rb
```diff
@@ -1,6 +1,29 @@
+require 'partials/journey_hinting_partial_controller'
+require 'partials/viewable_idp_partial_controller'
+require 'partials/user_cookies_partial_controller'
+
 class ProveIdentityController < ApplicationController
+  include JourneyHintingPartialController
+  include ViewableIdpPartialController
+  include UserCookiesPartialController
+
   def index
```
galund: For reasons that I forget, we wanted to have the hint on a separate URL. But it might have been @jakubmiarka that said we should definitely do that, in which case 🤷‍♂ :-D

jakubmiarka: I don't remember saying that :D we were discussing this in the morning with John, comparing pros and cons, and this method came out slightly better...

bethwalton:
Hi, as Si said on Slack can we change "recently" to "previously"? It might not have been recent for the user.

app/controllers/start_variant_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      @identity_provider = retrieve_decorated_singleton_idp_array_by_entity_id(current_available_identity_providers_for_sign_in, journey_hint_entity_id).first
+      if @identity_provider.nil?
+        return render 'start/start'
+      end
+
+      render 'shared/sign_in_hint', layout: 'main_layout'
+    end
+  end
+
+  def ignore_hint
```
CharlesIC: this method is duplicated in `prove_identity_variant_controller`. Can this be refactored to a partial controller or somewhere else?

jakubmiarka: we looked at extracting it, but given this is actually 3 lines, and two of them are calling different partials already it felt slightly as overkill to have another partial which imports two other partials... but if you have strong preference, we could do that

PR #822, opened 2020-03-30
HUB-566: Change how AB test is reported

jakubmiarka:
Piwik/Matomo tracks what variant the user is in during an AB test. This is achieved by setting a custom
variable for AB_TEST. Currently this gets set by reporting a virtual page view
"The user has started an AB test" at the start of the test.

This change stops using the extra call to piwik, but instead adds the custom variable
our existing array (along with LOA and RP). Which gets reported with a standard page view.
This eliminates the race condition between the virtual page view and a standard page view.

It's using flash on the request, as it's not needed to persist across and is only required
when user visits the routes which are being AB-tested.

Additionally, it also simplifies the set up of the AB test, since we no longer need
two set of constraints for piwik reporting and normal routes.

### DIFF
--- a/app/controllers/partials/analytics_partial_controller.rb
+++ b/app/controllers/partials/analytics_partial_controller.rb
@@ -40,6 +40,7 @@ private
       Analytics::CustomVariable.build_for_js_client(:rp, current_transaction.analytics_description),
       Analytics::CustomVariable.build_for_js_client(:loa_requested, session[:requested_loa])
     ]
+    @piwik_custom_variables.push(ab_test_variant) unless ab_test_variant.nil?
   end
 
   def custom_variables_for_img_tracker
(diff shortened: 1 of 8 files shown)


app/controllers/partials/analytics_partial_controller.rb
```diff
@@ -40,13 +40,21 @@ def custom_variables_for_js
...
   def custom_variables_for_img_tracker
     current_transaction_custom_variable = Analytics::CustomVariable.build(:rp, current_transaction.analytics_description)
     loa_requested_custom_variable = Analytics::CustomVariable.build(:loa_requested, session[:requested_loa])
     @piwik_custom_variables_img_tracker =
       current_transaction_custom_variable.merge(loa_requested_custom_variable)
+    @piwik_custom_variables_img_tracker.merge!(ab_test_variant) unless ab_test_variant.nil?
+  end
+
+  def ab_test_variant
+    unless flash[:ab_test_variant].nil?
```
CharlesIC: What's `flash`?

jakubmiarka: it's a way to store for a value for a duration of the request

JohnWattsGDS: It's also the reason a firewall somewhere wouldn't let me download the J2EE7 sdk because 'the file contains flash'. I kid you not.

JohnWattsGDS: Here's the nice explanation from Java's implementation:

_The Flash concept is taken from Ruby on Rails and provides a way to pass temporary objects between the user views generated by the faces lifecycle. As in Rails, anything one places in the flash will be exposed to the next view encountered by the same user session and then cleared out. It is important to note that “next view” may have the same view id as the previous view._

PR #824, opened 2020-03-30
Bau fix session timeout page

j-maynard:
Beth requested some changes to the session timeout page.  These are those changes.

### DIFF
--- a/app/controllers/partials/user_cookies_partial_controller.rb
+++ b/app/controllers/partials/user_cookies_partial_controller.rb
@@ -73,6 +73,10 @@ private
     nil
   end
 
+  def ab_test_variant_value
+    cookies[CookieNames::AB_TEST].value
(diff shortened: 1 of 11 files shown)


bethwalton:
Looks good except I think there are 2 full stops after "Please go to the government service you were trying to access to continue.." ?

app/controllers/partials/user_cookies_partial_controller.rb
```diff
@@ -73,6 +73,10 @@ def journey_hint_value
     nil
   end
 
+  def ab_test_variant_value
```
jakubmiarka: This does no belong to this PR

j-maynard: Let me sort out... its picked up my other branches commits...

PR #829, opened 2020-04-02
HUB-579: Enhance tracking of session timeouts

jakubmiarka:
We'd like to know how long people take to complete and when their session
expire. This adds a prometheus metrics to measure number of timeouts and also
a metric when their session expired (in minutes). Also enhanced the current log message
to include the time as well.

### DIFF
--- a/app/models/session_validator/session_start_time_validator.rb
+++ b/app/models/session_validator/session_start_time_validator.rb
@@ -18,8 +18,8 @@ class SessionValidator
 
     def validate_expiry(session, session_start_time)
       if session_start_time <= @session_duration.minutes.ago
-        session_id = session[:verify_session_id]
-        ValidationFailure.session_expired(session_id)
+        minutes_ago = ((DateTime.now - session_start_time) * 24 * 60).to_i
+        ValidationFailure.session_expired(session, minutes_ago)
       else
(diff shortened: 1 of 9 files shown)


app/models/session_validator/session_start_time_validator.rb
```diff
@@ -18,8 +18,8 @@ def validate(_cookies, session)
 
     def validate_expiry(session, session_start_time)
       if session_start_time <= @session_duration.minutes.ago
-        session_id = session[:verify_session_id]
-        ValidationFailure.session_expired(session_id)
+        minutes_ago = ((DateTime.now - session_start_time) * 24 * 60).to_i
```
galund, line 21: This doesn't look right to me, why 24? It's not starting out measured in days is it? Assuming it's seconds then isn't * 60 enough?

jakubmiarka: it is, DateTime objects when subtracted return the difference in day.

https://ruby-doc.org/stdlib-2.0.0/libdoc/date/rdoc/Date.html#method-i-2D

galund: And DateTime inherits from Date? :mindblown:

PR #832, opened 2020-04-02
BUGFIX: Missing link to the service on session expired page

jakubmiarka:
- the button on the "new" timeout page had a link missing because the instance
variable (`@redirect_to_destination`) did not get set.
- reclassified the HTTP code for session expiry to be
403 Forbidden to be consistent
- render error page when Single IDP cookie is invalid instead of session timeout

### DIFF
--- a/app/controllers/partials/user_errors_partial_controller.rb
+++ b/app/controllers/partials/user_errors_partial_controller.rb
@@ -1,6 +1,13 @@
 module UserErrorsPartialController
   def render_error(partial, status)
     set_locale
+    if partial == :session_timeout
+      @redirect_to_destination = if CONTINUE_ON_FAILED_REGISTRATION_RPS.include?(current_transaction_simple_id)
+                                   '/redirect-to-service/error'
+                                 else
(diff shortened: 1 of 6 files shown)


galund:
I wonder if 403 is useful because it gives us an easy existing way to compare the number of people seeing this page with the other 400's? (i.e. on the front-end grafana dashboard.)

jakubmiarka:
> I wonder if 403 is useful because it gives us an easy existing way to compare the number of people seeing this page with the other 400's? (i.e. on the front-end grafana dashboard.)

that's the thing, it doesn't currently. This page currently can show with either 400 or 403, it depends what triggered it. I can make it 403 if you think that's better.

CharlesIC:
I think 403 would be better as well

jakubmiarka:
@galund @CharlesIC Sorry, it's actually 403 Forbidden for session timeouts, my initial comment was wrong, but the code says 403 (https://github.com/alphagov/verify-frontend/pull/832/files#diff-42505943cadb28446637e2fa8393c573)

PR #838, opened 2020-04-06
HUB-582: Update session_timeout page content

CharlesIC:
This page is now used for all session timeouts and the content needs to reflect that.

Screenshot 2020-04-06 at 18 34 35

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -863,9 +863,15 @@ en:
       feedback_message: Use the %{feedback_link} to ask a question, report a problem or suggest an improvement.
       feedback: feedback
     session_timeout:
-      title: Continue to the government service you need
-      return_to_service: Please go to the government service you were trying to access to continue.
-      start_again: Continue
+      title: Return to the government service you need
+      return_to_service_html: |
+        <p class="govuk-body">Sorry, something went wrong.</p>
+        <p class="govuk-body">You need to try again.</p>
+        <ol class="govuk-list govuk-list--number">
+            <li>Return to the government service you were trying to use.</li>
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -863,8 +863,14 @@ en:
...
-      return_to_service: Please go to the government service you were trying to access to continue.
+      title: Return to the government service you need
+      return_to_service_html: |
+        <p class="govuk-body">Sorry, something went wrong.</p>
+        <p class="govuk-body">You need to try again.</p>
+        <ol class="govuk-list govuk-list--number">
+            <li>Return to the government service you were trying to use.</li>
+            <li>When you’re asked to prove your identity, create a new identity account or sign in with an existing one.</li>
+        </ol>
       start_again: Continue
```
jakubmiarka, line 874: I'm not a content person, but if the bullet point is saying "Return to..." , shouldn't the button say "Return ..." instead of "Continue"?

CharlesIC: One for @bethwalton

bethwalton: Yeah that's a good point, we should more clearly associate the button with the action. 

Can we change it to "Return to the service" ?

CharlesIC: Sure I'll update it 👍

CharlesIC: I was also thinking is it worth including the service action in the try again part? So we could say "You need to try again to [service action]", for instance "You need to try again to apply for Universal Credit". @bethwalton thoughts?

CharlesIC: Button updated ✅

PR #841, opened 2020-04-06
HUB-572: Show all IDPs after registration failure while splitting traffic between IDPs

CharlesIC:
Ignore traffic splitting logic if last status is FAILED
Change the link on the failed page to redirect to the IDP picker page.

Revert changes to the failure page from #821

### DIFF
--- a/app/views/failed_registration/index_LOA2.html.erb
+++ b/app/views/failed_registration/index_LOA2.html.erb
@@ -6,7 +6,7 @@
     <% if @custom_fail %>
         <%= render partial: 'failed_registration/custom_failed_registration', locals: {idp: @idp, transaction: @transaction, start_again_path: select_documents_path} %>
     <% else %>
-        <%= render partial: 'failed_registration/non_continue_rp', locals: {idp: @idp, transaction: @transaction, start_again_path: select_documents_path} %>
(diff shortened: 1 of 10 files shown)


app/views/failed_registration/index_LOA2.html.erb
```diff
@@ -6,7 +6,7 @@
     <% if @custom_fail %>
         <%= render partial: 'failed_registration/custom_failed_registration', locals: {idp: @idp, transaction: @transaction, start_again_path: select_documents_path} %>
     <% else %>
-        <%= render partial: 'failed_registration/non_continue_rp', locals: {idp: @idp, transaction: @transaction, start_again_path: select_documents_path} %>
+        <%= render partial: 'failed_registration/non_continue_rp', locals: {idp: @idp, transaction: @transaction, start_again_path: choose_a_certified_company_path} %>
```
galund, line 9: Perhaps it might be worth putting in logic to say "choose_a_certified_company_path if there's only 2 IDPs, otherwise select_documents_path". However I get this would probably make the tests very complicated. So it might just be worth putting in a comment to the effect that if there was more than 2 IDPs we should revert to the previous logic?

CharlesIC: Hmm yeah I was thinking about it and I concluded that with only 2 IDPs it's okay to just go to the picker. I agree we should have more general logic but I don't expect we'll have more than 2 IDPs any time soon so I took a shortcut. I will add a story to the backlog though as a low priority item

PR #845, opened 2020-04-07
BAU: Move away from govuk-lint

jakubmiarka:
The [govuk-lint](https://github.com/alphagov/govuk-lint) gem has been deprecated and archived. As per the gem's recommendation
switching to ruobocop-govuk and scss_lint-govuk.

This includes a bunch of stylistic updates to align us with GOV.UK styleguide.
I used the autocorrect function of rubocop and although I might not agree with all the styleguides or conventions, it makes it consistent with GOV.UK.

There are two commits, first to change the gems and second to make the stylistic changes.


AmritSidhu:
All these changes look good to me, that was way too many speech marks for my liking...👌🏾I ran locally and its all passing, there are some files in other directories that I assume the linter doesn't look through but should they be changed too or do we not care about those. An example being everything in the stub directory?

jakubmiarka:
> All these changes look good to me👌🏾that was way too many speech marks for my liking...I ran locally and its all passing, there are some files in other directories that I assume the linter doesn't look through but should they be changed too or do we not care about those. An example being everything in the stub directory?

Thanks for reviewing. Yeah, I agree but I guess it had to come at some point. We have never linted the stub folder. It's just a bunch of hashes, not much else so probably not worth it? Especially if the linter runs on every build/pre-commit. But if people disagree I'm happy to extend it to other directories...

PR #849, opened 2020-04-15
HUB-504: Pass variant name

jkunle:
**What**
This is the Frontend portion of the card below
https://govukverify.atlassian.net/browse/HUB-504
In a world in which we can't use analytics to measure AB tests, we need to compare completion rate across variants in some other way.

To do this, we'll need to pass the AB test variant name from the frontend to the events system.

Any event currently recorded in the events system that that has an analytics user ID should now also have a list of variant names that the user was in at the point the event occurred.


config/main_routes.rb
```diff
@@ -42,15 +36,15 @@ def add_routes(routes_name)
 end
 
 constraints IsLoa2 do
-  constraints sign_in_hint_control do
+  constraints sign_in_hint.use(alternative: "control") do
```
jakubmiarka, line 39: this is nice!

spec/features/user_visits_redirect_to_idp_warning_page_spec.rb
```diff
@@ -55,14 +55,17 @@
       encrypted_entity_id,
       PolicyEndpoints::PARAM_SELECTED_ENTITY_ID => idp_entity_id, PolicyEndpoints::PARAM_PRINCIPAL_IP => originating_ip,
       PolicyEndpoints::PARAM_REGISTRATION => true, PolicyEndpoints::PARAM_REQUESTED_LOA => "LEVEL_2",
-      PolicyEndpoints::PARAM_ANALYTICS_SESSION_ID => piwik_session_id, PolicyEndpoints::PARAM_JOURNEY_TYPE => nil
+      PolicyEndpoints::PARAM_ANALYTICS_SESSION_ID => piwik_session_id, PolicyEndpoints::PARAM_JOURNEY_TYPE => nil,
+      PolicyEndpoints::PARAM_VARIANT => nil
```
jakubmiarka, line 59: I wonder if there should be at least one test which is testing the value? i.e. where it's being reported and not being nil?

jkunle:
Thanks @jakubmiarka , I have put updating the Verify Team Manual as part of the Acceptance Criteria for this card.

PR #851, opened 2020-04-16
Revert "HUB-584: Update Rake version"

jakubmiarka:
Reverts alphagov/verify-frontend#848

This supposedly broke logging in AWS. Trying to see if it's really the case by reverting this.


jakubmiarka:
This can't be it! I'll have another PR shortly

PR #853, opened 2020-04-16
Revert "BAU: Fix (hopefully) logging to logstash"

jakubmiarka:
Reverts alphagov/verify-frontend#852

Nah, doesn't fix anything, but we might be on something...

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -42,7 +42,3 @@ Layout/LeadingCommentSpace:
 Style/PercentLiteralDelimiters:
   PreferredDelimiters:
     Default: ()
-
-Style/StringLiterals:
-  Exclude:
-    - 'config/initializers/*.rb'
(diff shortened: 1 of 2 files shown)


app/controllers/authn_request_controller.rb
```diff
@@ -63,11 +63,11 @@ def set_transaction_entity_id(entity_id)
   end
 
   def set_requested_loa(levels_of_assurance)
-    requested_loa = if levels_of_assurance.include? "LEVEL_1"
-                      "LEVEL_1"
-                    else
-                      levels_of_assurance.first
-                    end
+    if levels_of_assurance.include? "LEVEL_1"
+      puts requested_loa = "LEVEL_1"
```
JHjava, line 67: Can we keep this change in though?

jakubmiarka: 🙄 ugh, if you say so 😆  .. just pushed

PR #854, opened 2020-04-17
Allow any LOA1 service that has eIDAS enabled to see the prove identity page

JHjava:
- We want any service that use this ordering of LOAs [LOA2, LOA1] to be able to see the prove identity page when they have eIDAS enabled in their transaction config.
- To achieve this we need to add the prove identity to the LOA1 routes.

### DIFF
--- a/spec/features/user_sends_authn_request_spec.rb
+++ b/spec/features/user_sends_authn_request_spec.rb
@@ -35,6 +35,19 @@ describe "user sends authn requests" do
       expect(page.get_rack_session["transaction_supports_eidas"]).to eql false
     end
 
+    it "will redirect the user to /choose-a-country for an eidas journey where eidas is enabled and RP LOA is LEVEL_1" do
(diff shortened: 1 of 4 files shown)


spec/features/user_sends_authn_request_spec.rb
```diff
@@ -35,6 +35,19 @@
...
+      stub_countries_list
+
+      visit("/test-saml")
+      click_button "saml-post-eidas"
+
+      expect(page).to have_title t("hub.choose_a_country.title")
+      expect(page.get_rack_session["requested_loa"]).to eql "LEVEL_1"
+      expect(page.get_rack_session["transaction_supports_eidas"]).to eql true
+    end
+
```
galund, line 50: I wonder if the next test should be modified to more clearly target LOA2, if that's what's going on here?

JHjava: Good suggestion. I will amend and push up the change.

PR #857, opened 2020-05-07
HUB-595: Remove AB Test Sign in Hint

j-maynard:


app/controllers/authn_request_controller.rb
```diff
@@ -15,11 +15,6 @@ def rp_request
     session_journey_hint_rp = session.fetch(:journey_hint_rp, nil)
     create_session
 
-    # HUB-113: Temporarily disabling to allow the perf team to analyze the data
-    # session[:new_visit] = true
```
galund, line 19: HUB-113 is old so surely this needs to go back - maybe need to figure out if we will ever be re-enabling it though, and if not then fine perhaps it can go.

CharlesIC:
Replaced by #858

PR #858, opened 2020-05-07
HUB-595: Tear down sign-in hint AB test & make hint part of the journey

jkunle:
https://govukverify.atlassian.net/browse/HUB-595
**What**

These are the front-end changes required to turn of sign-in hint

We need to tear down the current signin hint AB test once 2 weeks have elapsed (after Monday 20 April).

Depending on 's analysis, we should either:

revert to the control (A), which has no signin hint

ship the variant (B), which has the signin hint

**Why**

Because we'll know whether the hint was successful or not, so we should now either implement it or not.

**How**
This PR ships variant (B) as live in place of variant (A), we have decided variant (B) is successful.


app/controllers/prove_identity_controller.rb
```diff
@@ -1,6 +1,34 @@
...
 class ProveIdentityController < ApplicationController
+  include JourneyHintingPartialController
+  include ViewableIdpPartialController
+  include UserCookiesPartialController
+
   def index
-    render :prove_identity
+    journey_hint_entity_id = success_entity_id
+    if journey_hint_entity_id.nil?
+      render "prove_identity/prove_identity"
```
jakubmiarka, line 13:
```suggestion
      render :prove_identity
```

app/controllers/prove_identity_controller.rb
```diff
@@ -1,6 +1,34 @@
...
+    else
+      @identity_provider = retrieve_decorated_singleton_idp_array_by_entity_id(current_available_identity_providers_for_sign_in, journey_hint_entity_id).first
+      if @identity_provider.nil?
+        return render "prove_identity/prove_identity"
```
jakubmiarka, line 17:
```suggestion
        return render :prove_identity
```

app/controllers/start_controller.rb
```diff
@@ -1,12 +1,38 @@
...
   before_action :set_device_type_evidence
 
   def index
     restart_journey if identity_provider_selected? && !user_journey_type?(JourneyType::VERIFY)
+    journey_hint_entity_id = success_entity_id
     @form = StartForm.new({})
     @journey_hint = flash[:journey_hint]
-    render :start
+    if journey_hint_entity_id.nil?
+      render "start/start"
```
jakubmiarka, line 17:
```suggestion
      render :start
```

app/controllers/start_controller.rb
```diff
@@ -1,12 +1,38 @@
...
+    else
+      @identity_provider = retrieve_decorated_singleton_idp_array_by_entity_id(current_available_identity_providers_for_sign_in, journey_hint_entity_id).first
+      if @identity_provider.nil?
+        return render "start/start"
```
jakubmiarka, line 21:
```suggestion
        return render :start
```

app/controllers/start_controller.rb
```diff
@@ -19,7 +45,7 @@ def request_post
       end
     else
       flash.now[:errors] = @form.errors.full_messages.join(", ")
-      render :start
+      render "start/start"
```
jakubmiarka, line 48:
```suggestion
      render :start
```

spec/features/user_visits_prove_your_identity_page_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "feature_helper"
 require "api_test_helper"
 
-RSpec.describe "When the user visits the prove identity page" do
+RSpec.describe "When the user visits the prove identity page on variant" do
```
jakubmiarka, line 4:
```suggestion
RSpec.describe "When the user visits the prove identity page" do
```

spec/features/user_visits_prove_your_identity_page_spec.rb
```diff
@@ -12,7 +12,7 @@
 
   context "will display the prove identity page" do
     before(:each) do
-      set_session_and_session_cookies!
+      set_session_and_ab_session_cookies!("sign_in_hint" => "sign_in_hint_variant")
```
jakubmiarka, line 15:
```suggestion
      set_session_and_session_cookies!
```

spec/features/user_visits_start_page_spec.rb
```diff
@@ -3,25 +3,56 @@
...
-  it "will display the start page in English" do
-    set_session_and_session_cookies!
-    visit "/start"
-    expect(page).to have_content t("hub.start.heading")
-    expect(page).to have_css "html[lang=en]"
-    expect_feedback_source_to_be(page, "START_PAGE", "/start")
-  end
+  context "when session is valid" do
+    before(:each) do
+      set_session_and_ab_session_cookies!("sign_in_hint" => "sign_in_hint_variant")
```
jakubmiarka, line 8:
```suggestion
      set_session_and_session_cookies!
```

spec/features/user_visits_start_page_spec.rb
```diff
@@ -3,25 +3,56 @@
...
+    end
 
-  it "will display the start page in Welsh" do
-    set_session_and_session_cookies!
-    visit "/dechrau"
-    expect(page).to have_content t("hub.start.heading", locale: :cy)
-    expect(page).to have_css "html[lang=cy]"
-  end
+    it "will display the start page in English on variant" do
```
jakubmiarka, line 11:
```suggestion
    it "will display the start page in English" do
```

CharlesIC:
The copy on the hint is incorrect - it should look like this:

Screenshot 2020-05-04 at 12 15 16

CharlesIC:
The copy has been updated in my commit

app/controllers/paused_registration_controller.rb
```diff
@@ -163,8 +163,8 @@ def select_resume(entity_id, idp_name)
   def is_resume_link_for_pending_idp?(idp_simple_id)
     return false unless is_last_status?(PENDING_STATUS)
 
-    idp = retrieve_decorated_singleton_idp_array_by_entity_id(current_identity_providers_for_registration_rp_loa2(last_rp), last_idp).first
+    idp = decorate_idp_by_entity_id(current_available_identity_providers_for_registration_loa2(last_rp), last_idp)
 
-    idp.simple_id == idp_simple_id
+    !idp.nil? && idp.simple_id == idp_simple_id
```
JohnWattsGDS, line 168: does `idp&.simple_id == idp_simple_id` do what you want there?  [Probably better than me trying to explain what I think it's doing](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/)

CharlesIC: There's no `&` there, though, just `idp.simple_id == ...`
Also, I didn't change that bit - only added the null check at the start of the line

JohnWattsGDS: the `&.` operator does your null check for you and returns `nil` if `idp` is `nil` so (assuming `idp_simple_id` isn't going to be `nil` at any point), you avoid the null check.

CharlesIC: Hmm it would but only because of a weird edge case where `nil` in conjunction with `&.` always responds with `nil` (and then `nil` is not equal to the string). But that's not how `&.` is intended to be used - it's not used to check that the object itself is null but that its specific defined properties are not null. For instance, `"hello"&.nonExistentProperty` would throw an error. `nil` is an exception in that if you did the same to `nil` it would just give you `nil`....
I'd rather keep the logic above as it is for avoidance of doubt and to not rely on weird edge case behaviour.

CharlesIC: (For people that know Ruby maybe it's not a weird edge case but I don't know it that well so I don't feel confident pulling too many tricks)

jkunle: Nice work on the refactoring, i don't think &. is a weird edge case https://bugs.ruby-lang.org/issues/11537 it was specifically and deliberately introduced as a "Safe navigation operator" by the ruby creator himself. So the above is equivalent to `idp && idp.simple_id == idp_simple_id` <==> `idp&.simple_id == idp_simple_id`

app/controllers/confirm_your_identity_controller.rb
```diff
@@ -6,17 +6,19 @@ class ConfirmYourIdentityController < ApplicationController
   include JourneyHintingPartialController
 
   def index
-    journey_hint_entity_id = attempted_entity_id
+    journey_hint_idp_entity_id = attempted_entity_id
```
jkunle, line 9: I think this `attempted_entity_id ` should be renamed in `JourneyHintingPartialController` instead of reassigned
```
    return cookie_error("missing verify-front-journey-hint") if journey_hint_idp_entity_id.nil?

    idp = decorate_idp_by_entity_id(current_available_identity_providers_for_registration, journey_hint_idp_entity_id)
    cookie_error("invalid verify-front-journey-hint entity-id #{journey_hint_idp_entity_id}") if idp.nil?

    @identity_providers = [idp]
    @transaction_name = current_transaction.name
```
But your implementation is fine too.

PR #861, opened 2020-05-18
HUB-247 Build new hub journey that comes after failure

JohnWattsGDS:
Present different content to users depending on how many IDPs are available to them to choose.

When a user fails registration, the IDP they failed with is saved into an array session variable `idps_tried`.  This array is used to determine how many untried IDPs remain which in turn determines which content is displayed.

Missing from this PR is the requirement for a link pointing to the IDPs sign-in page.  This has been left out to get this story over the line as soon as possible.  Passing the values for the links would require alterations to multiple microservices and configuration files in multiple repositories and can form an independent separate story without affecting the effectiveness of this story.


app/controllers/failed_registration_loa2_controller.rb
```diff
@@ -1,12 +1,12 @@
+require "partials/failed_registration_partial_controller"
+
 class FailedRegistrationLoa2Controller < ApplicationController
+  include FailedRegistrationPartialController
+
   def index
     @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider)
     @transaction = current_transaction
     @custom_fail = !current_transaction.custom_fail_heading.nil?
```
jkunle, line 9: I think this reads better `current_transaction.custom_fail_heading.present?`

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  def continue_on_failed_registration_rp?
+    CONTINUE_ON_FAILED_REGISTRATION_RPS.include?(current_transaction_simple_id)
+  end
+
+  def tried_all_idps?
+    (recommended_idps + unlikely_idps).reject { |idp| idps_tried.any? idp }.count.zero?
+  end
+
+  def suggested_idps
+    @suggested_idps || @suggested_idps = IDP_RECOMMENDATION_ENGINE.get_suggested_idps_for_registration(current_available_identity_providers_for_registration,
```
jkunle, line 41: Wouldn't the below work?
```
  @suggested_idps ||= IDP_RECOMMENDATION_ENGINE.get_suggested_idps_for_registration(current_available_identity_providers_for_registration,
                                                                                      selected_evidence,
                                                                                      current_transaction_simple_id)
```

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def choose_partial_for_loa2
+    choose_partial "LOA2"
+  end
+
+private
+
+  def choose_partial(loa_suffix)
+    if continue_on_failed_registration_rp?
+      "failed_registration/index_continue_on_failed_registration_" + loa_suffix
```
jkunle, line 20: I like "failed_registration/index_continue_on_failed_registration_#{loa_suffix}"

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+    else
+      add_current_idp_to_idps_tried
+      @remaining_options_partial = if tried_all_idps?
+                                     "failed_registration/options_with_no_other_idps"
+                                   else
+                                     "failed_registration/options_with_other_idps"
+                                   end
+      "failed_registration/index_" + loa_suffix
```
jkunle, line 28: `"failed_registration/index_#{loa_suffix}"` you don't have to change to this

app/controllers/failed_registration_loa2_controller.rb
JohnWattsGDS: It probably does.  The less `not`s the better in my book.

app/controllers/partials/failed_registration_partial_controller.rb
JohnWattsGDS: That is the more idiomatic-ruby way of doing it - my java background's showing.

app/controllers/partials/failed_registration_partial_controller.rb
JohnWattsGDS: But I will.

app/controllers/partials/failed_registration_partial_controller.rb
JohnWattsGDS: I think it should, and would be better.

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+      "failed_registration/index_#{loa_suffix}"
+    end
+  end
+
+  def continue_on_failed_registration_rp?
+    CONTINUE_ON_FAILED_REGISTRATION_RPS.include?(current_transaction_simple_id)
+  end
+
+  def tried_all_idps?
+    (recommended_idps + unlikely_idps).reject { |idp| idps_tried.any? idp }.count.zero?
```
jkunle, line 37: What do you think about  this `(recommended_idps + unlikely_idps - idps_tried).none?`

JohnWattsGDS: Will that really work? I think I like it, let me look at it in RubyMine.

JohnWattsGDS: So, I've opted for `((recommended_idps + unlikely_idps) - idps_tried).none?` as it's a little clearer on intent with the extra brackets.

jkunle: Yeah, this is fine.

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+    idps_tried.none? { |idp| idp == selected_identity_provider.simple_id }
+  end
+
+  def add_current_idp_to_idps_tried
+    idps_tried << selected_identity_provider.simple_id if not_tried_current_idp_before?
+  end
+
+  def idps_tried
+    session[:idps_tried] ||= []
+  end
```
jkunle, line 64: May be this should move to `viewable_idp_partial_controller.rb` then we can remove `idps_tried = session[:idps_tried] || []` in current_available_identity_providers_for_registration

JohnWattsGDS: At first I tried to include `FailedRegistrationPartialController` in 'ViewableIdpPartialController` but that results in circular dependencies so that's not an option.

That particular method was added to follow clean code ideas and to make the failed registration code as narrative as possible so I don't think it's right to put it in a different controller as it confuses things for the sake of a single line that's repeated.

There's the option of another partial but then that would probably only exist for the sake of housing that particular method so seems a little heavy handed but would be the right approach if we were to take DRY to that level.  I think leaving it as it is okay.

Would I like to do a larger scale refactoring job and find all of the similar bits of repeated code and gather them into related controllers and make the whole codebase read like a story? Yes!  But I think in this case, it's fine as it is.

spec/feature_helper.rb
```diff
@@ -185,6 +185,7 @@ def default_session
       transaction_entity_id: "http://www.test-rp.gov.uk/SAML2/MD",
       transaction_homepage: "http://www.test-rp.gov.uk/",
       selected_answers: { device_type: { device_type_other: true } },
+      idps_tried: []
```
jkunle, line 188: I see what you mean about `idp_tried`, am happy to approve this; once you put the comma **here** ```idps_tried: []```

JohnWattsGDS: I've gone one better and deleted it because I added it for a test but it turns out it's not needed. :)

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+private
+
+  def choose_partial(loa_suffix)
+    if continue_on_failed_registration_rp?
+      "failed_registration/index_continue_on_failed_registration_#{loa_suffix}"
+    else
+      add_current_idp_to_idps_tried
+      @remaining_options_partial = if tried_all_idps?
```
CharlesIC, line 23: You can use the ternary operator (`q ? a : b`) here instead of `if`

spec/controllers/failed_registration_loa1_controller_spec.rb
```diff
@@ -20,6 +20,7 @@
     end
 
     it "index view when rp is not allowed to continue on failed" do
+      stub_api_idp_list_for_registration(default_idps, "LEVEL_1")
```
CharlesIC, line 23: This is not enough - there's a missing test for the case when RP is not allowed to continue and all IDPs have been tried - this is the new logic. I know there's a feature test as well for this but we still need a unit test in this controller and the LOA2 version (yes, we need to duplicate the test because of duplicated controllers)

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -42,28 +42,60 @@
...
+    context "there are more IDPs to try" do
+      before(:each) do
+        session = default_session
+        session[:selected_answers] = {
+            "documents" => { "driving_licence" => true, "passport" => true },
+            "phone" => { "mobile_phone" => true },
+            "device_type" => { "device_type_other" => true },
+        }
+        page.set_rack_session(session)
+      end
```
CharlesIC, line 54:
```suggestion
      end

```

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -42,28 +42,60 @@
...
+        set_loa_in_session("LEVEL_1")
+        stub_api_idp_list_for_registration(default_idps, "LEVEL_1")
+        visit "/failed-registration"
+
+        expect_page_to_have_main_content_non_continue
+        expect(page).to have_content t("hub.failed_registration.remaining_idps.different_way",
+                                       service: "test GOV.UK Verify user journeys")
+        expect(page).to have_link t("hub.failed_registration.remaining_idps.link_text"), href: choose_a_certified_company_path
+      end
     end
```
CharlesIC, line 75:
```suggestion
    end

```

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -42,28 +42,60 @@
...
+        stub_api_idp_list_for_registration(default_idps, "LEVEL_1")
+        visit "/failed-registration"
+
+        expect_page_to_have_main_content_non_continue_for_no_idps
+        expect(page).to have_content t("hub.failed_registration.last_idp.different_way",
+                                       service: "test GOV.UK Verify user journeys")
+        expect(page).not_to have_link t("hub.failed_registration.start_again"), href: choose_a_certified_company_path
+      end
     end
   end
```
CharlesIC, line 98:
```suggestion
  end

```

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -86,11 +118,49 @@
...
 
+  context "IDPs are available again when the user starts a new session" do
```
CharlesIC, line 121: This test should be in the subcontext `there are no more IDPs to try` of context `relying party is not allowed to continue on fail` - `IDPs are available again when the user starts a new session` doesn't make sense as a context as it's the thing we're testing, not a condition.

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -86,11 +118,49 @@
...
+    it "makes a journey where no IDPs are left" do
+      page.set_rack_session(transaction_simple_id: DONT_CONTINUE_ON_FAILED_REGISTRATION_RP)
+      set_loa_in_session("LEVEL_2")
+      visit "/failed-registration"
 
-  def expect_page_to_have_main_content
+      expect_page_to_have_main_content_non_continue_for_no_idps
+
+      set_session_and_session_cookies!
```
CharlesIC, line 129: We should use `set_session!` here instead as in the real world you'd have a new session but the same cookies. Actually this line is not needed at all as the session is set below.

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -86,11 +118,49 @@
...
+      session = default_session
+      session[:selected_answers] = {
+          "documents" => { "driving_licence" => true, "passport" => true },
+          "phone" => { "mobile_phone" => true },
+          "device_type" => { "device_type_other" => true },
+      }
+      page.set_rack_session session
```
CharlesIC, line 136: We have a helper method to set the session so you can do `set_session!(session)`

spec/features/user_visits_failed_registration_page_spec.rb
```diff
@@ -86,11 +118,49 @@
...
+      set_selected_idp_in_session(entity_id: "http://idcorp.com", simple_id: "stub-idp-one")
+      page.set_rack_session(transaction_simple_id: DONT_CONTINUE_ON_FAILED_REGISTRATION_RP)
+      set_loa_in_session("LEVEL_2")
```
CharlesIC, line 139:
```suggestion
      set_loa_in_session("LEVEL_2")

```

app/controllers/partials/failed_registration_partial_controller.rb
CharlesIC: The extra brackets are not needed here - it's simple addition and subtraction read from left to right, it should be obvious what the intent is. I'm surprised Rubocop allows this 😬

jkunle: This is why am torn with ruby `recommended_idps.union(unlikely_idps).difference(idps_tried).none?`

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+    suggested_idps[:recommended]
+  end
+
+  def unlikely_idps
+    suggested_idps[:unlikely]
+  end
+
+  def not_tried_current_idp_before?
+    idps_tried.none? { |idp| idp == selected_identity_provider.simple_id }
+  end
```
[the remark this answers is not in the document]
JohnWattsGDS: I believe this is moot if I use a set.

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+                                     "failed_registration/options_with_no_other_idps"
+                                   else
+                                     "failed_registration/options_with_other_idps"
+                                   end
+      "failed_registration/index_#{loa_suffix}"
+    end
+  end
+
+  def continue_on_failed_registration_rp?
+    CONTINUE_ON_FAILED_REGISTRATION_RPS.include?(current_transaction_simple_id)
```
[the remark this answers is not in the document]
JohnWattsGDS: I didn't with this because of the argument and I'm not sure about the Ruby magic that gives us `current_transaction_simple_id` and whether it would be in-scope in the file that provides `CONTINUE_ON_FAILED_REGISTRATION_RPS`. That sort of thing would be part of a wider re-factoring effort I think to avoid pushing problems around without solving them.

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  def unlikely_idps
+    suggested_idps[:unlikely]
+  end
+
+  def not_tried_current_idp_before?
+    idps_tried.none? { |idp| idp == selected_identity_provider.simple_id }
+  end
+
+  def add_current_idp_to_idps_tried
+    idps_tried << selected_identity_provider.simple_id if not_tried_current_idp_before?
```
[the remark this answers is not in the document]
JohnWattsGDS: 👍

app/controllers/partials/failed_registration_partial_controller.rb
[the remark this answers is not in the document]
CharlesIC: Yep it can be done separately

app/controllers/partials/failed_registration_partial_controller.rb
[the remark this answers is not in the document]
CharlesIC: Yep indeed, this is not needed if we use a Set

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+private
+
+  def choose_partial(loa_suffix)
+    if continue_on_failed_registration_rp?
+      "failed_registration/index_continue_on_failed_registration_#{loa_suffix}"
+    else
+      add_current_idp_to_idps_tried
+      @remaining_options_partial = tried_all_idps? ? "failed_registration/options_with_no_other_idps" : "failed_registration/options_with_other_idps"
```
jkunle, line 23: This is slightly shorter `@remaining_options_partial = "failed_registration/#{tried_all_idps? ? 'options_with_no_other_idps' : 'options_with_other_idps'}"`

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      "failed_registration/index_#{loa_suffix}"
+    end
+  end
+
+  def continue_on_failed_registration_rp?
+    CONTINUE_ON_FAILED_REGISTRATION_RPS.include?(current_transaction_simple_id)
+  end
+
+  def tried_all_idps?
+    ((recommended_idps + unlikely_idps) - idps_tried.to_a).none?
```
jkunle, line 33: I think Karol preferred this `(recommended_idps + unlikely_idps - idps_tried.to_a).none?` and most people should understand `+` has precedence over `-`

app/controllers/partials/failed_registration_partial_controller.rb
JohnWattsGDS: Size isn't important though.  Certainly not in this case.  Personally I actually prefer the original `if` construct over the ternary operator as the if-statement version needs virtually no cognitive process to understand but the ternary operator is a fairly well known paradigm so I'll submit to peer pressure and use it for that reason.  I'm actually tempted to switch back to the if-statement because almost anyone could understand what's happening there whether they have a background in coding or not.

Shifting the ternary operator inside the string interpolation is a step to far though and hides point of the code - the important thing to note on this line is that we're making a decision based on whether or not we've tried all of the IDPs and putting that right at the start of the expression ensures that it's as clear as possible.

A long comment I know, but I value clarity of code over GitHub's storage costs.

app/controllers/partials/viewable_idp_partial_controller.rb
```diff
@@ -33,7 +33,7 @@ def for_viewable_idp(entity_id, identity_provider_list)
   end
 
   def current_available_identity_providers_for_registration
-    CONFIG_PROXY.get_available_idp_list_for_registration(session[:transaction_entity_id], session[:requested_loa]).idps
+    CONFIG_PROXY.get_available_idp_list_for_registration(session[:transaction_entity_id], session[:requested_loa]).idps.reject { |idp| idps_tried.include? idp.simple_id }
```
jkunle, line 36: This is fine but I prefer  `CONFIG_PROXY.get_available_idp_list_for_registration(session[:transaction_entity_id], session[:requested_loa])
     .idps.select(&:simple_id) - idps_tried.to_a`

app/controllers/partials/failed_registration_partial_controller.rb
jkunle: I actually preferred the `if` construct as well. For me the ternary looks to long, I tend to go for the `if`

app/controllers/partials/viewable_idp_partial_controller.rb
JohnWattsGDS: I don't think that actually works because `idps_tried.to_a` is an array of strings (the simple_id) but `CONFIG_PROXY.get_available_idp_list_for_registration` returns an array of IDPs so the subtraction always returns the full list because none of the IDPs are equal to simple_id.

app/controllers/failed_registration_loa1_variant_c_controller.rb
```diff
@@ -0,0 +1,12 @@
+require "partials/failed_registration_partial_controller"
```
galund, line 1: shtop shtop Variant C is only in the LOA2 journey so what is this thing?

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def choose_partial_for_loa2
+    choose_partial "LOA2"
+  end
+
+private
+
+  def choose_partial(loa_suffix)
+    if continue_on_failed_registration_rp?
+      "failed_registration/index_continue_on_failed_registration_#{loa_suffix}"
```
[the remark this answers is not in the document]
galund: Okay I don't get this at all - this feels like a case for regular inheritance rather than including partials, at which point having different behaviour between subclasses is easy and completely standard.

app/views/cancelled_registration_loa1/cancelled_registration_LOA1.html.erb
```diff
@@ -10,7 +10,7 @@
 
   <div>
     <p><%= t 'hub.cancelled_registration.try_another_text', idp_name: @idp.display_name %></p>
-    <%= link_to t('hub.cancelled_registration.about_choosing_a_company'), about_choosing_a_company_path%>
+    <p><%= link_to t('hub.cancelled_registration.about_choosing_a_company'), about_choosing_a_company_path%></p>
```
galund, line 13: We seem to be inconsistent about whether we need `class="govuk-body"` sprinkled everywhere. I'm very happy if in fact we definitely don't need any more :-)

app/views/failed_registration/_options_with_no_other_idps.html.erb
```diff
@@ -0,0 +1,3 @@
+<p class="govuk-body"><%= t 'hub.failed_registration.last_idp.text' %></p>
+
+<h2 class="govuk-heading-m"><%= t 'hub.failed_registration.last_idp.different_way', service:  @transaction.name %></h2>
```
galund, line 3: It's probably a bit weird leaving the final newlines out in terms of how this comes out in the user's browser, if we are ever (god forbid) needing to look at that!

app/controllers/partials/failed_registration_partial_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    choose_partial "LOA1"
+  end
+
+  def choose_partial_for_loa2
+    choose_partial "LOA2"
+  end
+
+private
+
+  def choose_partial(loa_suffix)
```
galund, line 18: I think you meant to call this 'choose_view' rather than 'choose_partial', unless I'm very confused.

I'd much rather use inheritance. Maybe this should be a discussion on Slack or a google meet :)

app/controllers/failed_registration_loa1_controller.rb
```diff
@@ -1,12 +1,11 @@
...
   def index
     @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider)
     @transaction = current_transaction
-
-    if CONTINUE_ON_FAILED_REGISTRATION_RPS.include?(current_transaction_simple_id)
-      render "failed_registration/index_continue_on_failed_registration_LOA1"
-    else
-      render "failed_registration/index_LOA1"
-    end
+    render choose_partial_for_loa1
```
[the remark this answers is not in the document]
galund: Simplest change would be to make this function call `choose_partial` (which is named wrong) directly, then you don't need a weird function in the partial class.

config/main_routes.rb
```diff
@@ -68,7 +68,7 @@ def add_routes(routes_name)
   # post "choose_a_certified_company", to: "choose_a_certified_company_loa2#select_idp", as: :choose_a_certified_company_submit
 
   get "why_companies", to: "why_companies_loa2#index", as: :why_companies
-  get "failed_registration", to: "failed_registration_loa2#index", as: :failed_registration
+  # get "failed_registration", to: "failed_registration_loa2_variant_c#index", as: :failed_registration
```
galund, line 71: I feel like the commented-out version should be the one without _variant_c on the end, but eh

PR #862, opened 2020-05-20
HUB-468, PS-53, create new persistent session id

instantiator:
* This PR updates the `analytics_session_id` method in `AnalyticsCookiePartialController` to generate and return its own cookie, distinct from Matomo.
* The cookie holds a random id, with an expiry of 13 months (matching the current behaviour).
* Tests that expect `analytics_session_id` to be `nil` are updated to reflect the fact that this id is always generated.

For context, see:

* [HUB-468](https://govukverify.atlassian.net/browse/HUB-468) (jira)
* [PS-53](https://trello.com/c/CxxaKSUg/53-preserve-ability-to-report-investigate-fraud-using-non-analytics-cookie) (trello)

### DIFF
--- a/app/controllers/partials/analytics_cookie_partial_controller.rb
+++ b/app/controllers/partials/analytics_cookie_partial_controller.rb
@@ -3,19 +3,15 @@ require "securerandom"
 # analytics_session_id is disconnected from matomo and generated if not present
 module AnalyticsCookiePartialController
   def analytics_session_id
-    cookie_value = cookies.fetch(persistent_cookie_name, nil)
+    cookie_value = cookies.fetch(CookieNames::PERSISTENT_SESSION_ID_COOKIE_NAME, nil)
     if cookie_value.nil?
       cookie_value = SecureRandom.uuid
-      cookies[persistent_cookie_name] = { value: cookie_value, expires: 13.months.from_now }
+      cookies[CookieNames::PERSISTENT_SESSION_ID_COOKIE_NAME] = { value: cookie_value, expires: 13.months.from_now }
     end
     cookie_value
   end
 
 private
-
-  def persistent_cookie_name
-    CookieNames::PERSISTENT_SESSION_ID_COOKIE_NAME
-  end
 
   def analytics_cookie_name
     cookies.each { |name, _value| return name if name.starts_with? CookieNames::ANALYTICS_SESSION_COOKIE_PREFIX }
--- a/app/controllers/partials/analytics_cookie_partial_controller.rb
+++ b/app/controllers/partials/analytics_cookie_partial_controller.rb
@@ -1,13 +1,18 @@
 require "securerandom"
 
-# analytics_session_id is disconnected from matomo and generated if not present
+# persistent session id is taken from matomo if available, and generated if not
 module AnalyticsCookiePartialController
   def analytics_session_id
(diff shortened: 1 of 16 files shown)


app/controllers/partials/analytics_cookie_partial_controller.rb
```diff
@@ -1,11 +1,22 @@
...
+      cookie_value = SecureRandom.uuid
+      cookies[persistent_cookie_name] = { value: cookie_value, expires: 13.months.from_now }
+    end
+    cookie_value
   end
 
 private
 
+  def persistent_cookie_name
+    CookieNames::PERSISTENT_SESSION_ID_COOKIE_NAME
```
galund, line 17: Why can't we just refer to the constant everywhere? Feels like this doesn't need to be a function, but maybe there is some Ruby-foo I'm not down with.

instantiator: You're right it's not essential - it just matches a pattern. As this is the only place it's referred to, we might as well do that. I'll change it now.

JHjava: Good point. I think because we just replicated the function below but now as it is just a constant then it probably makes more sense to refer to it directly within `def analytics_session_id`.

app/controllers/partials/analytics_cookie_partial_controller.rb
```diff
@@ -1,11 +1,22 @@
...
+  end
+
   def analytics_cookie_name
```
JHjava, line 20: We can remove this now as it isn't being used.

JHjava: I was too slow and I didn't refresh the page!

spec/features/idp_selections_reported_to_piwik_spec.rb
```diff
@@ -1,6 +1,7 @@
 require "feature_helper"
 require "api_test_helper"
 require "piwik_test_helper"
+require "securerandom"
```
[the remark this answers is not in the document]
instantiator: Ah yes - I'll remove.

app/controllers/partials/analytics_cookie_partial_controller.rb
```diff
@@ -1,7 +1,14 @@
+require "securerandom"
+
+# analytics_session_id is disconnected from matomo and generated if not present
 module AnalyticsCookiePartialController
   def analytics_session_id
```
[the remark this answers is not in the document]
instantiator: We've renamed to `persistent_session_id`.

app/controllers/partials/analytics_cookie_partial_controller.rb
```diff
@@ -1,7 +1,19 @@
...
+    cookie_value = cookies.fetch(CookieNames::PERSISTENT_SESSION_ID_COOKIE_NAME, nil)
+    if cookie_value.nil?
+      matomo_value = cookies.fetch(analytics_cookie_name, nil)
+      cookie_value = matomo_value.split(".").first unless matomo_value.nil?
+      if cookie_value.nil?
+        cookie_value = SecureRandom.uuid
+      end
+    end
+    # analytics session id lasts for 13 months from use - so always refresh the cookie
+    cookies[CookieNames::PERSISTENT_SESSION_ID_COOKIE_NAME] = { value: cookie_value, expires: 13.months.from_now }
```
galund, line 15: Arguably the 13 should be a config value somewhere but whatevs

PR #864, opened 2020-05-26
HUB-662: Remove delays and weird journeys from the hub

jkunle:
**What**
https://govukverify.atlassian.net/browse/HUB-662
When Covid-19 caused huge peaks in traffic for UC, we added stuff to the hub to:

manage expectations about the IDP delays

explain that users would need to go back to the service after the IDP

These changes appear in the registration users journey after the user selects an IDP on the IDP picker page.

We need to revert these changes now.

**Why**

The delays are gone.We've mitigated the journey problems in other ways.

Tech details

Remove what-happens-next

Update content on redirect-to-idp-warning to be what it was before the change
**How**
This reverts commit e02f399e9587aef7d3210158b4600108e0ee8ed1.
**Before**
image

image

**After**
image


jkunle:
@bethwalton the reason for the above is that the screenshots were taken in development mode, where the logo links are stubbed out. Those logos would be picked up from verify-frontend-hub-configuration when run in prod here is a review app https://verify-front-hub-662-re-txfzbt.herokuapp.com/ that show stub logos

PR #866, opened 2020-05-29
BAU Address Vulnerabilities

JohnWattsGDS:
Make sure active support is at version 5.2.4.3 (or above).


Gemfile.lock
```diff
@@ -75,26 +74,26 @@ GEM
...
+    dotenv-rails (2.7.5)
+      dotenv (= 2.7.5)
       railties (>= 3.2, < 6.1)
     email_validator (1.6.0)
       activemodel
     erubi (1.9.0)
     eventmachine (1.2.7)
     execjs (2.7.0)
-    faraday (0.15.4)
+    faraday (1.0.1)
```
galund, line 88: There are some bigger upgrades here that I think wouldn't have happened if we'd have done the 'normal' thing of leaving the lock file in-place when updating. But if it works, well this is great, and I think we should do it more often.

PR #867, opened 2020-06-01
Log sign in last successful IDP hint

JHjava:
- The 'sign-in-last-sucessful-idp' is never set in the session and if you look in the logs you will find no record of it. We are currently attempting to set it in the redirect_to_idp_controller.rb however this is not working. I want to find out exactly what this is doing before removing it. 
- In order for the 'sign-in-last-sucessful-idp'  to be set we need to set it in the 'journey_hinting_partial_controller.rb' just before the 'sign-in-hint' page is displayed to the user. Then the journey_type we be successfully populated and then later logged in policy. 
- Fix the spelling of successful


galund:
This feels like the stuff that @JohnWattsGDS was working on to do with the short-hub-failure, and that work is currently in pull request - see https://govukverify.atlassian.net/browse/HUB-247

Even if it's unrelated, then it sounds related _enough_ that I definitely want the two things to be looked at together :-)

JHjava:
@galund Well I'll leave it open so you have it for reference and can decide if you want to use it or not to fix the gap in the logs :)

instantiator:
> This feels like the stuff that @JohnWattsGDS was working on to do with the short-hub-failure, and that work is currently in pull request - see https://govukverify.atlassian.net/browse/HUB-247

FYI, HUB-247's PR is [#861 in verify-frontend](https://github.com/alphagov/verify-frontend/pull/861). It doesn't have any mention of `journey_type` or `sign-in-last-successful-idp` - so I guess it doesn't set out to fix the "`journey_type` is missing" issue.

The next question is if there are any new journeys that it creates where `sign-in-last-successful-idp` is no longer the intended/correct `journey_type`? As it doesn't set any new `journey_types` itself I'm guessing not - but perhaps we should chat to @JohnWattsGDS about it so we know for sure.

Either way, it looks a lot to me like this PR is a small fix that's reasonably independently from HUB-247. @galund I guess you know best about why there might be a cross-over?

galund:
In my head the connection is "stuff that happens after the user comes back from the IDP".

We definitely have working IDP-hinting, so is this a vestige of some previous version of that? I'd want to be more sure of what's going on by looking at git history, rather than make changes to prod, but if this is necessary then fine... I suppose.

PR #869, opened 2020-06-04
Fix the Dockerfile and add some notes about using it

galund:
- update readme so that people know it's there and how to use it
 - still work to do, to make it easy to update dependencies inside the container


galund:
I didn't actually get the app _running_ yet - but the tests pass :-)

PR #871, opened 2020-06-08
Hub 636 change short hub logic and kick out pages

jkunle:
see: https://govukverify.atlassian.net/browse/HUB-636
**What**
Users who have passport and smartphone should be directed straight to the IDP picker (not one of the kick out pages as is currently the case).

The kick out pages also need updating with new content. The new content is here: Short hub logic - NFC updates  and you can see it in context of the new journey here: https://docs.google.com/drawings/d/1zCBOKqsLQSJQl5NNqWUzs5FhlY0EXoy6JOxfRtv--qM/edit

**Why**
Digi have introduced NFC functionality in their iOS and Android apps that allows users to verify using just a passport, if their smartphone is NFC-enabled.

As we cannot tell which users have NFC and we don’t want to work that out pre-IDP, the content on the kickout pages cannot say exactly what other documents a user would need to be able to verify. We’ve changed them to a ‘check your answers’ style page instead.


app/views/select_documents_variant_c/advice.html.erb
```diff
@@ -1,30 +1,34 @@
...
-                try_again_link_end: "</a>") %>
-    <% end %>
+      <% @documents.group_by(&:last).each do |key, value| %>
+        <div class="govuk-grid-column-one-half">
+          <h2 class="govuk-heading-s">
+            <%= t "hub_variant_c.select_documents_advice.advice_html.#{key ? 'specified' : 'unspecified'}" %>
+          </h2>
+          <ul id=<%="#evidence_#{key ? 'specified' : 'unspecified'}"%> class="govuk-list">
+            <% value.map(&:first).each do |evidence| %>
+              <li><%=evidence%></li><hr>
```
galund, line 19: Didn't know <hr> was even still a thing, but in any event it isn't allowed inside a <ul> https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element

config/locales/cy.yml
```diff
@@ -357,6 +262,15 @@ cy:
       title: Dod o hyd i’r gwasanaeth rydych yn ei ddefnyddio i ddechrau eto
       hint: Dilynnwch y cyfarwyddiadau rydych wedi’u derbyn.
       other_services: Gwasanaethau eraill
+    what_happens_next:
+      title: What happens next
+      heading: What happens next
+      what_happens: |
```
galund, line 268: in the english, this is called `what_happens_next_html` or something. Different anyway.

config/locales/cy.yml
```diff
@@ -139,115 +140,19 @@ cy:
...
+        unspecified: Things you do not have with you
+      what_to_do_next:
+        heading: What to do next
+        change_your_answers_link: change your answers
+        prove_your_identity_link: prove your identity another way
+        sub_heading_html: <p class="govuk-body">You can still %{service_name}.</p>
+        content_html: |
+          <p class="govuk-body">If you have more of these things with you, %{change_your_answers_link}.</p>
+          <p class="govuk-body">If you have more of them but they’re somewhere else, come back when you have them with you.</p>
+          <p class="govuk-body">If you do not have any more of them, %{prove_your_identity_link}.</p>
```
galund, line 155: This kinda of thing doesn't really work for translations, because it relies on the word order being the same for English as for any other language. It's better for each entire sentence to be a separate translation string.

So instead, `content_html` needs to contain a full sentence "If you do not have any more of them, prove your identity another way.", and then "Prove your identity another way" needs to also be available in a separate translation string for your special 'other ways' page (in a separate top level key).

Yes this results in repetition, but in the case of translated strings that's the right thing to do.

galund: Sorry, should have been more clear because of course there needs to be a link in there somewhere, so you end up with a sentence that looks like

If you do not have any more of them, %{prove_your_identity_link_start}prove your identity another way%{prove_your_identity_link_end}.

spec/features/user_visits_select_documents_variant_c_page_spec.rb
```diff
@@ -118,4 +124,8 @@ def expect_reporter_to_receive(evidence:, attempts:)
       evidence_list: evidence,
     )
   end
+
+  def evidence_chosen(*values)
```
galund, line 128: To be honest, I prefer the more verbose way - in principle functional tests are supposed to be human-readable (e.g. so that they can be verified by a product manager), so having the English strings repeated seems nicer.

But assuming we keep this function, it should be called `choose_evidence`, I think.

config/locales/cy.yml
jkunle: So what make other areas in cy.yml like different
```
content_html: |
          <p class="govuk-body">Mae cofrestru gyda GOV.UK Verify fel arfer yn cymryd tua 15 munud. Mae'n gweithio orau os oes gennych:</p>
          <ul class="govuk-list govuk-list--bullet">
            <li>cyfeiriad yn y DU</li>
            <li>pasport neu drwydded gyrru gyda llun dilys</li>
          </ul>
          <p class="govuk-body">Bydd darparwr hunaniaeth ardystiedig yn gwirio eich hunaniaeth pan fyddwch yn cofrestru gyda GOV.UK Verify. Maent i gyd wedi cwrdd â safonnau diogewlch a osodwyd gan y llywodraeth.</p>
```

spec/features/user_visits_select_documents_variant_c_page_spec.rb
jkunle: I'll take your recommendation here thanks.

app/views/select_documents_variant_c/advice.html.erb
JohnWattsGDS: This can probably do what you need: https://design-system.service.gov.uk/components/summary-list/

config/locales/cy.yml
JohnWattsGDS: I'm surprised that didn't break the app as that kind of means that one of the translations must be "missing" doesn't it?

app/views/select_documents_variant_c/advice.html.erb
jkunle: Thanks wasn't apparent to me

PR #873, opened 2020-06-12
BAU: Upgrade testing dependencies

CharlesIC:
We've been using pretty old libraries for our testing suite, some of them 4 years old.... I've upgraded the most important ones and make the necessary changes to make tests pass.

- Upgraded Capybara and Rspec to the latest versions. Fix tests that were broken as a result.
- Upgraded rubocop-govuk to a newer version. Updated style in a few places to match the new guidelines and appease rubocop.
- Removed no longer needed `.pkgr.yml` which was used by Jenkins.
- Fixed `proxy_node_error.html.erb` layout to use the correct content column width.
- Separated Rake methods into separate classes to avoid defining them in the tasks.


CharlesIC:
Follow-up PR based on @JohnWattsGDS 's comment #874

PR #875, opened 2020-06-15
Hub-518: crash on paused frontend endpoint

jkunle:
see: https://govukverify.atlassian.net/browse/HUB-518
From https://sentry.tools.signin.service.gov.uk/verify/prod-frontend/issues/4309/?query=is:unresolved
https://www.signin.service.gov.uk/paused
Api::Error: Unexpected error whilst trying to communicate wth the Hub. Received 404 with error message: 'NONE', type: 'NONE' and id: 'NONE', Referer: ''
The Hub may be unreachable. Check all services are running and are accessible

This is (now) an expected case: the user has a pause-and-resume cookie indicating an RP using an entity ID that no longer exists.

We are seeing these because of a change to HMRC’s entity id’s.

We should catch these exceptions and the user should get the same behaviour as when they go to that URI with no pause-and-resume cookie.

We think that the function where we might catch this error

### DIFF
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -115,11 +115,16 @@ private
     selected_rp = get_rp_details(last_rp)
     return if selected_rp == nil
 
-    @transaction = {
-      name: get_translated_service_name(selected_rp.simple_id),
-      homepage: selected_rp.transaction_homepage,
-      start_page: preferred_start_page(selected_rp),
-    }
+    begin
+      @transaction = {
+        name: get_translated_service_name(selected_rp.simple_id),
+        homepage: selected_rp.transaction_homepage,
+        start_page: preferred_start_page(selected_rp),
+      }
+    rescue StandardError => e
+      logger.warn e.message
+      nil
+    end
   end
 
   def get_rp_details(last_rp_value)
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -102,7 +102,7 @@ private
 
   def set_transaction_from_session
     selected_rp = get_rp_details(current_transaction_entity_id)
-    return if selected_rp == nil
+    return if selected_rp.nil?
 
     @transaction = {
       name: current_transaction.name,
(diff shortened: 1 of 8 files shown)


app/controllers/paused_registration_controller.rb
```diff
@@ -102,23 +102,28 @@ def with_cookie
...
   def set_transaction_from_session
     selected_rp = get_rp_details(current_transaction_entity_id)
-    if selected_rp != nil
-      @transaction = {
-        name: current_transaction.name,
-        homepage: current_transaction_homepage,
-        start_page: preferred_start_page(selected_rp),
-      }
-    end
+    return if selected_rp == nil
```
galund, line 105: I don't know if it's just a style thing or more significant, but I thought you were supposed to write `selected_rp.nil?`. I guess you reversed the existing thing so :shrug: :-)

app/controllers/paused_registration_controller.rb
```diff
@@ -102,23 +102,28 @@ def with_cookie
...
-    if selected_rp != nil
+    return if selected_rp == nil
+
+    begin
       @transaction = {
         name: get_translated_service_name(selected_rp.simple_id),
         homepage: selected_rp.transaction_homepage,
         start_page: preferred_start_page(selected_rp),
       }
+    rescue StandardError => e
```
galund, line 124: A catch-all error handler here feels like poor practice. If this is essential to fixing this ticket (and I don't understand why it would be, if we are escaping the function early), then is there a more specific exception we could catch?

app/controllers/paused_registration_controller.rb
```diff
@@ -102,23 +102,28 @@ def with_cookie
...
+    return if selected_rp.nil?
+
+    begin
       @transaction = {
         name: get_translated_service_name(selected_rp.simple_id),
         homepage: selected_rp.transaction_homepage,
         start_page: preferred_start_page(selected_rp),
       }
+    rescue Api::Error => e
+      logger.warn e.message
```
galund, line 125: Can we be clear about the circumstances in which `get_rp_details` can succeed, but `get_translated_service_name` can fail, so we end up here? If we don't know about any such circumstances (except for something exceptional, that we'd want to know about via Sentry) then I don't think this `rescue` should be here; if there are some, it probably merits a comment and/or an addition to the message that's being logged.

spec/controllers/paused_registration_controller_spec.rb
```diff
@@ -154,25 +154,47 @@
...
-      cookies.encrypted[CookieNames::VERIFY_FRONT_JOURNEY_HINT] = front_journey_hint_cookie.to_json
-      expect(subject).to redirect_to start_path
-    end
-
     it "should render error page when user has no session" do
       session.clear
       expect(subject).to render_template(:something_went_wrong)
     end
+
+    context "config service in unavailable" do
```
galund, line 162: There's a typo here (in->is), but maybe we can be more descriptive. The scenario isn't that the config service is entirely unavailable - just that the RP's data isn't in there. Perhaps "invalid or obsolete RP entity ID not present in config service" would do.

app/controllers/paused_registration_controller.rb
```diff
@@ -102,35 +102,33 @@ def with_cookie
...
 
-    begin
-      CONFIG_PROXY.get_transaction_details(last_rp_value)
-    rescue StandardError
-      logger.warn "Error trying to retrieve transaction details for #{last_rp_value}"
-      nil
-    end
+    CONFIG_PROXY.get_transaction_details(last_rp_value)
+  rescue Api::Error => e
+    logger.warn "Trying to retrieve transaction details for #{last_rp_value}. #{e}"
```
galund, line 130: The message needs to read "Error trying [etc]" otherwise it looks like it's just logging an _attempt_.

spec/controllers/paused_registration_controller_spec.rb
```diff
@@ -154,25 +154,47 @@
...
+
+      it "redirects user to start page when RP present in cookie and there is a SESSION_ERROR error" do
+        error_body = { clientMessage: "Failure", exceptionType: "EXPECTED_SESSION_STARTED_STATE_ACTUAL_IDP_SELECTED_STATE", errorId: "0", Referer: "" }
+        stub_missing_transaction_details(body: error_body, status: 400)
+
+        cookies.encrypted[CookieNames::VERIFY_FRONT_JOURNEY_HINT] = front_journey_hint_cookie.to_json
+        expect(subject).to redirect_to start_path
+      end
+
+      it "redirects user to start page when RP present in cookie and there is a SESSION_TIMEOUT error" do
```
galund, line 191: I don't quite follow why this scenario needs the context about the RP being an obsolete entity id?

PR #876, opened 2020-06-16
Hub 636 bau change short hub logic and kick out pages

jkunle:
These are changes to fix comments from Beth, in addition changed Summary list to Ul list due to HTML validation issue, where Summary list require 'dt' tag

Some of the kick out pages have the columns the wrong way round. These are the ones for:

- driving licence + phone
- driving licence + card
-  phone + card

1. They all have the 'things you do not have with you' column first. This is confusing if a user has been back and forth a couple of times, and are expecting that column to be the same as the other pages.

2. The 'none of the above' version of the kick out page does not include the correct content - it should be like the 'none of the above' row of the table in this document: https://docs.google.com/document/d/1wDXEkqigqcy4q5-_xhFffO6kFCFdR559i-_gKuhVR8Q/edit#heading=h.p5rhhpvlqt6k

3. The ‘other ways' page includes the 'Other ways' h2 which should not be there, like the design: https://docs.google.com/drawings/d/1zCBOKqsLQSJQl5NNqWUzs5FhlY0EXoy6JOxfRtv--qM/edit
(Sorry if the design wasn’t clear enough here!)

4. The 'other ways page needs full stop after the service name in the first sentence.

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -142,7 +142,7 @@ cy:
       advice_html:
         heading: Based on your answers, you cannot use GOV.UK Verify
         sub_heading: Make sure you selected all the things you have with you.
-        none_heading_html: |
+        no_evidence_intro_html: |
           <p class="govuk-body">To set up an identity account, you need to provide evidence for the companies to check.</p>
           <p class="govuk-body govuk-!-margin-bottom-0">You said you have none of these things with you right now:</p>
         specified: Things you have with you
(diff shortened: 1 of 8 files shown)


config/locales/cy.yml
```diff
@@ -142,6 +142,9 @@ cy:
       advice_html:
         heading: Based on your answers, you cannot use GOV.UK Verify
         sub_heading: Make sure you selected all the things you have with you.
+        none_heading_html: |
```
galund, line 145: Feels like it's an 'intro' rather than a 'heading' :-)

jkunle: What would be a nice name?

PR #877, opened 2020-06-19
HUB-667 Fixed missing journey_type being passed to event system

j-maynard:
Fixed missing journey_type and also fixed the tests which broke as a result.

JIRA Ticket: https://govukverify.atlassian.net/browse/HUB-667


galund:
What happened to removing those methods that we worked out were unused?

Also: I'm not clear why we would need to set this in more than one place.

Also: as there's now a constant, can we use it instead of the raw string?

spec/features/idp_selections_reported_to_piwik_spec.rb
```diff
@@ -103,3 +110,16 @@ def stub_idp_select_request(idp_entity_id)
...
+
+def stub_idp_select_request_with_journey_type(idp_entity_id)
+  stub_session_select_idp_request(
+    encrypted_entity_id,
+    PolicyEndpoints::PARAM_SELECTED_ENTITY_ID => idp_entity_id,
+    PolicyEndpoints::PARAM_PRINCIPAL_IP => originating_ip,
+    PolicyEndpoints::PARAM_REGISTRATION => true,
+    PolicyEndpoints::PARAM_REQUESTED_LOA => "LEVEL_2",
+    PolicyEndpoints::PARAM_PERSISTENT_SESSION_ID => instance_of(String), # no longer comes from matomo
+    PolicyEndpoints::PARAM_JOURNEY_TYPE => instance_of(String),
```
jkunle, line 122: I think you'll still want to be passing the explicit value of JOURNEY_TYPE because it is the expected value when it is used. unlike PERSISTENT_SESSION_ID which can change. I'll suggest 
```
def stub_idp_select_request(idp_entity_id, journey_type: nil)
  stub_session_select_idp_request(
    encrypted_entity_id,
    PolicyEndpoints::PARAM_SELECTED_ENTITY_ID => idp_entity_id,
    PolicyEndpoints::PARAM_PRINCIPAL_IP => originating_ip,
    PolicyEndpoints::PARAM_REGISTRATION => true,
    PolicyEndpoints::PARAM_REQUESTED_LOA => "LEVEL_2",
    PolicyEndpoints::PARAM_PERSISTENT_SESSION_ID => instance_of(String), # no longer comes from matomo
    PolicyEndpoints::PARAM_JOURNEY_TYPE => journey_type,
    PolicyEndpoints::PARAM_VARIANT => nil,
  )
end
```
Then you can `stub_idp_select_request(idp_1_entity_id, journey_type: sign-in-last-successful-idp )`

j-maynard: We're not specifically testing for the journey_type... The jounrey_type now comes along for the ride on this test.

app/controllers/authn_request_controller.rb
```diff
@@ -103,9 +103,6 @@ def redirect_for_journey_hint(hint)
         do_eidas_sign_in_redirect(redirect_params)
       when "submission_confirmation"
         redirect_to confirm_your_identity_path(redirect_params)
-      when successful_idp_hint?
```
galund, line 106: Can we remove the method `successful_idp_hint` or is it somehow still used elsewhere?

app/controllers/partials/idp_selection_partial_controller.rb
```diff
@@ -4,6 +4,7 @@ module IdpSelectionPartialController
   include JourneyHintingPartialController
 
   def set_journey_hint_followed(entity_id)
+    session[:journey_type] = "sign-in-last-successful-idp"
```
galund, line 7: Does this actually do anything, seeing as it should've been set already?

app/controllers/partials/journey_hinting_partial_controller.rb
```diff
@@ -63,6 +63,7 @@ def decorate_idp_by_simple_id(providers, simple_id)
   def try_render_journey_hint
     journey_hint_entity_id = success_entity_id
     unless journey_hint_entity_id.nil?
+      session[:journey_type] = "sign-in-last-successful-idp"
```
galund, line 66: Can we use the constant please?

spec/features/idp_selections_reported_to_piwik_spec.rb
galund: In which case isn't the name of the function now wrong? I'm pretty confused about what's going on here.

app/controllers/partials/idp_selection_partial_controller.rb
j-maynard: It captures where the Ajax is used.  It needs to be here as well.

spec/features/idp_selections_reported_to_piwik_spec.rb
j-maynard: We have two methods to stub two different instance of the same request.  The stubbing framework expects specific requests to be made... any deviation the registered requests are ignored which is why we now have two methods.

The first method is the original method which doesn't specify a journey_type. So any requests where the jouney_type is nil are registered.

We have added an additional method which adds the journey_type with a string in.  Because this test isn't about exercising the journey_type in any meaningful way we only care that a string is present.  As I pointed out to Ola this allows use to register this stub and it covers any instance where the journey_type string is present.  As I speculated, we're almost certainly going to end up with a Journey Type present for the ShortHub A/B/C test in the coming weeks.

.generators
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Settings><!--This file was automatically generated by Ruby plugin.
```
galund, line 2: "Ruby plugin" for what? Not sure this should be checked in. (And if it should be, it deserves its own commit.)

.generators
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jkunle, line 1: I don't think you mean't to check these in `.generator` or `.rakeTasks`

app/controllers/authn_request_controller.rb
galund, line 106: There's still a function to be deleted here (viz `successful_idp_hint`)

app/controllers/partials/idp_selection_partial_controller.rb
```diff
@@ -4,6 +4,7 @@ module IdpSelectionPartialController
   include JourneyHintingPartialController
 
   def set_journey_hint_followed(entity_id)
+    session[:journey_type] = SIGN_IN_LAST_SUCCESSFUL_IDP
```
galund, line 7: Nope, still not getting we need this. Let's do a hangout.

spec/features/idp_selections_reported_to_piwik_spec.rb
jkunle: You can still have just one method for this if you replace
`stub_idp_select_request_with_journey_type("other-entity-id")`
with
`stub_idp_select_request("other-entity-id", journey_type: instance_of(String))`
And alter stub_idp_select_request above to an `optional` as above

galund: Don't get why we need a version that doesn't include the journey type. Why not make the general method include some journey type rather than make a new one? These tests aren't supposed to be testing the other microservices (are they?)

spec/features/user_submits_start_page_form_spec.rb
```diff
@@ -3,6 +3,28 @@
 require "api_test_helper"
 
 RSpec.describe "when user submits start page form" do
+
+  def then_im_at_the_idp(ab_value: nil, journey_type: "sign-in")
```
galund, line 7: If this is a copy of the other one then can we DRY it into feature_helpers perhaps?

spec/features/user_visits_sign_in_page_spec.rb
```diff
@@ -113,12 +89,24 @@ def expect_to_have_updated_the_piwik_journey_type_variable
       allow_any_instance_of(UserCookiesPartialController)
         .to receive(:ab_test_with_alternative_name).and_return(nil)
     end
+
+    it "will redirect the user to the IDP" do
```
galund, line 93: This seems to introduce a duplication (see immediately below). Aren't we trying to move this to some 'helpers' module?

JohnWattsGDS: That's a completely duplicated test for what reason, I don't know.  There were a couple of 'steps' that were duplicated across tests that we've moved to a helper.

spec/features/user_submits_start_page_form_spec.rb
```diff
@@ -53,4 +68,19 @@
     click_button("next-button")
     expect(page).to have_content t("hub.start.error_message")
   end
+
+  it "will redirect to the IDP when the user chooses a hinted IDP", js: true do
+    allow_any_instance_of(UserCookiesPartialController)
+    .to receive(:ab_test_with_alternative_name).and_return(nil)
```
galund, line 74: I'm not clear why this is here, it looks like it should be in the a/b test related tests only?

JohnWattsGDS: @jkunle - can you remember?

JohnWattsGDS: That mock / stub is used in multiple places in the codebase.  It appears that in tests the default behaviour for that method is to return `short_hub_2019_q3-preview_control_a` (or whatever the current AB Test is I guess).  It was introduced as part of HUB-504.  We set it explicitly to return nil to avoid having to figure which, if any ab_test is currently in use as the method is called during the calls to policy and so we need to have the correct call stubbed.

jkunle: B'cos of `then_im_at_the_idp ` which makes a call the policy endpoint that requires `PolicyEndpoints::PARAM_VARIANT => ab_value`.  The `ab_value` is not `nil`, the alternative is put the expected value but this would make the test more brittle. Each time a new experiment or ab test is set up with a different name in the main_route the tests would break.

galund: OK, I guess I'm confused why it's not in the "will redirect the user to the IDP" test then. But OK.

jkunle:
@galund it actually is, just placed in a before(:each) see https://github.com/alphagov/verify-frontend/blob/master/spec/features/user_visits_continue_to_your_idp_page_spec.rb#L27

PR #884, opened 2020-07-17
BAU Updated to Ruby 2.6.6

j-maynard:
A few version bumps in the Gemfile as well as the bump in ruby.  Some of
this work is to get ready for the eventual upgrade to Rails 6.0 as well
as keeping things up to date.

Also pegged the version of things in the stub/api/Gemfile as they were
now pull things for Rails 6 and as a result pre-commit was failing.

We should probably audit the version of things in the Gemfile.

This also fixes the Dependabot issue with Rack.


jkunle:
Seems there still more to do here
 # https://github.com/alphagov/verify-infrastructure-config/blob/339dab2/pipelines/deployer/deploy-verify-hub.yml#L1002
# https://github.com/alphagov/verify-infrastructure-config/blob/339dab2/pipelines/deployer/deploy-verify-hub.yml#L1107#

PR #885, opened 2020-07-22
HUB-679 Fix for the SameSite cookie issue

j-maynard:
Undid my previous versions and settled on using this Gem which should do what we need it to.  The Gem's homepage can be found at:

https://github.com/pschinis/rails_same_site_cookie

Jira Ticket: https://govukverify.atlassian.net/browse/HUB-679

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,12 @@ gem 'multi_json'
 
 gem 'browser'
 
+# SameSite is now a thing in Google Chrome and Chromium browsers which 
+# has started causing issues for endusers.  This gem works round the
+# issue.  See Chromium issue: https://www.chromium.org/updates/same-site
+
+gem 'rails_same_site_cookie', :git => "https://github.com/alphagov/rails_same_site_cookie.git", :ref => "704c1958bf2518ba8248fe3d21a49361e38e911a"
+
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -43,6 +43,10 @@ gem 'multi_json'
 
 gem 'browser'
 
+# SameSite is now a thing in GoogleChrome this gem works round the
+# issue.  See JIRA Ticket HUB-679
```
galund, line 47: Can you put a link to Google's explanation page, instead of to our ticket? People can get our ticket number from the git history if they really want it.

galund:
Ticket URL or number in the commit message please!

Gemfile
```diff
@@ -43,6 +43,12 @@ gem 'multi_json'
 
 gem 'browser'
 
+# SameSite is now a thing in Google Chrome and Chromium browsers which 
+# has started causing issues for endusers.  This gem works round the
+# issue.  See Chromium issue: https://www.chromium.org/updates/same-site
+
+gem 'rails_same_site_cookie', :git => "https://github.com/alphagov/rails_same_site_cookie.git", :ref => "704c1958bf2518ba8248fe3d21a49361e38e911a"
```
CharlesIC, line 50: This should be `https://github.com/alphagov/rails-same-site-cookie` - yes GH redirects but we should use the correct repo names in the source code

PR #887, opened 2020-07-22
2nd Attempt to fix...

j-maynard:
Sorry about this...  This was the second suggested fix for the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,9 @@
 FROM ruby:2.6.6
 
 RUN apt-get update && apt-get install -y firefox-esr nodejs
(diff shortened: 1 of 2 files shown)


j-maynard:
> Why did the SSL config move?

It didn't...  Seems to be a debian issue which affects PhantomJS

CharlesIC:
the env var was `/etc/ssl/` and now it's `/tmp/libssl.conf` so I'm wondering why it needs to point to a different place

j-maynard:
So the new commit points to an empty file... the issue is apparently caused by phantom not being able to read the file.

PR #892, opened 2020-08-18
Ensure Pwiki SameSite cookie behaviour

j-maynard:
This commit updates the piwik JavaScript to the latest version found at https://github.com/matomo-org/matomo/blob/4.x-dev/js/piwik.min.js and also adds the code to set the SameSite cookie behaviour.

As the new Matomo Javascript changes the interactions slightly and uses post requests rather than get requests this also updates some of the routes to support making post requests to the matomo stub we have in our code base.

PR #894, opened 2020-09-01
Hub-679 resolve same site warnings

jkunle:
See https://github.com/alphagov/verify-frontend/pull/892 (That PR used a higher version of piwik.js i.e. 4.x-dev)

Updated piwik.js to version **3.13.5** to keep in sync with the version on our server instance  https://analytics.tools.signin.service.gov.uk. 
This is done by switching the branch to version 3.13.5 on https://github.com/matomo-org/matomo
the copy of the javascript is taken from https://raw.githubusercontent.com/matomo-org/matomo/3.13.5/js/piwik.min.js
Matomo expectation is that we use https://analytics.tools.signin.service.gov.uk/piwik.js on our analytics server. 

Also set secure cookie in piwik_secure_cookie.js this is necessary since some of our tracking is done in client-side javascript
i.e piwik.js and also track other behaviours from ruby. 
see: https://matomo.org/blog/2020/02/new-cookie-behaviour-in-browsers-may-cause-regressions/

Unfortunately, it is not easy to test this; it requires having local server running under https; it is possible but would be too time consuming. We try this out in staging and see how it goes


instantiator (dismissed):

instantiator:
Ola - before you merge, please make sure the integration and prod deployments are paused.

instantiator (dismissed):

JohnWattsGDS (approved):

instantiator (approved):

idrop (approved):

PR #895, opened 2020-09-01
[DO NOT MREGE] test to check if  firefox is causing an issue with travis/selenium

idrop:


instantiator:
Should we use [extended support release](https://www.whatismybrowser.com/guides/the-latest-version/firefox) `78.2.0`?

PR #896, opened 2020-09-02
BAU run pre-commit script in docker

idrop:
Run `./pre-commit.sh` within a docker container.
This allows us to run acceptance tests against `firefox-esr` rather than later versions of firefox which do not currently work well with geckodriver.

Get travis to run tests against  v78 of firefox, which is known to work.


jkunle (approved):
🥇

instantiator (approved):

PR #898, opened 2020-09-11
BAU Repair the ignore hint route for LOA1

instantiator:
* Amends the routes, pulling the `start_ignore_hint` route out of the LOA2 constraint, and up to root-level.
* Adds `default_session_loa1` to `feature_helper.rb` so as to make it trivial to test under LOA1 circumstances.
* Adds 2 tests to `user_submits_start_page_form.spec` which, technically, isn't quite the right place because this isn't a form submission, it's a simple anchor click - but it still seems like the most theme-appropriate place to put them. They test that the ignore-hint link correctly takes the user to the start page, rather than the 404 (as seen previously).

_Also manually tested without the route fix to show that under LOA1 circumstances, the routing fails and the user sees a 404._


j-maynard (approved):
:lion:

jkunle (approved):
LGTM

PR #902, opened 2020-10-21
HUB-533 Add release information for sentry

j-maynard:


Dockerfile
```diff
@@ -24,4 +24,8 @@ RUN bundle exec rake assets:precompile
 RUN touch /tmp/libssl.conf
 ENV OPENSSL_CONF /tmp/libssl.conf
 
+# Get the current git head reference and set it as
+# release for Sentry
+RUN export RELEASE_VER=$(git rev-parse HEAD)
```
idrop, line 29: Should we consider the  rev-parse `--short` flag for log brevity?

j-maynard: They say you need to make the release unique across the organisation... So my gut is to go with the full sha as thats pretty much guaranteed to be unique

AmritSidhu (approved):

jkunle (approved):

PR #904, opened 2020-11-10
HUB-720 Align the ab_test.yml with prod

j-maynard:
This file hasn't been maintained but it is used by verify-local-startup and should be maintained, ideally with the same config we have in production.


galund (approved):
Yup, hopefully the short hub stuff is getting cleaned up real soon now too :-)

jkunle (approved):

PR #906, opened 2020-11-13
Hub-737: Tear down short-hub abtest and keep the variant

jkunle:
**What**
We need to remove the short hub AB test and keep the variant.
https://govukverify.atlassian.net/browse/HUB-737
see: https://github.com/alphagov/verify-frontend-federation-config/pull/184
see: https://github.com/alphagov/verify-acceptance-tests/pull/106

**Why**

The variant won! We've set it to 100% so all users are seeing it, but now we need to do the cleanup.

**Description**
Current control artefacts i.e. controllers, views was renamed, the variant is then re-named to the control, the Controllers and views that are associated with the **long hub journey**  below has been removed in favour of the short-hub journey.
This is easier to review commit by commit

image

image

The Redirect to warning controller  below has been removed

image

Throttling functionality has been remove, there may still be some artefacts of it left in the config

The about pages ie. the educational pages for the LOA1 now uses the short-hub combined about page,
and now use single controller and view.

The old idp rules have been removed in favour of the short idp rules.

All traces of ABTest experiments have been removed from controller specs


spec/models/select_documents_form_spec.rb
```diff
@@ -1,128 +0,0 @@
-require "spec_helper"
```
Wynndow, line 1: Deleting this spec file means that `app/models/select_documents_form.rb` now doesn't have any tests. I know the model is simpler but could you add something to cover it?

app/constraints/select_route.rb
```diff
@@ -30,6 +28,8 @@ def cookie_matches_experiment?(request)
     end
   end
 
+private
```
Wynndow, line 31: `cookie_matches_experiment?` only seems to be used in this  class, so could stay private?

Wynndow, line 31: `cookie_matches_experiment?` only seems to be used in this  class, so could stay private?

config/locales/cy.yml
```diff
@@ -184,25 +185,24 @@ cy:
...
-          <li>ydych dros 20 oed</li>
-          <li>oes gennych o leiaf un ddogfen hunaniaeth â llun, o unrhyw wlad</li>
+          <li>have a UK address</li>
+          <li>are over 20</li>
+          <li>have at least one photo identity document, from any country</li>
           </ul>
-        button: Defnyddiwch GOV.UK Verify
-      cannot_verify: Beth i'w wneud os na allwch ddefnyddio GOV.UK Verify
+        button: "Use GOV.UK Verify"
+      cannot_verify: What to do if you cannot use GOV.UK Verify
```
Wynndow, line 205: This content seems to have become English - should probably stay as Welsh

Wynndow (changes requested):
I don't have much experience with the AB stuff but the changes seem fine as far as I can tell. Only a few comments on some minor stuff - the translation stuff should probably be fixed up though.

spec/models/select_documents_form_spec.rb
jkunle: Thanks

config/locales/cy.yml
jkunle: Many thanks for spotting this

Wynndow (approved):

JohnWattsGDS (approved):

PR #917, opened 2020-12-07
Fixed Startup Script

j-maynard:
Changes to Puma means the `-d` option has been deprecated.  As an alternative I have remove the use of `-d` in favour of `&`

This should fix things for everyone.

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -18,4 +18,4 @@ then
 fi
 
 bundle check || bundle install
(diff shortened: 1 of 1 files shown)


idrop (approved):
That works for me

instantiator (approved):

CharlesIC:
Please review #918 instead

PR #918, opened 2020-12-08
BAU: Fix local startup after Puma upgrade

CharlesIC:
- Remove the `-d`(aemonize) option removed in version 5. Use the bash background task management instead. Append `&` to run the server in background when frontend is run as part of the full hub configuration. If running as stub API, run in foreground so it can be stopped by pressing Ctrl+C

- Create the `tmp` folder if it doesn't exist. Puma/Rake needs this folder to hold its state. If it doesn't exist, the process will quit immediately without printing any error message

- Don't print error message from `rm` when killing the service. Running `kill` removes the pid file under certain conditions

- Reorder gems in the Gemfile to be in alphabetical order to appease Rubocop
- Exclude the Gemfile from the `StringLiterals` Rubocop check

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -1,12 +1,17 @@
 #!/usr/bin/env bash
 
+set -e
 cd "$(dirname "$0")"
 
 ./kill-service.sh
 
+RUN_IN_BACKGROUND=true
+if ! [ -d tmp ]; then mkdir tmp; fi
+
 if [ "$1" == '--stub-api' ]
 then
   echo "Starting stub-api server on port 50199"
+  unset RUN_IN_BACKGROUND
   export CONFIG_API_HOST=http://localhost:50199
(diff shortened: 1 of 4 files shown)


instantiator (approved):

idrop (approved):

startup.sh
```diff
@@ -1,12 +1,17 @@
 #!/usr/bin/env bash
 
+set -e
 cd "$(dirname "$0")"
 
 ./kill-service.sh
 
+RUN_IN_BACKGROUND=true
+if ! [ -d tmp ]; then mkdir tmp; fi
```
galund, line 9: `mkdir -p` probably does what you want without the `if`, for next time :-)

CharlesIC: Ah you're right!

PR #920, opened 2020-12-09
BAU-2024 Make sure logging is initialized first

JohnWattsGDS:
We want to log the value of EIDAS_DISABLED_AFTER once and only once.  Our first attempt was to log this when the configuration initializer was run but we couldn't see the output in Splunk even though "it works on my machine"™.  I'm taking a punt on this being because the logging isn't configured when️ the configuration initializer is run so I've renamed some files so that logging is the first thing that's configured.

I've also removed our log spammer that logged the value of EIDAS_DISABLED_AFTER every time a method was called.

"It works on my machine"™


CharlesIC:
Huh, I've always wondered why those files were named xx_something. Now I know! 👍 
Neat solution 👌

CharlesIC (approved):

idrop (approved):

PR #923, opened 2021-01-13
HUB-783: Failed registration bug

Wynndow:
The journey now looks like this:

failed_registration_2
### HUB-783: Don't return failed IDPs from recommendations engine
This fixes a bug in the frontend. If a user failed to register with an
IDP and their AuthnResponse was returned with a "FAILED" state, they
would be directed via a landing page and a link back to the choose a
company page.

The logic on that page would retrieve a list of IDPs from the
recommendation engine, including the IDP they just failed with. If they
the selected that IDP again, it's entity ID would be compared to a
separate list of available IDPs - this time with any IDPs the user had
already failed with removed. This caused an error.

This change ensures the recommendation engine doesn't return an IDP the
user has failed with, and so it's not presented to them on the choose an
IDP page.

### HUB-783: Remove IDP recommendation where not needed
The rendered partial never uses it, so there is no need to pass it in.

### DIFF
--- a/app/controllers/choose_a_certified_company_about.rb
+++ b/app/controllers/choose_a_certified_company_about.rb
@@ -8,7 +8,6 @@ module ChooseACertifiedCompanyAbout
     matching_idp = current_available_identity_providers_for_registration.detect { |idp| idp.simple_id == simple_id }
     @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(matching_idp)
     if @idp.viewable?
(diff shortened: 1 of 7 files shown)


instantiator (dismissed):

instantiator:
LGTM, but I imagine I'll have to re-review anyway once the linting's fixed and Travis accepts it :)

JohnWattsGDS (approved):

instantiator (approved):

PR #928, opened 2021-01-17
Make skiplink anchor link to main content area

tombye:
When I was logging into Verify I noticed the skiplink on the /start page doesn't work because it's URL doesn't point to the id of the main content area.

This should fix it for that and other pages I think. I chose to change the skiplink URL rather than the main content area id because:
- www.gov.uk uses that id for its one
- I can't find anything in this repository using the 'main-content' id

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -30,7 +30,7 @@
 
     <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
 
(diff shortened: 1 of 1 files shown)


CharlesIC (approved):
This indeed should be `content`
https://github.com/alphagov/verify-frontend/blob/9b247b10642b7360d45cc11f3ec3ea7218014413/app/views/layouts/application.html.erb#L75

Well spotted!

Wynndow (approved):

PR #931, opened 2021-01-20
HUB-740: Handle paused journeys for sign in only IDPs

Wynndow:
We get a few errors coming through for users attempting to restart
journeys with IDPs that are no longer providing registrations. The
cookie that informs this last for 18 months.

This will show users a new page if their paused journey is for an IDP
that no longer offers registrations and advise them to go back to their
service and try again.

This PR needs a content review as well as Welsh translations.

### Screenshot of new page.

Screenshot 2021-01-27 at 12 53 44

### Screenshots of existing page
This new page is a new variant for an existing endpoint. The two screenshots below show what happens if the IDP is still offering registrations - the difference between them is whether the user had a session or not at the time of the page load. The new page is for when the IDP is no longer offering registrations.

Screenshot 2021-01-20 at 11 14 34
Screenshot 2021-01-20 at 11 15 18

### DIFF
--- a/app/controllers/partials/journey_hinting_partial_controller.rb
+++ b/app/controllers/partials/journey_hinting_partial_controller.rb
@@ -44,6 +44,11 @@ module JourneyHintingPartialController
     last_status_value.nil? ? nil : last_status_value.fetch("RP", nil)
   end
 
+  def last_verify_journey_type
+    last_status_value = last_status
+    last_status_value.nil? ? nil : last_status_value.fetch("VERIFY_JOURNEY_TYPE", nil)
+  end
+
   def user_followed_journey_hint(entity_id_followed_by_user)
     hinted_id = success_entity_id
     !hinted_id.nil? && hinted_id == entity_id_followed_by_user
--- a/app/controllers/partials/user_cookies_partial_controller.rb
+++ b/app/controllers/partials/user_cookies_partial_controller.rb
@@ -36,6 +36,7 @@ module UserCookiesPartialController
     journey_hint_by_status_value["STATE"] = { IDP: idp_entity_id,
                                               RP: rp_entity_id.nil? ? session[:transaction_entity_id] : rp_entity_id,
                                               STATUS: status }
+    journey_hint_by_status_value["STATE"][:VERIFY_JOURNEY_TYPE] = session[:journey_type] if status == "PENDING"
 
     cookies.encrypted[CookieNames::VERIFY_FRONT_JOURNEY_HINT] = { value: journey_hint_by_status_value.to_json, expires: 18.months.from_now }
   end
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -95,10 +95,10 @@ private
   def with_cookie
     set_transaction_from_cookie
 
-    if (idp = last_idp_from_those_offering_registration)
+    if (idp = get_idp_from_idps_available_for_registration)
       @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
       render :with_user_session
-    elsif (idp = last_idp_from_those_offering_sign_in)
+    elsif (idp = get_idp_from_idps_available_for_sign_in)
       @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
       render :idp_no_longer_providing_registrations
     else
@@ -155,11 +155,11 @@ private
     list.idps
   end
 
-  def last_idp_from_those_offering_sign_in
+  def get_idp_from_idps_available_for_sign_in
     get_idp_choice(get_idp_list_for_sign_in, last_idp)
   end
 
-  def last_idp_from_those_offering_registration
+  def get_idp_from_idps_available_for_registration
     get_idp_choice(get_idp_list_for_registration(last_rp), last_idp)
   end
 
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -94,14 +94,15 @@ private
 
   def with_cookie
     set_transaction_from_cookie
-    enabled_idp_list = get_idp_list(last_rp)
-    idp = get_idp_choice(enabled_idp_list, last_idp)
-    @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
-    if not @idp.viewable?
-      logger.error "IDP from cookie not viewable. IDP found was '#{@idp}' from enabled IDP list '#{enabled_idp_list}'. Journey hint value: '#{journey_hint_value}'. Referer: '#{request.referer}'"
-      render :without_user_session
-    else
+
+    if (idp = last_idp_from_those_offering_registration)
+      @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
       render :with_user_session
+    elsif (idp = last_idp_from_those_offering_sign_in)
+      @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
+      render :idp_no_longer_providing_registrations
+    else
+      raise(Errors::WarningLevelError, "IDP not found in those providing registration or sign in. IDP was '#{last_idp}'")
     end
   end
 
@@ -140,11 +141,26 @@ private
     CONFIG_PROXY.get_transaction_translations(simple_id, params[:locale]).fetch(:name, nil)
   end
 
-  def get_idp_list(transaction_id)
+  def get_idp_list_for_registration(transaction_id)
     list = CONFIG_PROXY.get_available_idp_list_for_registration(transaction_id, "LEVEL_2")
     return nil if list.nil?
 
     list.idps
+  end
+
+  def get_idp_list_for_sign_in
+    list = CONFIG_PROXY.get_idp_list_for_sign_in(last_rp)
+    return nil if list.nil?
+
+    list.idps
+  end
+
+  def last_idp_from_those_offering_sign_in
+    get_idp_choice(get_idp_list_for_sign_in, last_idp)
+  end
+
+  def last_idp_from_those_offering_registration
+    get_idp_choice(get_idp_list_for_registration(last_rp), last_idp)
   end
 
   def preferred_start_page(selected_rp)
--- a/app/controllers/paused_registration_controller.rb
+++ b/app/controllers/paused_registration_controller.rb
@@ -95,7 +95,10 @@ private
   def with_cookie
     set_transaction_from_cookie
 
-    if (idp = get_idp_from_idps_available_for_registration)
+    if last_verify_journey_type == JourneyType::Verify::SIGN_IN
+      @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(get_idp_from_idps_available_for_sign_in)
+      render :with_user_session
+    elsif (idp = get_idp_from_idps_available_for_registration)
       @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
       render :with_user_session
     elsif (idp = get_idp_from_idps_available_for_sign_in)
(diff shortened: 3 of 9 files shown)


app/controllers/paused_registration_controller.rb
```diff
@@ -139,13 +141,28 @@ def get_translated_service_name(simple_id)
...
   end
 
+  def get_idp_list_for_sign_in
+    list = CONFIG_PROXY.get_idp_list_for_sign_in(last_rp)
+    return nil if list.nil?
+
+    list.idps
+  end
+
+  def last_idp_from_those_offering_sign_in
```
CharlesIC, line 158:
```suggestion
  def last_idp_offers_sign_in
```
Make method name more clear

app/controllers/paused_registration_controller.rb
```diff
@@ -139,13 +141,28 @@ def get_translated_service_name(simple_id)
...
+    get_idp_choice(get_idp_list_for_sign_in, last_idp)
+  end
+
+  def last_idp_from_those_offering_registration
```
CharlesIC, line 162:
```suggestion
  def last_idp_offers_registration
```
Make method name more clear

config/locales/en.yml
```diff
@@ -599,6 +599,10 @@ en:
           <p class="govuk-body">You'll need the email or username and password you used to create an account with %{idp_name}.</p>
         continue_verifying: 'Continue verifying with %{idp_name}'
         return_html: 'Or return to <a href="%{rp_start_page}">%{rp_name}</a>.'
+      idp_no_longer_providing_registrations:
+        title: "Verification Paused %{idp_name}"
+        heading: "%{idp_name} are no longer providing registrations"
```
CharlesIC, line 604:
```suggestion
        heading: "You can no longer create an identity account with %{idp_name}"
```
This is a variation of the text on the `/about` page
> You create an identity account with a company that meets government privacy and security standards.

This should be reviewed by @sabrinaharris for final approval

config/locales/en.yml
```diff
@@ -599,6 +599,10 @@ en:
...
+        return_html: 'You can return to <a href="%{rp_start_page}">%{rp_name}</a> and register with a different certified company.'
```
CharlesIC, line 605:
```suggestion
        return_html: 'You can try again to <a href="%{rp_start_page}">%{rp_name}</a> and register with a different certified company.'
```
The `name` property in RP config is always an action (verb), see [example](https://github.com/alphagov/verify-hub-federation-config/blob/3858195a3da3f8ef59ef6e41bb7e9e8dd1e018b3/configuration/display-locales/transactions/dvla-view-driving-record.yml#L5)

This also needs @sabrinaharris 's review

app/controllers/paused_registration_controller.rb
```diff
@@ -94,14 +94,16 @@ def with_session
...
   def with_cookie
     set_transaction_from_cookie
-    enabled_idp_list = get_idp_list(last_rp)
-    idp = get_idp_choice(enabled_idp_list, last_idp)
-    @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
-    if not @idp.viewable?
-      logger.error "IDP from cookie not viewable. IDP found was '#{@idp}' from enabled IDP list '#{enabled_idp_list}'. Journey hint value: '#{journey_hint_value}'. Referer: '#{request.referer}'"
-      render :without_user_session && return
+
+    if (idp = last_idp_from_those_offering_registration)
```
JohnWattsGDS, line 98: I'm assuming `idp = blah` and not `idp == blah` because of some evil ruby assign and evaluate the assigned value shenanigans.

Wynndow: I want to assign `idp` and then trigger that `if` branch if it's assigned to something truthy. That way I can use the assigned `idp`. 
The result of assigning idp is "truthy" unless it's assigned to nil - which will happen if the idp isn't offering registrations.

app/controllers/paused_registration_controller.rb
Wynndow: That's not really what the method is doing though. That name change implies that a boolean is going to be returned. It's actually returning the IDP.

config/locales/en.yml
Wynndow: Yep, Sabrina is going to take the content to the content folks.

config/locales/en.yml
Wynndow: As above

app/controllers/paused_registration_controller.rb
Wynndow: As above

app/controllers/paused_registration_controller.rb
CharlesIC: If I understand correctly, the `last_idp_from_those_offering_registration` filters a list and returns a value or nil. If the value is not nil, then `idp` gets assigned that value, otherwise the branch is not executed and the `elsif` branch gets a go.

It would be better to use a method here that returns a boolean, the Ruby trickery here is not easy to follow

JohnWattsGDS: I figured as much.  When in Rome etc.  Just another reason to dislike Ruby for me :-)

app/controllers/paused_registration_controller.rb
CharlesIC: But the current name implies that it return a boolean too. I think this logic could be split in two to make it easier to follow. We're doing two things - checking that last IDP offers sign in or registration, and then retrieving that IDP

app/controllers/paused_registration_controller.rb
Wynndow: That's correct. The reason for doing it this way was to reduce the number of nested if statements that resulted (what if the IDP isn't found in the IDPs offering sign in) and to reduce the number of API calls. If we return a boolean then we have to go call config again to get the actual IDP. Or make the calls up front and store the list, but that's going to result in a more complex if statement.

app/controllers/paused_registration_controller.rb
Wynndow: I don't feel the name does imply that. `last_idp` is a construct used throughout this controller, and it's being selected from those offering sign in.

Either way - i'll rewrite this logic to make it clearer.

app/controllers/paused_registration_controller.rb
Wynndow: OK - have pushed up a change to hopefully make it more readable. I've not gone as far as creating new methods to return booleans to use in the conditionals as the logic was because they'd just be wrappers around the methods I've used.

app/controllers/paused_registration_controller.rb
```diff
@@ -94,14 +94,16 @@ def with_session
...
   def with_cookie
     set_transaction_from_cookie
-    enabled_idp_list = get_idp_list(last_rp)
-    idp = get_idp_choice(enabled_idp_list, last_idp)
-    @idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(idp)
-    if not @idp.viewable?
-      logger.error "IDP from cookie not viewable. IDP found was '#{@idp}' from enabled IDP list '#{enabled_idp_list}'. Journey hint value: '#{journey_hint_value}'. Referer: '#{request.referer}'"
-      render :without_user_session && return
+
+    if get_idp_from_those_available_for_registration
```
CharlesIC, line 98: I still think this should be a different method, say `idp_offers_registration`. This method would then do something like this:
```
idp_offers_registration
   idp_in_list(reg_idps, idp)

idp_in_list
   list.contains( |idp| idp.entity_id = idp)
```

Then `get_idp_choice ` can be used to retrieve the actual value.

Same pattern can be applied for sign-in

CharlesIC: Really the list should be a hash and then we wouldn't have this problem 😬

Wynndow: In your example `idp_in_list` is the same as `get_idp_choice`. The only difference being `get_idp_choice` returns the idp to you, rather than a boolean. The presence (or lack of) a returned IDP means the method can be used as a conditional.

It's the same for `get_idp_from_those_available_for_sign_in` (which is really just a wrapper around `get_idp_choice`). It can be used as a conditional depending on the return value.

CharlesIC: Yep, that's right. I think the main issue with the original was naming, which is better now. 
If you want to do `if idp  = get_idp_from_those_available_for_registration` to avoid calling the method twice I think that would be fine now

CharlesIC: Actually I would also rename `get_idp_from_those_available_for_registration` to `get_idp_from_idps_available_for_registration ` or `get_idp_from_idp_list_for_registration ` - this is the bit that was confusing me

CharlesIC (approved):

paola-roccuzzo:
@Wynndow can you please update the content as follows?

> ##You can no longer verify your identity with [IDP name]
> 
> Go back to the [service you were trying to access](URL) where you can choose either:
> 
> - a new identity provider
> - another way to prove your identity

Screenshot 2021-01-26 at 10 43 13

Unless it's fixed on all pages for a very good reason, please remove the feedback link from this one. 

@sabrinaharris if you figured out how to get the Welsh translation can you go ahead please? Tom or Kirsty should be able to help if you have any questions.

Wynndow:
@paola-roccuzzo Thanks for this. After we spoke yesterday I actually discovered that users can pause their journey when signing in (rather than registering) with Experian which is ...unexpected, so I'm going to have to have a think about what we can do.

Wynndow:
@paola-roccuzzo I can remove the feedback line - can I ask why we want to though?

app/controllers/partials/journey_hinting_partial_controller.rb
```diff
@@ -44,6 +44,11 @@ def last_rp
     last_status_value.nil? ? nil : last_status_value.fetch("RP", nil)
   end
 
+  def last_journey_type
```
CharlesIC, line 47:
```suggestion
  def last_verify_journey_type
```

There already is `user_session_type` method in `UserSessionPartialController` to distinguish between Verify and eIDAS journeys. The journey type here is the "Verify journey type" from the Verify module in `JourneyType`. Maybe better naming is needed to minimise confusion.

https://github.com/alphagov/verify-frontend/blob/6571fcbdf14d204a6bae4f589570056bab955ce7/app/controllers/partials/user_session_partial_controller.rb#L33-L40

app/controllers/partials/user_cookies_partial_controller.rb
```diff
@@ -36,6 +36,7 @@ def set_journey_hint_by_status(idp_entity_id, status, rp_entity_id = nil)
     journey_hint_by_status_value["STATE"] = { IDP: idp_entity_id,
                                               RP: rp_entity_id.nil? ? session[:transaction_entity_id] : rp_entity_id,
                                               STATUS: status }
+    journey_hint_by_status_value["STATE"][:JOURNEY_TYPE] = session[:journey_type] if status == "PENDING"
```
CharlesIC, line 39:
```suggestion
    journey_hint_by_status_value["STATE"][:VERIFY_JOURNEY_TYPE] = session[:journey_type] if status == "PENDING"
```
Maybe better naming?

app/controllers/partials/journey_hinting_partial_controller.rb
Wynndow: Makes sense, will update.

app/controllers/partials/journey_hinting_partial_controller.rb
```diff
@@ -44,6 +44,11 @@ def last_rp
...
+    last_status_value = last_status
+    last_status_value.nil? ? nil : last_status_value.fetch("JOURNEY_TYPE", nil)
```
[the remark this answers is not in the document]
Wynndow: This is just using the same pattern that nearly all the other methods in this module use. I hesitant to just change one, and feels unnecessary to change them all.

app/controllers/partials/user_cookies_partial_controller.rb
Wynndow: Yep, makes sense.

CharlesIC (approved):

JohnWattsGDS (approved):

PR #932, opened 2021-01-25
HUB-744 Provide release infomration to Frontend

j-maynard:
This PR updates the Dockerfile to provide sentry with the current release information as an environment variable.


Wynndow (commented):
So the build-arg `release` now needs to be passed it when the image is built? Does that happen in the pipeline?

CharlesIC (approved):
Remember to also pass the `build-args` file to the `put verify-frontend-image` step in the pipeline

jkunle (approved):

PR #936, opened 2021-01-27
HUB-794 Remove modal dialog

JohnWattsGDS:
We don't need a modal dialog if the user is going to timeout.

Remove related javascript and stylesheets.

Remove words for dialog.

Modify controller that renders dialog.


jkunle (approved):
LGTM

instantiator (approved):
LGTM

PR #939, opened 2021-02-03
HUB-242: Make short-Hub page title more accessible

jkunle:
see: https://govukverify.atlassian.net/browse/HUB-242
**Why**
As a part of UCD feedback for the ABC testing variant C (short hub) journey, and this uncovered the following accessibility feedback:

We generally advise that page titles and H1s should match. Use the H1 as the definitive version where they differ. This seems to be inconsistent everywhere but I’ve spotted them not matching on:

'Sign in with GOV.UK Verify' page
'What GOV.UK Verify is' page
'About you' page
'Which of these do you have available right now?' page
‘Pick a company to verify you’ page

**How**
This requirement implies that **page titles** and **H1** should be semantically
similar. It was decided all page titles are made exactly the same as the respective H1
this would ensure that this accessibility criteria is always respected and
prevent any deviation or inconsistencies from occurring in overtime.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,8 +1,8 @@
 module ApplicationHelper
-  def page_title(title_key, locale_data = {}, extra_string = nil)
+  def page_title(title_key = nil, locale_data = {}, extra_string = nil)
     title = ""
     title << "#{t('title.error', locale_data)}: " if flash[:errors]
-    title << t(title_key, locale_data)
+    title << (block_given? ? yield : t(title_key, locale_data))
     en_title = [t(title_key, locale_data.merge(locale: :en)), extra_string, "GOV.UK Verify", "GOV.UK"]
     en_title << session[:requested_loa] if session[:requested_loa]
     content_for :page_title, title
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,5 +1,7 @@
 module ApplicationHelper
   def page_title(title_key = nil, locale_data = {}, extra_string = nil)
+    raise ArgumentError.new("Missing page title") if title_key.nil? && !block_given?
+
     title = ""
     title << "#{t('title.error', locale_data)}: " if flash[:errors]
     title << (block_given? ? yield : t(title_key, locale_data))
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,7 +5,7 @@ module ApplicationHelper
     title = ""
     title << "#{t('title.error', locale_data)}: " if flash[:errors]
     title << (block_given? ? yield : t(title_key, locale_data))
-    en_title = [t(title_key, locale_data.merge(locale: :en)), extra_string, "GOV.UK Verify", "GOV.UK"]
+    en_title = [t(title_key, locale_data.merge(locale: :en)), extra_string, "GOV.UK Verify"]
     en_title << session[:requested_loa] if session[:requested_loa]
     content_for :page_title, title
     analytics_title en_title.compact.join(" - ")
--- a/app/views/failed_registration/_custom_failed_registration.html.erb
+++ b/app/views/failed_registration/_custom_failed_registration.html.erb
@@ -1,4 +1,4 @@
-<h1 class="govuk-heading-l"><%= raw transaction.custom_fail_heading  %></h1>
+<h1 class="govuk-heading-l"><%= raw transaction.custom_fail_heading %></h1>
 
 <p class="govuk-body"><%= raw(transaction.custom_fail_what_next_content % { idp_name: idp.display_name }) %></p>
 
--- a/app/views/failed_registration/_non_continue_rp.html.erb
+++ b/app/views/failed_registration/_non_continue_rp.html.erb
@@ -1,4 +1,5 @@
-<h1 class="govuk-heading-l"><%= t 'hub.failed_registration.alt_heading', idp: idp.display_name %></h1>
+<%= page_title 'hub.failed_registration.heading', idp_name: idp.display_name %>
+<h1 class="govuk-heading-l"><%= t 'hub.failed_registration.heading', idp_name: idp.display_name %></h1>
 <p class="govuk-body"><%= t 'hub.failed_registration.you_can_still', service: @transaction.name %></p>
 
 <%= render partial: @remaining_options_partial %>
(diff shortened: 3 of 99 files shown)


CharlesIC:
Is this going to affect Matomo analytics?

config/locales/en.yml
```diff
@@ -260,19 +248,17 @@ en:
...
     privacy_notice:
-      title: Privacy notice
+      heading: Privacy notice
     verify_services:
-      title: Verify services
       heading: GOV.UK Verify
       message: "The current services using GOV.UK Verify are:"
     transaction_list:
-      title: Find the service you were using to start again
+      sub_heading: Find the service you were using to start again
```
CharlesIC, line 258: Why is this `sub_heading` instead of `heading`?

app/views/failed_registration/index_LOA2.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= page_title 'hub.failed_registration.title' %>
+<%= page_title 'hub.failed_registration.heading' %>
```
CharlesIC, line 1: Token `hub.failed_registration.heading` needs variable `idp_name`

app/views/failed_registration/index_LOA1.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= page_title 'hub.failed_registration.title' %>
+<%= page_title 'hub.failed_registration.alt_heading', idp: @idp.display_name %>
```
CharlesIC, line 1: Why does this file have `hub.failed_registration.alt_heading` but `index_LOA2.html.erb` has `hub.failed_registration.heading`? Both say the same thing but with slightly different wording. We should pick one and remove the other.

app/views/failed_registration/_continue_rp.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= page_title 'hub.failed_registration.heading' %>
```
CharlesIC, line 1:
```suggestion
```
This is a partial view so I don't think it should  have its own title. The parent views `index_continue_on_failed_registration_LOA{1|2}.html.erb` set the title for the whole page. I'm not sure if setting the title twice would even work

app/views/failed_registration/_non_continue_rp.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= page_title 'hub.failed_registration.alt_heading', idp: idp.display_name %>
```
CharlesIC, line 1:
```suggestion
```
This is a partial view so I don't think it should have its own title. The parent views `failed_registration/index_LOA{1|2}.html.erb` set the title for the whole page. I'm not sure if setting the title twice would even work

app/views/failed_registration/_custom_failed_registration.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= page_title(raw transaction.custom_fail_heading)  %>
```
CharlesIC, line 1:
```suggestion
```
This is a partial view so I don't think it should have its own title. The parent view index_continue_on_failed_registration_LOA2.html.erb set the title for the whole page. I'm not sure if setting the title twice would even work

app/views/feedback_landing/index.html.erb
```diff
@@ -1,14 +1,13 @@
...
 
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l"><%= @feedback_landing_heading %></h1>
     <h2 class="govuk-heading-m"><%= link_to(t('hub.feedback_landing.idps.heading'), t('hub.feedback_landing.idps.href')) %></h2>
     <p><%= raw t('hub.feedback_landing.idps.description', bank_holidays: link_to(t('hub.feedback_landing.idps.bank_holidays'), "http://www.gov.uk/bank-holidays")) %></p>
 
-    <% unless @other_ways_text.nil? %>
+    <% if @other_ways_text.present? %>
```
CharlesIC, line 10: Nice 👍

app/views/shared/_disconnected_idp_list.erb
```diff
@@ -1,5 +1,5 @@
 
-<h2 class="govuk-heading-m"><%= t 'hub.certified_companies_disconnected.title' %></h2>
+<h2 class="govuk-heading-m"><%= t 'hub.certified_companies_disconnected.sub_heading' %></h2>
```
CharlesIC, line 2:
```suggestion
<h2 class="govuk-heading-m"><%= t 'hub.certified_companies_disconnected.heading' %></h2>
```
I don't see `certified_companies_disconnected.sub_heading` in `en.yml`

config/locales/en.yml
```diff
@@ -282,14 +268,13 @@ en:
...
     will_it_work_for_me:
-      title: Can I be verified?
       heading: About you
       question:
         option_yes: 'Yes'
         above_age_threshold: Are you 20 or over?
         resident_last_12_months: Have you lived in the UK for the last 12 months?
         not_resident_reason:
-          title: Which of these applies to you?
+          sub_heading: Which of these applies to you?
```
CharlesIC, line 277:
```suggestion
          heading: Which of these applies to you?
```
Why `sub_heading` instead of `heading`?

CharlesIC (commented):
A few comments to clarify below

spec/features/users_browser_sends_client_side_analytics_spec.rb
```diff
@@ -115,22 +115,22 @@
...
+          "action_name" => "Sign in with GOV.UK Verify - GOV.UK Verify - GOV.UK - LEVEL_2",
           "url" => /start/,
           "new_visit" => "0",
         ),
       )
       visit "/start"
     end
 
-    it "sends a page view with a new_visit parameter if new session and on next page the parameter is not present" do
+    it "sends a page view with a new_visit parameter What GOV.UK Verify is" do
```
Wynndow, line 126: Was this meant to change?

app/helpers/application_helper.rb
```diff
@@ -1,5 +1,7 @@
 module ApplicationHelper
   def page_title(title_key = nil, locale_data = {}, extra_string = nil)
+    raise ArgumentError.new("Missing page title") if title_key.nil? && !block_given?
```
Wynndow, line 3: I guess this should catch issues before we deploy anything. I don't think we'd want uses to get an error if the title was wrong.

Wynndow (commented):
What was the decision about if these changes would affect analytics data?

app/helpers/application_helper.rb
jkunle: @jamiepetherbridgeconroy had a look, and said it doesn't impact our analytics

spec/features/users_browser_sends_client_side_analytics_spec.rb
jkunle: Nope, nice spot

Wynndow (dismissed):

CharlesIC:
Can the commits be tidied up please 🙏

Wynndow (approved):

CharlesIC (approved):

PR #941, opened 2021-02-05
HUB-794 Tidy up redundant variables and method

JohnWattsGDS:
`@seconds_to_timeout` and the method `get_seconds_to_timeout` which populated it are no longer used now that the modal dialog has been removed so they have been removed also.

### DIFF
--- a/app/controllers/further_information_controller.rb
+++ b/app/controllers/further_information_controller.rb
@@ -1,6 +1,5 @@
 class FurtherInformationController < ApplicationController
   def index
-    @seconds_to_timeout = get_seconds_to_timeout
     return redirect_to further_information_timeout_path if expired?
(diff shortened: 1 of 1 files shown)


CharlesIC (approved):

app/controllers/further_information_controller.rb
```diff
@@ -57,8 +55,4 @@ def submit_null_attribute
   def expired?
     !session[:assertion_expiry].nil? && Time.parse(session[:assertion_expiry]) <= Time.now.utc
```
jkunle, line 56: `!session[:assertion_expiry].nil? = session[:assertion_expiry].present?`

Wynndow (approved):

PR #942, opened 2021-02-05
BAU: Fix pre-commit

CharlesIC:
This PR fixes issues with pre-commit not working locally due to issues with selenium/capybara not working with some versions of Firefox.

- Use the webdrivers gem which supports Chrome as well as the latest Firefox version
This requires whitelisting a few URLs with WebMock so the library can pull the latest webdriver version

- Remove the old geckodriver-helper gem and update other gems
- Add support for running tests with Chrome
- Add logic to pre-commit.sh to run with Chrome
- Rename the docker pre-commit to be more clear
- Remove unused docker-run.sh script
- Clean up Dockerfile

Test by running:
`pre-commit.sh`
`pre-commit.sh --chrome`
`pre-commit-docker.sh`


jkunle (approved):
Nice, it all works 🥇

pre-commit.sh
```diff
@@ -19,6 +19,12 @@ else
   git secrets --register-aws
 fi
 
+if [[ $1 == "--chrome" ]]; then
```
j-maynard, line 22: I am not going to block... But I really hate this.  There is a proper way of handling script arguments and we really should follow it.

j-maynard (approved):
Other than my complaint about bash arguments... here have a lion head...

🦁

pre-commit.sh
CharlesIC: I know, if I had more then one arg I'd have used the args method but just for one I didn't bother

PR #944, opened 2021-02-10
BAU Fix options in pre-commit script

j-maynard:
I couldn't help myself.  Bash has a a sensible way of handling command line options which I think should generally be followed so I have updated the pre-commit script to follow this pattern.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,5 +1,32 @@
 #!/usr/bin/env bash
 
+show_help() {
+  cat << EOF
+Usage:
+
+  Options:
+    -c, --chrome    Tell this script to run tests using chrome driver
+
+    -h, --help      Show's this help message
+EOF
+}
+
+CHROME=false
+
+while [ "$1" != "" ]; do
+    case $1 in
+        -c | --chrome)          CHROME=true
+                                ;;
+        -h | --help)            show_help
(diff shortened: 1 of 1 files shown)


pre-commit.sh
```diff
@@ -19,7 +46,7 @@ else
   git secrets --register-aws
 fi
 
-if [[ $1 == "--chrome" ]]; then
+if [[ $CHROME == "true" ]]; then
```
jkunle, line 49: So is this really $CHROME == "true" , i.e. is true == "true" in bash if not this would still be running under the default firefox

j-maynard: So Bash doesn't actually understand true or false... Its comparing strings.  If you look at line 14 and 18 you'll see this is the two points we set the variable `$CHROME`... The if statement then just does a quick string comparison.  You could equally use 0 and 1 but I find personally true and false easier to read.

jkunle (approved):

pre-commit.sh
idrop: TIL 1) bash case statement doesn't do a fall through unless you issue `;&` (I think).
TIL 2) the bash convention of handling args is to while on $1 and shift ( I think).

idrop (approved):

PR #948, opened 2021-02-17
BAU: fix alignment issues on picker page

jkunle:
This PR is the 3rd refinement of this
- Horizontally center buttons and buttons same width
- Center table rows vertically
- Reduce space between primary and secondary buttons

image

  **when in mobile view**

image

- Also attempted to fix who do you have an identity account with?
  The buttons are now relatively the same width and only wrap if
  the text is really very long

image

Other buttons are not affected by this change

image

### DIFF
--- a/app/assets/stylesheets/pages/_company.scss
+++ b/app/assets/stylesheets/pages/_company.scss
@@ -44,13 +44,12 @@
 .company-logo {
   position: relative;
   width: 50%;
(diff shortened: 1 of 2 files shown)


instantiator (approved):
This seems to adjust well to a full range of widths. Thanks Ola!

JohnWattsGDS (approved):

PR #950, opened 2021-02-22
BAU: Fix who do you have account with in mobile

jkunle:
Current master is looking like this in mobile view
image
This PR correct's this
image


j-maynard (approved):
🦁

Wynndow:
In that second screen shot, it still looks like there's an issue? The EU flag looks like it's behind the button?

PR #952, opened 2021-03-01
HUB-809 add legend to start page

instantiator:
Christopher Thomas took a look at hub as a part of UCD feedback for the short hub, before it was famous. (The variant formerly known as "C".)

* The hub's start page is missing a legend.
* It should contain the question that the user is being asked.
* In this case, that's "Have you used GOV.UK Verify before?"
* Also, Welsh.

### DIFF
--- a/app/views/start/start.html.erb
+++ b/app/views/start/start.html.erb
@@ -9,13 +9,15 @@
 
   <%= form_for @form, url: start_path, html: {id: 'start-page-form', class: 'js-validate', novalidate: 'novalidate'} do |f| %>
     <%= content_tag :div, class: form_question_class do %>
+      <% if @form.errors.include?(:selection_true) %>
+          <span class="govuk-error-message">
+            <%= @form.errors[:selection_true].first %>
+          </span>
+      <% end %>
       <fieldset class="govuk-fieldset">
-        <% if @form.errors.include?(:selection_true) %>
-            <span class="govuk-error-message">
-              <%= @form.errors[:selection_true].first %>
-            </span>
-        <% end %>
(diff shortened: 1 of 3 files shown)


app/views/start/start.html.erb
```diff
@@ -9,12 +9,15 @@
 
   <%= form_for @form, url: start_path, html: {id: 'start-page-form', class: 'js-validate', novalidate: 'novalidate'} do |f| %>
     <%= content_tag :div, class: form_question_class do %>
+      <% if @form.errors.include?(:selection_true) %>
```
Wynndow, line 12: I've had a look compared it to the original, and can't see any difference, but what impact does the error not being in the fieldset have? If any?

instantiator: Good question. I looked at the design system docs, and all their examples had the error summary outside of the `fieldset`. I think it's well illustrated by the example in the section **Where to put the error summary**, in the error summary page: https://design-system.service.gov.uk/components/error-summary/

That said I've not experimented with it inside or outside - as I assumed this was the 'correct' implementation.

Wynndow: Interesting - totally the right thing to do according to the docs. I wonder if the custom styles on the start page (blue and what not) are masking something...
I span up your change locally and compared it to prod, and the error stuff seems exactly the same. Shrug...

#### Your change
Screenshot 2021-03-02 at 16 51 39


#### Prod
Screenshot 2021-03-02 at 16 51 13

Wynndow (approved):

j-maynard (approved):
🦁

PR #955, opened 2021-03-09
HUB-853 DRAFT First attempt at warning

j-maynard:


app/controllers/sign_in_controller.rb
```diff
@@ -31,15 +31,54 @@ def index
   def select_idp
     select_viewable_idp_for_sign_in(params.fetch("entity_id")) do |decorated_idp|
       set_journey_hint_followed(decorated_idp.entity_id)
-      sign_in(decorated_idp.entity_id, decorated_idp.display_name)
-      redirect_to redirect_to_idp_sign_in_path
+      idp_reg_expiry = if decorated_idp.provide_authentication_until.nil? || decorated_idp.provide_authentication_until == true
```
jkunle, line 34: Maybe it's me, i don't quite get the logic here generally, a description would have been nice. 

- Is `provide_authentication_until` a DateTime holder or a boolean `hash.fetch("provideAuthenticationUntil", true)` suggests it is a boolean; but it can't be b'cos `DateTime.parse(decorated_idp.provide_authentication_until)`
- I think the below should be encapsulated in `decorated_idp`
```
idp_reg_expiry = if decorated_idp.provide_authentication_until
                         6.month.from_now
                       else
                         DateTime.parse(decorated_idp.provide_authentication_until)
                       end
      if (idp_reg_expiry - 1.month) > DateTime.now
```
So you can do something like this `if (decorated_idp.encapsulated_logic_here > DateTime.now || session[:warning_shown] == true` In sign_in_controller
In rails you `DateTime.now + 6.month` is equivalent to `6.month.from_now`

app/controllers/sign_in_controller.rb
```diff
@@ -32,15 +32,45 @@ def select_idp
     select_viewable_idp_for_sign_in(params.fetch("entity_id")) do |decorated_idp|
       set_journey_hint_followed(decorated_idp.entity_id)
       sign_in(decorated_idp.entity_id, decorated_idp.display_name)
-      redirect_to redirect_to_idp_sign_in_path
+      if !decorated_idp.provide_authentication_until.nil? && (decorated_idp.provide_authentication_until - 1.month) < DateTime.now
```
jkunle, line 35: `!decorated_idp.provide_authentication_until.nil?` is equivalent `decorated_idp.provide_authentication_until` (prescence of a thing is true / abscence of thing is false) or if really want to pedantic decorated_idp.provide_authentication_until.present?

app/controllers/sign_in_controller.rb
```diff
@@ -32,15 +32,45 @@ def select_idp
...
+        render json: redirect_obj
+      else
+        ajax_idp_redirection_sign_in_request(decorated_idp.entity_id)
+      end
+    end
+  end
+
+  def warn
+    #logger.warn("Session => #{session[:selected_provider]["identity_provider"]["entity_id"]}")
+    select_viewable_idp_for_sign_in(session[:selected_provider]["identity_provider"]["entity_id"]) do |decorated_idp|
```
jkunle, line 68: Didn't this `@idp = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate(selected_identity_provider)` work

app/controllers/sign_in_controller.rb
instantiator: ok fine `.present?` is nice

instantiator (approved):
🚀

CharlesIC (approved):

PR #960, opened 2021-03-22
HUB-853 Add functional test for new warning page

j-maynard:
This adds the functional test for the new IDP disconnecting warning page.

### DIFF
--- a/spec/features/user_visits_idp_warning_page_spec.rb
+++ b/spec/features/user_visits_idp_warning_page_spec.rb
@@ -1,0 +1,49 @@
+require "feature_helper"
+require "api_test_helper"
+require "piwik_test_helper"
+require "sign_in_helper"
+
+RSpec.describe "user selects an IDP on the sign in page" do
+  def given_api_requests_have_been_mocked!
+    stub_session_select_idp_request(encrypted_entity_id)
+    stub_session_idp_authn_request(originating_ip, location, false)
+  end
+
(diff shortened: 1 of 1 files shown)


spec/features/user_visits_idp_warning_page_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+                                   {  "simpleId" => "stub-idp-two",
+                                      "entityId" => "http://idcorp-two.com",
+                                      "levelsOfAssurance" => %w(LEVEL_2),
+                                      "provideAuthenticationUntil" => 10.day.from_now.to_s }])
+    visit "/#{t('routes.sign_in', locale: locale)}"
+  end
+
+  def when_i_click_start_now
+    click_link("begin-registration-route")
+  end
```
instantiator, line 27: It looks a lot like the other features rely on these sorts of methods living in the helpers in the spec folder. I wonder if these should live in one of the helpers there, too? `sign_in_helper.rb`?

JohnWattsGDS (approved):

spec/features/user_visits_idp_warning_page_spec.rb
j-maynard: As I think we only have 2 tests depending on provide_authentication_until I am personally incline to leave this as is.  Also the setup for this test is slightly different to the other tests I wrote the other week.

instantiator (approved):
Thanks Jamie!

PR #966, opened 2021-04-07
BAU: Bump Ruby to 2.7.3

Wynndow:
**PR for pipeline changes required to support this https://github.com/alphagov/verify-infrastructure-config/pull/518**

The 2.6 line is only supported for security updates.

There were a couple of CVEs released for 2.6.6 and 2.7.2, so moving to
2.7.3.

https://www.ruby-lang.org/en/news/2021/04/05/xml-round-trip-vulnerability-in-rexml-cve-2021-28965/
https://www.ruby-lang.org/en/news/2021/04/05/tempfile-path-traversal-on-windows-cve-2021-28966/

This is also pins the version of Webmock to 3.10.0. This is what we were
using before this update. Version 3.11.0 has a change which is
incompatible with the version of http (~> 2.2.0) we're pinned to. The
change is here: https://github.com/bblimke/webmock/pull/896

The raises an interesting point. We're pinned to version 2.2 of http and
have been since 2016: https://github.com/alphagov/verify-frontend/commit/f93d26c1549b663120baf25b51d6d4e3f1c9d2b8

Maybe we should be updating out stuff...


jkunle (commented):
👍🏾

AmritSidhu (approved):

jkunle (approved):

PR #968, opened 2021-04-13
BAU: Bump http gem to 4.4.1

Wynndow:
This is the latest version. Version 4.3.0 included a bug fix for a
regression with ruby 2.7. If we want to use 2.7 we'll need this version.

It updates the http config initializer to work with the new gem.

It also removes the pin on Webmock, which was initially put in place
when we upgraded to ruby 2.7 to mitigiate an issue with having http
pinned to version 2. This is no longer an issue.


JohnWattsGDS (approved):
🤷🏻‍♂️

instantiator (approved):

PR #970, opened 2021-05-12
HUB-920: Add logging to help diagnose issue

Wynndow:
We're getting TypeErrors in Sentry which are implying that the
transaction ID retrieved from the users session is nil. We have a
validator however that should be preventing requests without this value
in their session from being accepted.

This logging will allow us to inspect the session in these cases to
determine what is actually happening.


j-maynard (approved):
🦁

chriscoffee (approved):
:shipit:

PR #971, opened 2021-05-12
HUB-920: Fix logging

Wynndow:
The session object is not available in a model (confusingly it is when
running tests...).

The only reasonable way to get that information into the model is to
pass it in from the controller.


JohnWattsGDS (approved):
🤷🏻‍♂️

jkunle (approved):

j-maynard (approved):
🦁

PR #974, opened 2021-05-19
HUB-920: Fail invalid CSRF with exception on sign in

Wynndow:
We set `protect_from_forgery with: :exception` in the
`ApplicationController`. This means that if we receive an invalid CSRF
token Rails will throw an error. This behaviour is propagated to all
controllers that inherit from the `ApplicationController` (all of them).

The default behaviour of `protect_from_forgery` (if we didn't set `with:
:exception`) is to use the `:null_session` strategy. This sets the
session to empty for the duration of the request but not completely
reset it.

In the sign-in controller we redefine `protect_from_forgery` so that we
exclude one particular action. We don't set the `with:` option though.
This overrides the `with: :exception` set in the `ApplicationController`
and means we run with the default for this controller.

This is problematic for us. The session is validated before the CSRF
token check. This results in the session passing validation but the
getting wiped by the CSRF protection. The request then continues as
usual. We then attempt to use data from the session, which throws
errors.

This updates the overriding `protect_from_forgery` in the
`SignInController` to maintain the expected behaviour as defined in the
`ApplicaitonController`.

### DIFF
--- a/app/controllers/partials/viewable_idp_partial_controller.rb
+++ b/app/controllers/partials/viewable_idp_partial_controller.rb
@@ -37,7 +37,7 @@ module ViewableIdpPartialController
   end
 
   def current_identity_providers_for_sign_in
-    CONFIG_PROXY.get_idp_list_for_sign_in(session[:transaction_entity_id], session).idps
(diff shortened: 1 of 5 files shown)


j-maynard (approved):
🦁

JohnWattsGDS (approved):

PR #978, opened 2021-05-26
HUB-863: Welsh translations

Wynndow:
Various Welsh translations.

Also fixes an issue with the wrong key name in the Welsh locales file.

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -485,7 +485,7 @@ cy:
       verify_another_company_text: Gallwch geisio nes ymlaen eto neu wirio'ch hunaniaeth gyda chwmni ardystiedig arall.
       company_text: Nid yw %{company} ar gael
     certified_companies_disconnected:
-      sub_heading: Nid yw'r cwmnïau canlynol bellach yn rhan o GOV.UK Verify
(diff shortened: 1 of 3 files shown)


config/locales/cy.yml
```diff
@@ -485,7 +484,7 @@ cy:
       verify_another_company_text: Gallwch geisio nes ymlaen eto neu wirio'ch hunaniaeth gyda chwmni ardystiedig arall.
       company_text: Nid yw %{company} ar gael
     certified_companies_disconnected:
-      sub_heading: Nid yw'r cwmnïau canlynol bellach yn rhan o GOV.UK Verify
+      heading: Nid yw'r cwmnïau canlynol bellach yn rhan o GOV.UK Verify
```
JohnWattsGDS, line 487: I'm assuming you've tested this and it works, I'm just confused at why it didn't break in the past.

JohnWattsGDS (approved):

config/locales/cy.yml
Wynndow: I think it falls back to the english version of it can't find the Welsh

instantiator (approved):

PR #983, opened 2021-09-07
HUB-971: Cleanup

CharlesIC:
HU9-971: Cleanup - continue on failed registration RP
Remove the "continue on failed registration RP" functionality

This is not used in production, so we can remove extra code.

HU9-971: Cleanup - fix failing tests
  - The tests occasionally fail due to an uninitialised constant "FederationTranslator". This class was removed back in 2018. Remove the remaining bits and replace with I18n.

Original removal commit: 8b52be4

  - Set correct LOA in response_processing_controller spec


j-maynard (approved):
🦁

Wynndow (approved):

PR #985, opened 2021-09-15
HUB0-971: Remove unused "select documents" page

CharlesIC:
**Remove the "select documents" page and the checkboxes form**

The page is no longer used (replaced by static "about documents"), so we can remove the associated styles, views and logic.

  - Remove logic to report user evidence, as users no longer need to provide any evidence to get through to the IDP picker page

  - Render the "something went wrong page" if there no IDPs available for registration
Since we no longer filter IDPs based on the user's evidence, we will always show all available IDPs. If there are no IDPs, this is a critical state which would indicate there is a problem with hub federation config. This should never occur.

**Small fixes and improvements**

- Make Stub API redirect to IDP work for LOA1

The `received-authn-request` needs to be parametrised to work with any relay state as we use a different relay state value for LOA1 journeys.

- Use the `to_json` conversion method to generate JSON strings in Stub API

This keeps the code human-friendly and easier to modify as we don't have to operate on strings, escape special characters or keep track of correct JSON formatting - the computer can do this instead.

- Use a helper method to URL encode a URL

- Make tests in `user_visits_continue_to_your_idp_page_spec` work offline

The `will redirect the user to the IDP on Continue` test only works online as the redirect URL of the RP (www.gov.uk) must be reachable by Capybara/Selenium and the GET request must return a 200 response in order for the `click_button "initiate-single-idp-post"` action to succeed.

If the test is run offline, the redirect URL cannot be resolved and the test fails even though there are no bugs in the code.

To resolve this, set the redirect URL to the `/test-saml` page of the local Capybara server. This way, the URL can always be resolved regardless of the presence or lack of internet connection.

- Remove unused tests and translations for `redirect_to_idp_question`


j-maynard (approved):
🦁

Wynndow (approved):

PR #992, opened 2021-11-03
HUB-1117: One IDP updates to How Verify Works Page

j-maynard:
Preparing for a single IDP world we've been asked to make some content changes to the How Verify Works page.  This commit introduces an if statement to the view which renders the existing content if there is more than one page or the new content if only 1 IDP is present.

Multiple IDP Jounrey:
image

Single IDP Journey:
image

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -18,6 +18,7 @@ class AboutController < ApplicationController
   end
 
   def about_documents
+    @identity_providers = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(identity_providers_available_for_registration)
     render :documents
   end
 
--- a/app/views/about/documents.html.erb
+++ b/app/views/about/documents.html.erb
@@ -1,24 +1,58 @@
-<%= page_title 'hub.about.documents.heading' %>
+<% if @identity_providers.size > 1 %>
+  <%= page_title 'hub.about.documents.multi_idp.heading' %>
+<% else %>
+  <%= page_title 'hub.about.documents.one_idp.heading', idp: @identity_providers.first.display_name %>
+<% end %>
+
 <% content_for :feedback_source, 'ABOUT_DOCUMENTS_PAGE' %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+<% if @identity_providers.size > 1 %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
 
-    <h1 class="govuk-heading-l"><%= t 'hub.about.documents.heading' %></h1>
-    <p><%= t 'hub.about.documents.explanation' %></p>
+      <h1 class="govuk-heading-l"><%= t 'hub.about.documents.multi_idp.heading' %></h1>
+      <p><%= t 'hub.about.documents.multi_idp.explanation' %></p>
 
-    <p><%= t 'hub.about.documents.document_list.heading' %></p>
-    <p><%= t 'hub.about.documents.document_list.list_html' %></p>
+      <p><%= t 'hub.about.documents.multi_idp.document_list.heading' %></p>
+      <p><%= t 'hub.about.documents.multi_idp.document_list.list_html' %></p>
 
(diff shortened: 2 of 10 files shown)


app/controllers/about_controller.rb
```diff
@@ -1,4 +1,5 @@
 require "partials/viewable_idp_partial_controller"
+require "pry"
```
CharlesIC, line 2: This needs to be removed from a non-test class

app/controllers/about_controller.rb
```diff
@@ -9,7 +10,12 @@ def about_verify
     @tailored_text = current_transaction.tailored_text
     @identity_providers = IDENTITY_PROVIDER_DISPLAY_DECORATOR.decorate_collection(identity_providers_available_for_registration)
     @next_page_path = is_journey_loa1? ? choose_a_certified_company_path : will_it_work_for_me_path
-    render :how_verify_works
+    if @identity_providers.size > 1
+      render :how_verify_works
+    else
+      @idp = @identity_providers.first
+      render :how_verify_works_single_idp
+    end
```
CharlesIC, line 18: We don't need a duplicate template. The `how_verify_works` template can have an if statement to display different content if there's only one IDP

spec/controllers/about_controller_spec.rb
```diff
@@ -21,7 +21,10 @@
       end
 
       it "renders the certified companies for on the combined about view" do
-        expect(identity_provider_display_decorator).to receive(:decorate_collection).with(a_list_of_size(6)).and_return([])
+        # expect(IDENTITY_PROVIDER_DISPLAY_DECORATOR).to receive(:decorate_collection).once do |idps|
+        #   idps.each { |idp| expect(idp.levels_of_assurance).to include "LEVEL_2" }
+        # end
+        expect(identity_provider_display_decorator).to receive(:decorate_collection).with(a_list_of_size(6)).and_return([{ "simpleId" => "stub-idp-loa1", "entityId" => "http://idcorp-loa1.com", "levelsOfAssurance" => %w(LEVEL_1 LEVEL_2), "enabled" => true },{ "simpleId" => "stub-idp-loa2", "entityId" => "http://idcorp-loa2.com", "levelsOfAssurance" => %w(LEVEL_1 LEVEL_2), "enabled" => true },])
```
CharlesIC, line 27: This shouldn't change as this is the default case with multiple IDPs and we haven't modified the default list for this test.

spec/controllers/about_controller_spec.rb
```diff
@@ -42,7 +45,10 @@
       end
 
       it "renders the LOA2 certified companies for the combined about view" do
-        expect(identity_provider_display_decorator).to receive(:decorate_collection).with(a_list_of_size(6)).and_return([])
+        # expect(IDENTITY_PROVIDER_DISPLAY_DECORATOR).to receive(:decorate_collection).once do |idps|
+        #   idps.each { |idp| expect(idp.levels_of_assurance).to include "LEVEL_2" }
+        # end
+        expect(identity_provider_display_decorator).to receive(:decorate_collection).with(a_list_of_size(6)).and_return([{ "simpleId" => "stub-idp-loa1", "entityId" => "http://idcorp-loa1.com", "levelsOfAssurance" => %w(LEVEL_1 LEVEL_2), "enabled" => true },{ "simpleId" => "stub-idp-loa2", "entityId" => "http://idcorp-loa2.com", "levelsOfAssurance" => %w(LEVEL_1 LEVEL_2), "enabled" => true },])
```
CharlesIC, line 51: This shouldn't change as this is the default case with multiple IDPs and we haven't modified the default list for this test.

spec/controllers/about_controller_spec.rb
```diff
@@ -71,4 +77,20 @@
       end
     end
   end
+
+  context "SINGLE_IDP" do
+    before(:each) do
+      stub_request(:get, CONFIG.config_api_host + "/config/transactions/enabled")
```
CharlesIC, line 83: there's a helper method for that

idrop (approved):

jkunle (dismissed):

jkunle (dismissed):

jkunle (approved):

JohnWattsGDS (approved):

PR #993, opened 2021-11-12
HUB-1118 Last page update for the single idp world

j-maynard:
This is the last change for single ISP world which removes the option to try another IDP on the cancel registration controller.


JohnWattsGDS (approved):

jkunle:
```
I think you should have had just one context here, instead of two, that are exactly the same wording

context "If there is more than one IDP" do
   it "the page is rendered with the correct content for LOA2 journey" do
   it "the page won't have try another idp" do
end

```

jkunle (approved):

PR #999, opened 2022-04-13
Upgrade ruby version to 2.7.6 to fix CVE-2022-28739

ctazeez:
This is a fix for https://www.ruby-lang.org/en/news/2022/04/12/buffer-overrun-in-string-to-float-cve-2022-28739/


AmritSidhu (approved):

jkunle:
What's with the Travis failures, has this actually worked?

PR #1000, opened 2022-04-13
BAU: Add github action

AmritSidhu:
We are removing travis and replacing it with a github action.
This action installs the required dependencies, runs the rake tasks and then the tests
We have also updated the Ruby version to be 2.7.6 - This is a fix for https://www.ruby-lang.org/en/news/2022/04/12/buffer-overrun-in-string-to-float-cve-2022-28739/


JohnWattsGDS (dismissed):
🤞🏻 
I'm assuming you want to see this working before removing the Travis file ...

instantiator:
🎉 🎈 1️⃣0️⃣0️⃣0️⃣ 🎈 🎉

instantiator (approved):

PR #1075, opened 2022-11-16
stop auto generating services page

elena-vi:
## What 

The list of services on [this page](https://www.signin.service.gov.uk/verify-services) currently reflects which ones are still connected in the config. 

We need to stop this from happening, and just show the list of services in here as plain text that we’ll manually update in future. 

## Why

We’re going to let some services stay technically connected to Verify even after they’ve switched off sign ups/ins, just in case something goes wrong and they need to revert to using Verify. 

We don’t want this page to be misleading for users, or the journalists who use this page as a source of truth.

Screenshot 2022-11-21 at 13 04 44
Screenshot 2022-11-21 at 13 04 51

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,8 @@ ADD . /verify-frontend/
 
 RUN useradd frontend
 RUN chown -R frontend verify-frontend/
+RUN mkdir /home/frontend
+RUN chown -R frontend home/
 
 USER frontend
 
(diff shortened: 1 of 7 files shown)


JHjava:
@elena-vi PR looks good but please could you add some detail to the PR description and commit of what the change is trying to achieve and why it is needed

owenblacker (approved):
LGTM

dushan-madetech (approved):

PR #1078, opened 2022-11-23
add updated gov uk styles

elena-vi:
Styles are not coming through on the notification component 

Screenshot 2022-11-23 at 13 39 41


owenblacker (commented):
That's a lot of additional and updated files. Have I correctly understood that you're adding in the Gov.uk styles SCSS because it's not pulling through right?

That feels like it might be a sledgehammer/walnut solution, but I'm open to persuasion…

owenblacker (approved):
Makes sense

dushan-madetech (approved):

